PR #3, opened 2017-05-02
Component/image

wokalski:
Add image

### DIFF
--- a/src/components/imageRe.re
+++ b/src/components/imageRe.re
@@ -40,8 +40,10 @@ external rawImageSourceJS : 'a => rawImageSourceJS = "";
 module Event = {
   type t;
   type error;
+  type _progress = Js.t {. loaded : float, total : float};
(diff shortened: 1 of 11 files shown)


src/components/imageRe.re
```diff
@@ -0,0 +1,134 @@
...
+
+type rawImageSourceJS;
+
+external rawImageSourceJS : 'a => rawImageSourceJS = "";
+
+module Event = {
+  type t;
+  type error;
+  type progress = {loaded: float, total: float};
+  external progress : t => progress = "nativeEvent" [@@bs.get];
```
saschatimme: I think this doesn't work, `progress` would have to be `Js.t {. loaded: float, total: float }`

wokalski: good catch!

wokalski: fixed

PR #4, opened 2017-05-03
Animated API

### LINKED ISSUE #1: Image + Animated
WIP

wokalski:
Fixes #1 
Todo:
- [x] Spring animation config
- [x] Timing animation config
- [x] `.rei`


wokalski:
Ready for review

wokalski:
I think I should extract the animated components into another file. Also maybe it's worth it to consider other solution than functors for them but it was my first thought to use them.

wokalski:
I'm also a bit bummed about the repetition in `.rei` but on the other hand it feels more readable.

Issue #7, opened 2017-05-06
Better bindings to `Animated.event` and friends


wokalski:
Animated is pretty hard to type correctly because it uses mad dynamic js. The problem with it is that in the current version of bindings `event` is a completely generic function and you don't get any assist from the compiler when working with it. It's a problematic one because `event` returns a function which [takes `varargs`](https://github.com/facebook/react-native/blob/master/Libraries/Animated/src/AnimatedImplementation.js#L2468-L2485). What we should be looking at when solving this issue is not necessarily resembling the existing Animated API but rather making it work in the context of these bindings.

PR #9, opened 2017-05-08
Event api/PanResponder

wokalski:
This is a long overdue PR. The main changes include addition of `PanResponder` (solving #5) and fixes to the `Animated.Event` API (#7). 

There are few other fixes including:
- addition of getters for Animated.ValueXY
- Fixes to animated binding implementation
- few new styles
- The combine function in `Style` was fixed

### DIFF
--- a/src/styleRe.re
+++ b/src/styleRe.re
@@ -1,8 +1,7 @@
 type t;
 
 external to_style : Js.Dict.t Js.Json.t => t = "%identity";
-
-external style_to_dict : t => Js.Dict.t Js.Json.t = "%identity";
+external array_to_style : array t => t = "%identity";
 
 type style =
   | ArrayStyle string (array Js.Json.t)
@@ -15,9 +14,8 @@ type style =
       string [ | `value AnimatedRe.Value.t | `interpolation AnimatedRe.Interpolation.t];
 
 let combine a b => {
-  let entries =
-    Array.append (Utils.dictEntries (style_to_dict a)) (Utils.dictEntries (style_to_dict b));
-  Utils.dictFromArray entries |> to_style
+  let entries = Array.of_list [a, b];
+  array_to_style entries;
 };
 
 let arrayStyle key value => ArrayStyle key value;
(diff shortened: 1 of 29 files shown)


src/styleRe.re
```diff
@@ -15,9 +14,8 @@ type style =
       string [ | `value AnimatedRe.Value.t | `interpolation AnimatedRe.Interpolation.t];
 
 let combine a b => {
-  let entries =
-    Array.append (Utils.dictEntries (style_to_dict a)) (Utils.dictEntries (style_to_dict b));
-  Utils.dictFromArray entries |> to_style
+  let entries = Array.of_list [a, b];
```
saschatimme: You can create an array directly with `[| a, b |]`.

src/styleRe.re
```diff
@@ -15,9 +14,8 @@ type style =
       string [ | `value AnimatedRe.Value.t | `interpolation AnimatedRe.Interpolation.t];
 
 let combine a b => {
```
saschatimme: Does the array syntax also work with `StyleSheets`, i.e. can a value be an array? Otherwise we would need two separate functions

src/styleRe.re
wokalski: Ah, didn't know about this!

src/styleRe.re
wokalski: No, it doesn't. I had mixed. 

Usually I either put something in separate styles and combine them in render, or put in one object in a `StyleSheet`.I have never combined styles in stylesheets because I always define styles inline. 

On the other hand it's a completely logical thing to do - I should add it. Also, it makes me come back to the list-style situation. I'd really like to define styles as just lists. Then our users wouldn't have to know all those functions.

What if we changed `StyleSheet` to the following:
```
let create (x: list ('a, Style.t)): StyleSheet.t 'a;
let get 'a : Style.t;
```

wokalski: Nevermind, it won't work if one combines the styles from stylesheet in a render.

saschatimme:
Just did the quick fix myself :)

PR #10, opened 2017-05-08
Functorize SectionList

saschatimme:
The previous Version doesn't had a satisfying user experience.

 ### Previously
```reason

module UIExplorerApp = {
  include ReactRe.Component;
  type props = {components: array ExampleList.item};
  let name = "UIExplorerExampleList";

  let renderItem sectionProps => {
      <Text>(ReactRe.stringToElement (sectionProps##item: ExampleList.item).key)</Text>
  };

  let render {props} => { 
      <View style=styles##listContainer>
        <SectionList
            sections=[| SectionList.section data::props.components key::"components"  ()|]
            renderItem=renderItem
           keyExtractor=(fun item _ => item.key)
        />
    </View>  
    };
};

include ReactRe.CreateComponent UIExplorerApp;
```
Notice that the compiler cannot infer the type of `item` in `renderItem`.

First I changed that `renderItem` takes a record instead of a `Js.t ...` object since I found the sudden contact with the FFI akward and now it resembles more the usual `render` function.

This resulted in the problem that a pattern match is only possible if the record is defined in the same module (without explicit annotation).
To get everything working as I want, I had to pick bigger weapons. 

### Now
```reason

module UIExplorerApp = {
  include ReactRe.Component;
  include SectionList.Make {type item = ExampleList.item };
  type props = {components: array ExampleList.item};
  let name = "UIExplorerExampleList";
  let renderItem {item} => <Text> (ReactRe.stringToElement item.key) </Text>;
  let render {props} =>
    <View style=styles##listContainer>
      <SectionList
        sections=[|SectionList.section data::props.components key::"components" ()|]
        renderItem
        keyExtractor=(fun item _ => item.key)
      />
    </View>;
};

include ReactRe.CreateComponent UIExplorerApp;
```

The `SectionList` module is now a functor which just takes its `item` type as argument.
We include the resulting module to get the type definitions in the current module.
Our ReactComponent has again the `SectionList` module name which overwrites the `SectionList` from `ReactNative`.

### DIFF
--- a/src/components/sectionListRe.re
+++ b/src/components/sectionListRe.re
@@ -1,23 +1,17 @@
 external view : ReactRe.reactClass = "SectionList" [@@bs.module "react-native"];
 
-module Make (Item: {type item;}) => {
-  type section =
-    Js.t {
-      .
-      data : array Item.item,
-      key : Js.Undefined.t string,
-      renderItem :
-        Js.Undefined.t (
-          Js.t {
-            .
-            item : Item.item,
-            index : int,
-            section : section,
-            separators : Js.t {. highlight : unit => unit, unhighlight : unit => unit}
-          } =>
-          ReactRe.reactElement
-        )
-    };
+module CreateComponent (Item: {type item;}) => {
+  type renderBag = {
+    item: Item.item,
+    index: int,
+    section,
+    separators: Js.t {. highlight : unit => unit, unhighlight : unit => unit}
+  }
+  and section = {
+    data: array Item.item,
+    key: option string,
+    renderItem: option (renderBag => ReactRe.reactElement)
+  };
   type viewToken =
     Js.t {
       .
@@ -27,12 +21,6 @@ module Make (Item: {type item;}) => {
       isViewable : Js.boolean,
       section : section
     };
-  type renderBag = {
-    item: Item.item,
-    index: int,
-    section,
-    separators: Js.t {. highlight : unit => unit, unhighlight : unit => unit}
-  };
   type separatorProps = {
     highlighted: bool,
     leadingItem: option Item.item,
@@ -57,8 +45,15 @@ module Make (Item: {type item;}) => {
     trailingSection
   };
   module SectionList = {
-    let section ::data ::key=? ::renderItem=? () =>
-      Js.Undefined.({"data": data, "key": from_opt key, "renderItem": from_opt renderItem});
+    let renderItemFromJS renderItem jsItems =>
+      renderItem (
+        createRenderBag
+          item::jsItems##item
+          index::jsItems##index
+          section::jsItems##section
+          separators::jsItems##separators
+      );
+    let section ::data ::key=? ::renderItem=? () => {data, key, renderItem};
     let createElement:
       sections::array section =>
       renderItem::(renderBag => ReactRe.reactElement) =>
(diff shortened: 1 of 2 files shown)


wokalski:
Instead of `include SectionList` and `include ReactRe,Component`, `Section.Make` could return a ready to use Reason component. If somebody wants for example state with it, then they put the returned component in their own stateful one.

saschatimme:
Not sure about that since if you want to have a stateful component you have another level of indirection...

wokalski:
Yeah, but that's good. You compose it on top of this component.

src/components/sectionListRe.re
```diff
@@ -1,92 +1,153 @@
...
-  }
-and viewToken 'item =
-  Js.t {
-    .
-    item : 'item,
-    key : string,
-    index : Js.undefined int,
-    isViewable : Js.boolean,
-    section : section 'item
+module Make (Item: {type item;}) => {
```
wokalski: We need consistent naming. `Make` is fine but it should be consistent across the whole project. Maybe `CreateCompnent` would be suitable, too. It's used by `ReactRe`.

src/components/sectionListRe.rei
```diff
@@ -1,68 +1,76 @@
...
-  Js.t {
-    .
-    item : 'item,
-    index : int,
-    section : section 'item,
-    separators :
+module Make:
+  (Item: {type item;}) =>
+  {
+    type section =
```
wokalski: We should decide if we want to expose JS primitives in the public API. It's mixed now. In some places they are converted to records, in others they remain JS. I'm personally in favour of converting them to records, but also, what's more important, keeping it consistent.

src/components/sectionListRe.rei
```diff
@@ -1,68 +1,76 @@
...
+      leadingSection: option section,
+      section,
+      trailingItem: option Item.item,
+      trailingSection: option section
+    };
+    module SectionList: {
+      let section:
+        data::'a =>
+        key::'b? =>
+        renderItem::'c? =>
```
wokalski: Does this type check correctly with `renderItem` later on? Shouldn't it be annotated as `renderBag => ReactRe.reactElement`?

wokalski: Also, we don't need to define `renderItem` inside sections since `item` type is the same for all the sections anyway.

src/components/sectionListRe.rei
saschatimme: Good point, I changed it to a record. But we should keep in mind that this induces some performance penalties (but probably this is more a concern further down the line).

The `viewToken` is too edge case that I wanted to transform it to a record...

src/components/sectionListRe.rei
saschatimme: Sorry that was too late yesterday.
The `renderItem` inside section has also the purpose to overwrite the top `renderItem` function.

src/components/sectionListRe.re
saschatimme: `Make` is common in OCaml but I can see your point, so I changed it to `CreateComponent`

saschatimme:
made some changes!

wokalski:
How is the current API used? Do *you* like it?

src/components/sectionListRe.re
wokalski: Yaa, so maybe we should make it `Make`. I'm not sure, I'm just rooting for consistency

src/components/sectionListRe.rei
wokalski: Yes, the perf penalty might be there.

saschatimme:
Sorry didn't see your previous comment.
I like your second proposal (the first one has again problems with pattern matching since the record types are specified elsewhere), but for example `itemSeparatorComponent` is an optional argument but would have to be in the "bundle".  Obviously we could that with an option, this doesn't become pretty...

To the `keyExtractor` thing: In JS you could specify your `key` in each item, but since our items could be records the JS component doesnt recognize that keys are avaiable. So I made it a required prop since I stumbled on it last night.

saschatimme:
I tend to just go with the current solution and iterate later if we have some more feedback. Are you fine with that?

wokalski:
How does one use the current API? There seems to be additional nested [`SectionList`](https://github.com/BuckleTypes/bs-react-native/pull/10/files#diff-1db4ee06662762fa59aba4115185b374R32) module. Does that mean that I'd have to do the following?
```
<ImageSectionList.SectionList />
```

PR #17, opened 2017-05-10
Fix obtaining NativeEvent* values

wokalski:


src/RNEvent.re
```diff
@@ -3,23 +3,23 @@ type t;
 module NativeEvent = {
   type t;
   external changedTouches : t => array (Js.t {..}) = "" [@@bs.get];
-  external identifier : t => int = "" [@@bs.get];
-  external locationX : t => int = "" [@@bs.get];
-  external locationY : t => int = "" [@@bs.get];
-  external pageX : t => int = "" [@@bs.get];
-  external pageY : t => int = "" [@@bs.get];
+  external identifier : t => float = "" [@@bs.get];
```
saschatimme: is the identifier really a `float`?

wokalski: no, it should be int. Search and replace got me!

saschatimme: happens the best!

It's the same for `timestamp`

wokalski:
Fixed. Additionally I changed externals in `rei` to `let`s.

PR #25, opened 2017-05-15
Polishing

saschatimme:


src/components/textRe.re
```diff
@@ -83,10 +86,16 @@ module CreateComponent (Impl: ViewRe.Impl) :TextComponent => {
           "minimumFontScale": from_opt minimumFontScale,
           "suppressHighlighting": from_opt (Utils.optBoolToOptJsBoolean suppressHighlighting)
         }
+      )
+      children::(
+        switch value {
+        | Some string => [ReactRe.stringToElement string, ...children]
```
wokalski: 😊

saschatimme:
Sure, but it should be the default width from Bucklescript to reduce friction

PR #27, opened 2017-05-15
Add NativeModules and NativeEventEmitter

wokalski:


src/NativeEventEmitter.re
```diff
@@ -0,0 +1,17 @@
+type t;
+
+type emitterSubscription;
+
+external create : NativeModules.t => t =
+  "NativeEventEmitter" [@@bs.new] [@@bs.module "react-native"];
+
+external addListener : t => string => (Js.Dict.t 'a => unit) => emitterSubscription = "" [@@bs.send];
```
saschatimme: Maybe `Js.t. {.} ` instead of `Js.Dict.t 'a`?

src/NativeModules.re
```diff
@@ -0,0 +1,8 @@
+type t;
+
+external nativeModules : Js.Dict.t 'a = "NativeModules" [@@bs.module "react-native"];
```
saschatimme: gleiche wie oben

src/NativeEventEmitter.re
wokalski: yeah, why not

wokalski: hm, no. It doesn't work. `Js.t` is different than `Js.Dict.t`.

saschatimme: Well, if you got something working with the API then we got with your version :)

PR #31, opened 2017-05-16
Add Vibration

saschatimme:
Such a sweet little module 🐤


src/vibrationRe.rei
```diff
@@ -0,0 +1,3 @@
+let vibrate: pattern::array int => repeat::bool => unit;
+
+external cancel : unit => unit = "" [@@bs.scope "Vibrrate"] [@@bs.module "react-native"];
```
wokalski: In other interface files the convention is that we hide externals with `let`s. Do you like the convention?

saschatimme:
I hid the external for now. But we should probably make it consistent in the rest of the repo. I think modules from me have externals exposed and yours not :)

PR #34, opened 2017-05-19
Improve RNTester setup

saschatimme:
Previously I did all kind of magic with the RN Packager, but this was again somehow broken.

Now RNTester app just copies all `src` files from the bindings into the `RNTester` folder.

### DIFF
--- a/RNTester/README.md
+++ b/RNTester/README.md
@@ -8,12 +8,13 @@ Before running the app, make sure you ran:
 
     # in this folder
     npm install
+    # start Bucklescript compiler in watch mode
+    npm run watch
 
 ### Running on iOS
 
 Mac OS and Xcode are required.
 
-- Open a new tab in your terminal and start the bucklescript watcher with `npm run watch`
 - Open `RNTester/RNTester.xcodeproj` in Xcode
 - Hit the Run button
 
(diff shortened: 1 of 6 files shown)


RNTester/README.md
```diff
@@ -8,21 +8,15 @@ Before running the app, make sure you ran:
...
-    # and in the toplevel folder
-    cd ..
-    npm install
 
 ### Running on iOS
 
 Mac OS and Xcode are required.
 
-- Start the packager with `npm start` **[important]**
 - Open a new tab in your terminal and start the bucklescript watcher with `npm run watch`
```
wokalski: This is the same for both platforms. I think it could be avoided with a post install script.

saschatimme: The step is already gone ;)

wokalski: did you forget to push :) ?

saschatimme: Isn't that a negative diff?

wokalski: I mean `npm run watch` 😛

saschatimme: sure we could build it initially, but if the people don't run the watcher, no changes will have an effect

wokalski: haha. What I'm generally advocating is to put it in the `npm install` (because I didn't notice it at first :P). The post install script was an additional idea.

PR #36, opened 2017-06-08
Reason-React 2!

saschatimme:
Adapted everything to the new API (inclusive RNTester) and got rid of the old API. So this would be a breaking change, but I think that's okay at this early stage...

Before we merge this, we should wait for the official roll out.

TODO:
- [ ] Fix `package.json`s to the release versions


src/components/imageRe.re
```diff
@@ -56,17 +53,11 @@ module CreateComponent (Impl: ViewRe.Impl) :ImageComponent => {
...
      * https://github.com/facebook/reason/issues/821 (resolved, not released yet)
      */
-    cache::(
-      [ 
-      `default 
-      | `reload 
-      | `forceCache [@bs.as "force-cache"] 
-      | `onlyIfCached [@bs.as "only-if-cached"]
-      ] [@bs.string])? =>
+    cache::[ | `default | `reload | `forceCache | `onlyIfCached] [@bs.string]? =>
```
wokalski: gotcha!

src/components/textRe.re
```diff
@@ -43,59 +40,60 @@ module CreateComponent (Impl: ViewRe.Impl) :TextComponent => {
...
+              ),
+            "adjustsFontSizeToFit": from_opt (UtilsRN.optBoolToOptJsBoolean adjustsFontSizeToFit),
+            "minimumFontScale": from_opt minimumFontScale,
+            "suppressHighlighting": from_opt (UtilsRN.optBoolToOptJsBoolean suppressHighlighting)
+          }
+        )
+      (
         switch value {
-        | Some string => [ReactRe.stringToElement string, ...children]
+        | Some string => Array.append [|ReasonReact.stringToElement string|] children
```
wokalski: What happened that everything had to become an `array`?

RNTester/src/RNTesterPage.re
```diff
@@ -10,67 +10,45 @@ let styles =
...
-    noSpacer: bool,
-    children: list ReactRe.reactElement,
-    title: option string
-  };
-  let contentWrapper props children =>
-    if props.noScroll {
-      <View key="wrapper" style=styles##wrapper> (ReactRe.listToElement children) </View>
+let component = ReasonReact.statelessComponent "RNTesterPage";
+
+let make ::title=? ::noScroll=false ::noSpacer=false children => {
```
wokalski: disclaimer: I don't necessarily think it should be changed

Why do you use components instead of plain funcs? For JSX?

src/components/imageRe.re
saschatimme: grr

RNTester/src/RNTesterPage.re
saschatimme: Thought, that it's nice with the JSX in the other component. And at least with the new API its not that much difference any more :)

src/components/textRe.re
saschatimme: The `children` which are passed to `make` are now `array reactElement` and no more `list reactElement`

saschatimme:
some updates

wokalski:
Bravo 👏

Issue #11, opened 2017-05-09
Revisit SectionList API


saschatimme:
See discussion in the PR https://github.com/BuckleTypes/bs-react-native/pull/10

Still not completely happy with the current approach, but better as the initial one. This issue is here to remind us on revisiting the API.

saschatimme:
Just merged #37

PR #37, opened 2017-06-27
SectionList API

saschatimme:
Adressing #11.
I wasn't really happy with the SectionList API, especially since Reason-React isn't functor based anymore.
Another thing which came in mind: We conversted from records to `Js.t 'a` in render, but this also created a new function each time, i.e. quite bad for performance. I now moved all conversions to separate functions and made the prop types abstract. [Here](https://github.com/BuckleTypes/bs-react-native/blob/5ca9819929e8a2900d0cc88a86e60889030a9d30/RNTester/src/RNTesterExampleList.re#L26) is an example.

One problem we ran into earlier was that we could deconstruct the `renderBag` but then the compiler couldn't access `item.foo`. I now learned that you can qualify things with `item.MyModule.foo` and everything works superb :)

### DIFF
--- a/RNTester/src/RNTesterExampleList.re
+++ b/RNTester/src/RNTesterExampleList.re
@@ -21,37 +21,36 @@ let styles =
       }
     );
 
-include
-  SectionList.CreateComponent {
-    type item = ExampleList.item;
-  };
-
 let component = ReasonReact.statelessComponent "RNTesterExampleList";
 
-let make = {
-  let renderItem onPress {item} =>
-    <TouchableHighlight onPress=(fun () => onPress item)>
-      <View style=styles##row>
-        <Text style=styles##rowTitleText> (ReasonReact.stringToElement item.title) </Text>
-        <Text style=styles##rowDetailText> (ReasonReact.stringToElement item.description) </Text>
-      </View>
-    </TouchableHighlight>;
-  let itemSeparatorComponent {highlighted} =>
-    <View
-      style=(
-              if highlighted {
-                styles##separatorHighlighted
-              } else {
-                styles##separator
-              }
-            )
-    />;
-  fun ::components ::onPress _children => {
+let renderItem onPress =>
+  SectionList.renderItem (
+    fun {item} =>
+      <TouchableHighlight onPress=(fun () => onPress item)>
+        <View style=styles##row>
+          <Text style=styles##rowTitleText>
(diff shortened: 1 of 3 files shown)


RNTester/src/RNTesterExampleList.re
```diff
@@ -21,37 +21,36 @@ let styles =
...
+          <Text style=styles##rowTitleText>
+            (ReasonReact.stringToElement item.ExampleList.title)
+          </Text>
+          <Text style=styles##rowDetailText> (ReasonReact.stringToElement item.description) </Text>
+        </View>
+      </TouchableHighlight>
+  );
+
+let itemSeparatorComponent =
+  SectionList.separatorComponent (
```
wokalski: do these functions exist to make inferring easier?

saschatimme: I don't think so, they are there because of the conversion problematic I tried to explain above :D
As you know, for good performance in React functions as props should be memoized. But we did conversions from record to jsObject in render, i.e. created a new function each time! My idea with these functions was to make this conversion step more explicit and **outside** render, so that things could get memoized if necessary.

saschatimme: Does this makes sense? :D

wokalski: The conversion step is still inside render, just in another component. 😛

saschatimme: That's true, but there will not be a new function generated in each render, i.e. for the underlying JS component it is always the same function

wokalski: yeah, that's true. Good change. I forgot how it was done, had to check.

wokalski:
Merge when ready then!

saschatimme:
Another thought:
I made types like `renderItem 'item = jsRenderBag 'item => ReasonReact.reactElement` abstract. But we could expose them for performance reasons, if one wants to avoid all conversions. But this would obviously make the API surface less clear :/
What do you think?

wokalski:
What's the performance hit on a list with 1000 rows?

saschatimme:
Don't know, but we can always expose the raw interface later, so maybe we just move this train of thought until someone complains about perf?

wokalski:
I agree. I think (hope) the perf hit is not significant.

saschatimme:
Yeah, but you know, people do crazy things^^

Will then merge everything

Issue #12, opened 2017-05-10
Standarize polymorphic API convention


wokalski:
Currently some of our APIs use suffix based convention (e.g. `diffClampPct` and `diffClamp`) . Others use polymorphic variants. e.g. here:

```
  let interpolate:
    t =>
    inputRange::list float =>
    outputRange::[< | `float (list float) | `string (list string)] =>
    ...
```

It would be great to somehow standardise the naming convention and probably write it down somewhere.

saschatimme:
I think we should explore GADTs outside of style also, with this we could avoid/reduce the use of `Obj.magic` or the `Encode.` which is still somewhere.

wokalski:
This has been solved for styles. There are some remaining uses for instance in animation APIs.

MoOx:
We are going to do our best in next rewrite

PR #39, opened 2017-06-27
Style [GADT Proof of concept]

saschatimme:
Wanted to show you this to address #12 (I think the inconsistency is mostly `Style` vs. the rest).

If we start to type things using GADTs we can achieve things like this

```reason
let pt = Style.bottom Pt 4.;

let pct = Style.bottom Pct 4.;

let animated = Style.bottom Animated (Animated.Value.create 1.0);
```

The neat thing is that everything is like using normal variants, **but** we don't need any additional parens. I think this would avoid a lot of a pain until https://github.com/facebook/reason/pull/953 lands

### DIFF
--- a/src/styleRe.re
+++ b/src/styleRe.re
@@ -16,6 +16,12 @@ type style =
   | AnimatedStyle
       string [ | `value AnimatedRe.Value.t | `interpolation AnimatedRe.Interpolation.t];
 
+type lengthOrAnimated _ =
+  | Pt :lengthOrAnimated float
+  | Pct :lengthOrAnimated float
+  | Animated :lengthOrAnimated AnimatedRe.Value.t
+  | Interpolated :lengthOrAnimated AnimatedRe.Interpolation.t;
+
 let combine a b => {
   let entries =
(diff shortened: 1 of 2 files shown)


wokalski:
This is soo sweet. :shipit:

wokalski:
I don't understand though. :(. Why don't GADTs require parens? We are not using any property of GADTs here, besides the way they are parsed?

saschatimme:
They also don't require to be in the namespace, i.e. the example above works although the GADT is defined inside `Style` 🤔.

But the parens thing makes sense: They are a separate argument to the function, i.e. you have `let bottom: lengthOrAnimated 'a => 'a => style;` So `Pt` is of type `lengthOrAnimated float` which fixes the second argument.

wokalski:
omg now I get it!! So in this example `4.` is an argument to `bottom` and `GADT`'s value is just `Pt`? Now I get it, and this is amazing.

```
let pt = Style.bottom Pt 4.;
```

wokalski:
I'm gonna implement this the moment I start building the app (later this week/at the end of next week).

PR #40, opened 2017-06-27
Add Linking

saschatimme:
🔥

### DIFF
--- a/src/linkingRe.re
+++ b/src/linkingRe.re
@@ -1,0 +1,17 @@
+external openURL : string => Js.Promise.t unit =
+  "openUrl" [@@bs.scope "Linking"] [@@bs.module "react-native"];
+
+external _canOpenUrl : string => Js.Promise.t Js.boolean =
+  "canOpenUrl" [@@bs.scope "Linking"] [@@bs.module "react-native"];
+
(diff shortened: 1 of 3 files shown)


src/linkingRe.re
```diff
@@ -0,0 +1,17 @@
+external openURL : string => Js.Promise.t unit =
+  "openUrl" [@@bs.scope "Linking"] [@@bs.module "react-native"];
+
+external _canOpenUrl : string => Js.Promise.t Js.boolean =
+  "canOpenUrl" [@@bs.scope "Linking"] [@@bs.module "react-native"];
+
+let canOpenUrl url =>
+  _canOpenUrl url |> Js.Promise.then_ (fun bool => Js.Promise.resolve (Js.to_bool bool));
```
wokalski: Nothing important but I used to do this, too until I noticed that you can simply create another function with the same name. It has the benefit that _canOpenUrl is shadowed when generating `rei`

saschatimme: Ah cool, but at the same time confusing :D

wokalski: yeah, I mainly use it in my apps code to shadow utility functions.

PR #43, opened 2017-07-02
Make TouchableOpacity styleable

kennetpostigo:
TouchableOpacity was missing style prop, making it unable to be styled.

### DIFF
--- a/src/components/touchableOpacityRe.re
+++ b/src/components/touchableOpacityRe.re
@@ -11,6 +11,7 @@ let make
     ::delayPressOut=?
     ::disabled=?
     ::hitSlop=?
+    ::style=?
     ::onLayout=?
     ::onPress=?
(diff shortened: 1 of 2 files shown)


kennetpostigo:
I suspect this is missing in other components. I'll send more PRs for those when I run into them.

saschatimme:
Good catch, it's probably missing since it doesn't occur in the docs (so much fun).
Could you add style also to the `*.rei`?

kennetpostigo:
@saschatimme it seems style is already a prop in the rei:
https://github.com/kennetpostigo/bs-react-native/blob/patch-1/src/components/touchableHighlightRe.rei#L37

kennetpostigo:
whoops, wrong component 😓 Touchable* lol. I'll add the style to the rei for TouchableOpacity.

PR #46, opened 2017-07-06
Unknown prop accessibleLeft -> accessibilityLabel

arnarthor:
I'm not sure where the accessibleLeft prop came from, but it seems to be spread all over the library. 

I tried searching for this prop in the react-native codebase but didn't find any results so my guess is that this was a typo at some point + copy paste after that.


saschatimme:
Good catch! Don't know what I did there, I probably read whatever I wanted to read...

PR #47, opened 2017-07-07
Adding WebView

arnarthor:
Adding webView bindings. I followed a lot of the patterns seen in View and Image bindings. 

This adds support for all documented React Native props to WebView and has a super simple example of WebView in the RNTester app as well.

Screenshot from RNTester WebView example directing to https://example.com
image


src/components/webViewRe.re
```diff
@@ -0,0 +1,289 @@
+module type WebViewComponent = {
```
saschatimme: Can you extract the content into `webViewRe.rei`? The component doesn't need to be a functor. The reason that `View` is a functor, is that there is also `Animated.View` where we reuse the implementation.

src/components/webViewRe.re
```diff
@@ -0,0 +1,289 @@
...
+    decelerationRate::list [ | `normal | `fast | `value float]? =>
+    onShouldStartLoadWithRequest::(webViewIOSLoadRequestEvent => bool)? =>
+    scrollEnabled::bool? =>
+    array ReasonReact.reactElement =>
+    ReasonReact.component ReasonReact.stateless ReasonReact.noRetainedProps;
+};
+
+module type Impl = {let webView: ReasonReact.reactClass;};
+
+module CreateComponent (Impl: Impl) :WebViewComponent => {
```
saschatimme: You can remove the functor (see above)

src/components/webViewRe.re
```diff
@@ -0,0 +1,289 @@
...
+        | `pageTurn
+      ]? =>
+    accessibilityViewIsModal::bool? =>
+    shouldRasterizeIOS::bool? =>
+    automaticallyAdjustContentInsets::bool? =>
+    contentInsets::webViewContentInsets? =>
+    injectJavaScript::(string => unit)? =>
+    injectedJavaScript::string? =>
+    mediaPlaybackRequiresUserAction::bool? =>
+    onError::(webViewNavState => unit)? =>
```
saschatimme: At the moment is `webViewNavState` an abstract type, so you cannot do anything with it.
What you actually want is that onError has the signature
```reason
Js.t {. target: int, canGoBack: Js.boolean, lockIdentifier: int, ...} => unit
```

src/components/webViewRe.re
```diff
@@ -0,0 +1,289 @@
...
+    domStorageEnabled::bool? =>
+    javaScriptEnabled::bool? =>
+    mixedContentMode::list [ | `never | `always | `compatibility]? =>
+    thirdPartyCookiesEnabled::bool? =>
+    userAgent::string? =>
+    allowsInlineMediaPlayback::bool? =>
+    bounces::bool? =>
+    dataDetectorTypes::list [ | `phoneNumber | `link | `address | `calendarEvent | `none | `all]? =>
+    decelerationRate::list [ | `normal | `fast | `value float]? =>
+    onShouldStartLoadWithRequest::(webViewIOSLoadRequestEvent => bool)? =>
```
saschatimme: same as for `onError` etc

src/components/webViewRe.re
```diff
@@ -0,0 +1,289 @@
...
+    url::string? =>
+    title::string? =>
+    loading::bool? =>
+    canGoBack::bool? =>
+    canGoForward::bool? =>
+    unit =>
+    webViewNavState =
+    "" [@@bs.obj];
+  type webViewIOSLoadRequestEvent;
+  external webViewIOSLoadRequestEvent :
```
saschatimme: I think this is the wrong way. `onShouldStartLoadWithRequest` accepts a cb (see above) which has an `Js.t {.…}` as input.
Now there are two ways to handle it.
1) Expose the Js.t object in the types, i.e. the signature I commented above
2) Expose a corresponding record type (say `iOSLoadRequestEvent`) and a function which takes as argument a function `iOSLoadRequestEvent  => unit`  and returns a function `Js.t {..} => unit`. See the discussion in https://github.com/BuckleTypes/bs-react-native/pull/37 for the some additional explanation in the case of SectionList.

src/components/webViewRe.re
```diff
@@ -0,0 +1,289 @@
...
+    scrollEnabled::bool? =>
+    array ReasonReact.reactElement =>
+    ReasonReact.component ReasonReact.stateless ReasonReact.noRetainedProps;
+};
+
+module type Impl = {let webView: ReasonReact.reactClass;};
+
+module CreateComponent (Impl: Impl) :WebViewComponent => {
+  type webViewNavState;
+  external webViewNavState :
```
saschatimme: see below

src/components/webViewRe.re
```diff
@@ -0,0 +1,289 @@
...
+    lockIdentifier::int =>
+    loading::bool =>
+    title::string =>
+    canGoForward::bool =>
+    navigationType::[ | `other | `click] =>
+    url::string =>
+    unit =>
+    webViewIOSLoadRequestEvent;
+  type webViewSource;
+  let webViewSource:
```
saschatimme: I think `source` is enough since you would access it with `WebView.source`, so `WebView.webViewSouce` would be kinda redundant

wokalski:
Awesome, thank you @arnarthor

PR #49, opened 2017-07-31
Style.flatten

saschatimme:
This is equivalent:

Javascript:
```jsx
<View style={[styleA, styleB]} />
```
Reason:
```reason 
<View style=(Style.flatten [|styleA, styleB|]) />
```

### DIFF
--- a/src/styleRe.rei
+++ b/src/styleRe.rei
@@ -1,5 +1,14 @@
 type t;
 
+/** Coerces an array of styles to a single style.
+ * This is equivalent:
+ * // js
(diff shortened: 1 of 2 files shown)


src/styleRe.rei
```diff
@@ -1,5 +1,14 @@
 type t;
 
+/** Coerces an array of styles to a single style.
```
wokalski: Let's use simple vocabulary in comments. How about `Generates a style out of array of styles` or something?

saschatimme: Yes sir

PR #50, opened 2017-07-31
Added TabBarIOS and TabBarItemIOS

lpalmes:
Added missing components for my own use and i thought they may come in handy for somebody else.


src/components/tabBarIOSRe.rei
```diff
@@ -0,0 +1,10 @@
+let make:
+  barTintColor::'a? =>
```
saschatimme: Could you fix the types here?

src/components/tabBarItemIOSRe.rei
```diff
@@ -0,0 +1,13 @@
+let make:
+  title::string? =>
+  badgeColor::string? =>
+  badge::'a? =>
```
saschatimme: here also

saschatimme:
Really cool! We just need to fix the `.rei`s and we are good to got :)

lpalmes:
I've added types, one thing that i found out is that it might be a good idea to extract imageSource from the ImageRe module to have other modules use it like this one, let me know your opinion about it. Thanks!

saschatimme:
This looks good now :)
I think it would make sense to extract imageSource but we should do this separate since this occurs on other places as well. I will open a separate issue to discuss the best approach.

PR #52, opened 2017-08-02
Fix interface implementation mismatch for tabBarIOS

gunnigylfa:
Ran into a problem compiling the RNTester project. It was failing because the implementation for `tabBarIOS` did not match the interface for it. This tiny PR fixes it.


saschatimme:
Oh this slipped through in the last PR :( thanks for fixing it!

PR #53, opened 2017-08-03
Add travis-ci

### LINKED ISSUE #18: Continuous Integration
Developing an OSS project is not too sustainable without some kind of Continuous Integration. I'm not sure how to approach it because FFI by definition is hard to test. Maybe some integration tests using the UI Explorer app?

saschatimme:
This should hopefully catch all compile errors at least :)

- [x]  I still have to setup travis-ci (I don't have proper rights atm).

Closes #18

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,10 @@
+language: node_js
+node_js:
+  - "7"
+
+script:
+  - npm run build
+
+cache:
+  directories:
+      - node_modules
--- a/README.md
+++ b/README.md
@@ -105,4 +105,4 @@ let styles =
 ```
 
 ## Status
-You can find an overview of the implemented components and APIs [here](STATUS.md)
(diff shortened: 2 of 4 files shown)


wokalski:
We should build `RNTester`, too.

saschatimme:
Not sure how to configure travis, because we need a second npm install etc.

arnarthor:
You could of course introduce something like Lerna to have RNTester be a separate package, which would then never be published while the src code would be. 

That way you could easily run build for both RNTester and the lib itself. But it of course adds overhead to the project.

saschatimme:
I think we can avoid to publish RNTester if we just add a `.npmignore`. I'm not sure Lerna works for us since bucklescript can only build files which are at the same level as the root or below and note above.

saschatimme:
Let's move the discussion to #54

PR #58, opened 2017-08-14
Fixed bindings for the Linking module.

mbirkegaard:
This fixes wrong capitalization in the Linking module, which previously bound to the nonexistent openUrl, canOpenUrl, and getInitialUrl methods.

### DIFF
--- a/src/linkingRe.re
+++ b/src/linkingRe.re
@@ -1,14 +1,14 @@
 external openURL : string => Js.Promise.t unit =
-  "openUrl" [@@bs.scope "Linking"] [@@bs.module "react-native"];
+  "openURL" [@@bs.scope "Linking"] [@@bs.module "react-native"];
 
-external _canOpenUrl : string => Js.Promise.t Js.boolean =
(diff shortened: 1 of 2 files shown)


wokalski:
thank you very much, this looks great. While we're on it though, could you change the return type of `getInitialURL` to `Js.Promise.t (Js.Null.t string)`. It's incorrect now according to [this](https://github.com/facebook/react-native/blob/91493f6b9d5013b4d49c9d6418af00603fde023f/Libraries/Linking/Linking.js#L210).

mbirkegaard:
Done!

Though, while we're at it wouldn't we want to make it return `Js.Promise.t option string` (similar to `_canOpenURL`)?

If yes, I'll make the necessary change.

saschatimme:
I think that makes sense

wokalski:
Yes. We probably would

mbirkegaard:
That should do it.

wokalski:
Great. Thank you!

PR #68, opened 2017-09-04
Add separate funcs for native driver animations

wokalski:
When a value is animated with the native driver
you cannot pass easing to it (even if it's undefined).

---

We could have it verified statically but it'd require a rewrite of the whole API which notably does require rewrite because it's awfully verbose. There'd have to be `NativeValue`, `NativeValueXY`, and the derivatives in addition to the existing modules

### DIFF
--- a/src/animatedRe.re
+++ b/src/animatedRe.re
@@ -16,16 +16,16 @@ module CompositeAnimation = {
 
 module Animations = {
   module Decay (Val: Value) => {
-    type config =
-      Js.t {
-        .
-        velocity : Val.rawJsType,
-        deceleration : Js.undefined float,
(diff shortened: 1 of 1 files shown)


saschatimme:
Couldn't you use an external instead of `bs.obj` to create the objects? Then all omitted values wouldn't be in the generated JavaScript object and RN would be happy

wokalski:
Damn. That's right. I should've. I forgot about externals.

wokalski:
I need to check it but I had a shower thought that I cannot because the argument will still be passed (just with null value) so in the result the object will still have the easing property

saschatimme:
No I think that was a bug and is now fixed https://github.com/BuckleScript/bucklescript/issues/1409

PR #69, opened 2017-09-05
Add image background component

gunnigylfa:
Added the <ImageBackground/> component which is intended for having nested components Inside of an Image. 
RN docs state that:

> A common feature request from developers familiar with the web is background-image. To handle this use case, you can use the  <ImageBackground> component, which has the same props as <Image>, and add whatever children to it you would like to layer on top of it.

There is a slight difference between props are I have added them to this. But other than that I just copied the ImageRe interface and implementation.

Edit: I reimplement these bindings so they share common props as much I could and got rid of the functor so there is far less duplicate code.

Here is an example of the component being used as a background image with a <Text/> as it's child.

image


src/components/imageBackgroundRe.re
```diff
@@ -0,0 +1,182 @@
...
+      | `auto => "auto"
+      | `resize => "resize"
+      | `scale => "scale"
+      };
+    let encodeDefaultSource (x: defaultSource) =>
+      switch x {
+      | URI x => rawImageSourceJS x
+      | Required x => rawImageSourceJS x
+      };
+    let make
```
wokalski: This module should only contain `make`.

We don't need a functor unless there exists `Animated.ImageBackground`. The other types and utility functions should be the same as in `ImageRe.re`. If they are unaccessible due to `reis` feel free to put them in a separate file. Like `ImageCommonRe.re`.

wokalski:
Ah. sorry for confusion. The `ImageRe` functor won't work because of different props. We need to keep your code, however it'd be great to eliminate the duplication.

gunnigylfa:
@wokalski  Yeah I just realized that when trying to fix the RNTester example once I changing it.  I might try to extract the two interfaces two a common one once I have gotten more familiar with the workings of reason. Would this implementation be sufficient for now?

arnarthor:
I agree that less duplication would be great, but should we merge this for now and try to optimize later. Keeping this open too long might make everyone lose context when coming back to this.

wokalski:
And merging it will introduce duplication and simply wrong code. The functor is not necessary here and the types which are the same as in Image should simply reference the types in Image. Otherwise the types become different and you cannot pass the same image from `Image` to `ImageBackground` because types are nominal

arnarthor:
Riiight, I misunderstood your comment. Thought you had taken the review back with the second comment.

I agree, that all makes sense 👍. @gunnigylfa I can help if needed as well

gunnigylfa:
Ah okay I thought that you meant keep the code as it was but thanks for the clarification @wokalski  I will try and fix this as soon as I can, I will probably take you up on that offer @arnarthor  😄

gunnigylfa:
@wokalski @arnarthor I managed to get this working again with a different implementations of the bindings. Now I'm not using the functor and I'm sharing types with `ImageRe` so an Image can be sent into a ImageBackground just as it could be sent into an Image.

wokalski:
Awesome. It looks good. Sorry for nitpicking!

gunnigylfa:
No worries @wokalski I appreciate the feedback 🙂

PR #72, opened 2017-09-22
Fixes onScroll event type on ScrollView

kristinns:
Current implementation of _ScrollView.onScroll_ takes a callback function with event of type `RNEvent.NativeEvent.t` where every property of the callback event is `undefined`.

This PR fixes that by replacing the `onScroll` event type with a new module type `RNEvent.NativeScrollEvent` which coordinates with the implementation found [here](https://github.com/facebook/react-native/blob/master/React/Views/RCTScrollView.m#L87).

### DIFF
--- a/src/RNEvent.re
+++ b/src/RNEvent.re
@@ -16,15 +16,56 @@ module NativeEvent = {
 
 module NativeLayoutEvent = {
   type t;
-  type layout = {x: float, y: float, width: float, height: float};
-  external _layout : t => Js.t 'a =
-    "nativeEvent" [@@bs.get];
+  type layout = {
+    x: float,
+    y: float,
+    width: float,
+    height: float
+  };
+  external _layout : t => Js.t 'a = "nativeEvent" [@@bs.get];
   let layout (t: t) => {
     let l = (_layout t)##layout;
     {x: l##x, y: l##y, width: l##width, height: l##height}
   };
 };
 
+module NativeScrollEvent = {
+  type t;
+  type point = {
+    x: float,
+    y: float
+  };
+  type size = {
+    width: float,
+    height: float
+  };
+  type contentInset = {
+    bottom: float,
+    top: float,
+    left: float,
+    right: float
+  };
+  external _nativeEvent : t => Js.t 'a = "nativeEvent" [@@bs.get];
+  let contentOffset (t: t) => {
+    let co = (_nativeEvent t)##contentOffset;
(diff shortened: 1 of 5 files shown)


src/RNEvent.re
```diff
@@ -16,15 +16,56 @@ module NativeEvent = {
...
+    let lm = (_nativeEvent t)##layoutMeasurement;
+    {width: lm##width, height: lm##height}
+  };
+  let contentInset (t: t) => {
+    let ci = (_nativeEvent t)##contentInset;
+    {bottom: ci##bottom, top: ci##top, left: ci##left, right: ci##right}
+  };
+};
+
 external nativeEvent : t => NativeEvent.t = "" [@@bs.get];
```
saschatimme: Isn’t this wrong and the reason you got undefined?

saschatimme:
LGTM, only thing I noticed that we defined `contentInset` again, there is also currently [`TypesRN.insets` widely used](https://github.com/reasonml-community/bs-react-native/search?utf8=✓&q=TypesRN.insets&type=).  But maybe we should unify these at a later point...
Any thoughts @arnarthor @wokalski ?

wokalski:
`WebView` also has a separate `contentInsets` (mess). That said it would be cool to make the type the same.

arnarthor:
I think it's out of scope for this PR and since we have the regression already it might make sense to let this through and then create a separate PR that unifies all of these inset types.

kristinns:
Ahh I didn't know that type existed, I can use that instead.
But that will cause some inconsistency because the `TypesRN.insets` is a Js type but `size` and `point` are reason types. See the example below
```jsx
<ScrollView
  onScroll=(
    fun event => {
      let contentInsetLeft = (RNEvent.NativeScrollEvent.contentInset event)##left;
      let contentOffsetX = (RNEvent.NativeScrollEvent.contentOffset event).x;
      ()
    }
 )>
...
</ScrollView>
```
What's your take on that? I see that this is the same for [ScrollView.contentOffset](https://github.com/reasonml-community/bs-react-native/blob/master/src/components/scrollViewRe.re#L70). and [ScrollView.contentInset](https://github.com/reasonml-community/bs-react-native/blob/master/src/components/scrollViewRe.re#L69)

wokalski:
ok, it makes me a bit sad but screw it

saschatimme:
I think we need to have a broader discussion about the use of `Js.t` vs records, but this is definitely out of scope of this PR.

PR #75, opened 2017-09-25
Fixes wrong binding for processColor

kristinns:
Fixes
`Unable to resolve module 'processColor' from .../bs-react-native/src/styleSheetRe.js': Module does not exist in the module map`


saschatimme:
Thanks!

PR #76, opened 2017-09-27
Add local image support to imagebackground

gunnigylfa:
Adding missing type for `imageSource` in `ImageBackground`   
Now it can handle using local assets in `ImageBackground` just like we can with the `Image` component.

Example usage would be: 
```ocaml
<ImageBackground
        /* all the other props here */
        source=(Required (Packager.require "../static/img/image.jpg"))>
</ImageBackground>
```

This PR also fixes broken imports in the RNTester app which broke after the capitalization of the compiled .js files

### DIFF
--- a/RNTester/index.android.js
+++ b/RNTester/index.android.js
@@ -4,10 +4,8 @@
  * @flow
  */
 
-import { comp as App } from "../lib/js/RNTester/src/rNTesterApp.js";
-import React from "react";
-import {
-  AppRegistry
-} from 'react-native';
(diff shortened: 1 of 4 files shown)


saschatimme:
Could you explain why we need a new `imageSource` type and cannot reuse the one from `Image`?

gunnigylfa:
@saschatimme I was experiencing problems with using local images with the current implementation and thought required images did not work but I was mistaken this PR is quite unncessary sorry about that. Will close it.

PR #79, opened 2017-10-16
Fixing types of maximumValue and minimumValue from Slider component

Denommus:
When using string, the runtime gives me a dynamic type error as the
following: "TypeError:expected dynamic type 'double',but had type
'string'"

This commit fixes this problem.


saschatimme:
Thanks!

PR #83, opened 2017-10-26
Add props to FlatList

arnarthor:


saschatimme:
I think there is nothing in it what could trigger a runtime error (as we had once previously). So this should be fine.

wokalski:
if I have a prop and it has no effect it's also a runtime error, for the user.

arnarthor:
These props all come from either VirtualizedList or ScrollView, which makes me think if FlatList should be functor based, but either way I'm pretty sure these props have been there for a pretty long time, just not documented under the FlatList section.

wokalski:
It's good to merge if they were in RN `^0.46.0`

arnarthor:
Awesome I'll check and merge if that's the case 👍

arnarthor:
Verified that this was all in React-Native 0.45.x 👍 merging this.

I don't have permission to publish new versions, this is not urgent on our end though, we've now vendored this into our project for simpler contributions and support for ourselves.

saschatimme:
@arnarthor I made you an admin here and make you an owner at npm. So all the power is now also in your hand 💪

PR #84, opened 2017-10-26
Tab bar item bug

wokalski:
Fixes aliasing to `TabBarItemIOSRe` which doesn't exist


saschatimme:
Any idea why the compiler didn't complain? :O

wokalski:
it does complain ;)

saschatimme:
Haha, maybe we should configure Travis more aggressive? Could we make all warnings to errors?

PR #86, opened 2017-10-29
Upgrade to Reason 3 and enable namespaces

arnarthor:
This covers the changes me and @wokalski discussed on discord regarding upgrades for namespaces, Reason 3 and changing reason-react to a peer dependency.

Do you have permission to submit under @reasonml-community @saschatimme? I might also be mistaken on how organization namespaced packages work, and I wasn't sure if the bsconfig name should be `react-native` or `@reasonml-community/react-native`. I thought it made sense that it should be `react-native` based on my knowledge on namespaces.

Correct me if I'm wrong here 👍 

Regarding the review it's probably wise to read through the last 2 commits separately. The first commit is just the automated script to update the syntax.


saschatimme:
I don’t think I have the permissions. @chenglou could you help with that?

chenglou:
Just added both of you to the npm reasonml-communtity org =)

wokalski:
So the next task now is removing the ReactNative module and hiding the private modules.

arnarthor:
Yeah that sounds good @wokalski 👍 

I can work on that tonight unless you want to take it

arnarthor:
We're already vendoring this in our project so I can then make sure everything works properly as well before publishing 👌

wokalski:
We’re vendoring it too. I’m going to sleep so feel free to take it.

PR #92, opened 2017-11-19
Add Troubleshooting section to README.md

rrdelaney:
I had some problems getting started with create-react-native-app so I figured I'd add a troubleshooting section in case anyone else had the same problems 😄


wokalski:
Thank you!

PR #97, opened 2017-11-22
Check in JS sources

wokalski:
Thanks to checking in JS sources it will be easier to reason about performance.


MoOx:
Looks like a bad idea imo to check in the sources. For example I just took master, run the build, and I am getting plenty of changes. Mostly camel case changes.
What is the real motivation behind this PR? I don't see how versionning js will help to mesure perf. We can always build previous js by just using a older commit?

wokalski:
You can take a look at the diff during PR.

MoOx:
FYI, I took master and I am getting this diff in all files

screen shot 2017-12-07 at 11 50 51

I am on macOS (APFS non sensitive to camel case). Any idea how we could avoid this kind of diff?

wokalski:
It's correct - capitalisation has changed in the meantime.

MoOx:
Ok. What do you think about adding automatic changes in js via husky/precommit (automatic git hook setup on "npm install")?

wokalski:
I don't follow. Pre commit hook doing what?

MoOx:
Precommit doing:
- npm run build
- check in diff to js automatically so nobody will ever "miss" to update JS files.

wokalski:
I'd rather have [Danger](http://danger.systems/js/) or something similar set up checking the pull requests. Long running pre commit hook is a no no IMO.

MoOx:
To be honest, clean + build take around 2s on my computer, I guess it should not be really different (and I counted the clean, I think it's not necessary so build will be even faster). It will save us time at the end cause on the medium/long run we will never have to manually update, which will take much more time "oh danger told you to do X, so please do it; - oh ok, so I just ran "Y" and update files?", - yes please". This will at the end be always longer compared to a 1s magical precommit hook imo!

This solution:

- Don't require doc
- don't require human action for PR

wokalski:
what? You always need human action in PRs and you need CI anyway. You cannot depend on somebody's local circumstances. Pre commit hooks are also very annoying when you know something is broken but you want to commit it. Not a fan.

MoOx:
Sure that PR require human interaction, but it's better to get something updated when it can be, rather than having to check that the change has been made (or wait for Danger to tell us) and tell people to made a change.
I use precommit hook daily for very small tasks like linting/formatting and minor task like this and I love it, but I totally understand that you are not interested.

I will add  that it's *very* easy to disable a precommit hook setup with husky (I did this several times, cause, yeah sometimes I commit broken stuff haha).
You just have to rename or delete the script "precommit" (since husky rely) on this, which is realllly nice (you don't have to touch .git files!) and the hook will be skipped!

For example in the package.json

```js
{
  "scripts": {
    "precommit": "lint-staged",
  },
  "lint-staged": {
    "*.re": "npm run build && git add -A js"
  }
}
```

This way, precommit is taking time only when a .re file is being modified.
If you want to disable the precommit hook for a commit, just rename "precommit" to "whatever" or just "precommit-lolno", make your commit (and do not stage this specific tiny change ^^) and you won't have any trouble.

If you want, we could try for a few commits/PR, and if you find that's a pain, we rollback the changes.
Otherwise, just tell me and I won't bother you with this for a while.

wokalski:
I'm saying it's a pain because I've used pre commit hooks in the past. I don't want to single handedly reject this suggestion. What do you think? @cullophid @arnarthor @gunnigylfa @kristinns (pinging recent contributors)

arnarthor:
I like having the compiled code checked in. It's great to catch bugs where bindings have gone wrong.
Pre commit hooks IMO are definitely a huge red flag!

wokalski:
what do you mean @arnarthor ? Do you mean that a failing pre commit hook is a huge red flag or that you don't like them 😛 ?

arnarthor:
I hate precommit hooks 😄

arnarthor:
I would much rather (like @wokalski suggested) have Danger set up in CI to warn about these things

MoOx:
I guess we could also setup a simple travis command too.

PR #98, opened 2017-11-23
Add DatePickerAndroid API

tonyhb:
Adds the DatePickerAndroid API bindings.


src/datePickerAndroid.re
```diff
@@ -0,0 +1,57 @@
...
+};
+
+[@bs.scope "DatePickerAndroid"] [@bs.module "react-native"]
+external _open : optsJs => Js.Promise.t(responseJs) =
+  "open";
+
+let open_ = (~date: Js.Date.t, ~minDate=?, ~maxDate=?, ~mode=Default, ()) =>
+  _open({
+    "date": date,
+    "minDate": Js.Undefined.from_opt(minDate),
```
arnarthor: I'd go with `Js.Nullable` here it covers both null and undefined but the from_opt value get's compiled down to `undefined`

src/datePickerAndroid.re
```diff
@@ -0,0 +1,57 @@
...
+      | Calendar => "calendar"
+      | Spinner => "spinner"
+      }
+  })
+  |> Js.Promise.then_(
+       (resp: responseJs) => {
+         let an = resp |> action;
+         Js.Promise.resolve(an)
+       }
+     );
```
arnarthor: This is real nitpicky, but I like this style of Promise based flow better 🙈 

```ocaml
Js.Promise.(
  _open({
    "date": date,
    "minDate": Js.Undefined.from_opt(minDate),
    "maxDate": Js.Undefined.from_opt(maxDate),
    "mode":
      switch mode {
      | Default => "default"
      | Calendar => "calendar"
      | Spinner => "spinner"
      }
  })
  |> then_((resp: responseJs) => resp |> action |> resolve)
);
```

src/datePickerAndroid.re
arnarthor: I'm also pretty sure `Js.Undefined` and `Js.Null` are getting deprecated in favour of `Js.Nullable` soon

src/datePickerAndroid.rei
```diff
@@ -0,0 +1,24 @@
+[@bs.scope "DatePickerAndroid"] [@bs.module "react-native"] external dateSet : string =
```
wokalski: The convention in this database is to put externals as let bindings in `rei`s. So in this case it'd be `let dateSet : string;` etc.

wokalski: `externals` are exposed in `rei`s as let bindings. In this case it'd be `let dateSet : string;`

src/datePickerAndroid.re
tonyhb: :) thanks!

src/datePickerAndroid.rei
tonyhb: thanks :)

tonyhb:
Thanks for the updates!  Just heading back from Thanksgiving - will amend and push tonight or tomorrow AM

arnarthor:
@tonyhb did you forget about this one 😄

tonyhb:
Haha sorry, got busy at work!  Jumping on it now

tonyhb:
Fixed - didn't rebase so that the diff is visible.  We can rebase when it merges if necessary. :)

Thanks again for the input - good to learn about externals in REI for example.  Sorry it took a while!

tonyhb:
Should I move reason-react from a peer dependency or change the CI script to explicitly install it?

arnarthor:
No we don't want to move it from peer dependencies. It's essential for Reason projects to not have hard depedencies.

Fixing the CI script would be better

MoOx:
See #105 for resolution for the issue you are facing. I think #105 should be merged asap, and then this PR could be rebased on master.

MoOx:
A fix landed in master, @tonyhb you can just rebase and it should be good.

wokalski:
JS sources are still missing. The rest is great!

wokalski:
@arnarthor @MoOx this one is done, too.

wokalski:
@arnarthor please remove the last merge commit

wokalski:
nvm, I'll take care of it

wokalski:
@arnarthor it's ready.

wokalski:
In my usual fashion I took this one a bit too far. I'll revert 0a0f944

arnarthor:
It can't automatically merge 🤔 do you know why?

wokalski:
you have to accept it 👍

arnarthor:
Ahh right, I accepted it before you reverted a commit 👍

PR #101, opened 2017-11-24
Making the RNEvent api public

gunnigylfa:
Exposing the RNEvent api to the public (edit: As pointed out by @arnarthor  it was never public in the first place). Making it possible to manipulate UI based on scroll interactions.


arnarthor:
So this was never public initially (to clarify, it wasn't part of the `reactNative.re` file based namespace), it just happend to be so because of OCamls module system.

What was the Reason for this @wokalski. I noticed this isn't documented in the React Native docs either so I have no idea what the status of this API is.

MoOx:
It's also interesting in some edges cases to get private to be accessible.
I am currently in the process of making an app/website with the same reason/rn codebase thanks to https://github.com/necolas/react-native-web (which works great by itself, I used RNW more than once for web only or mobile/web projects)).
In some cases, RNW allows web only props. In practise, `Text` accepts an `href`. That's just an example and they might be more (another one: https://github.com/necolas/react-native-web/blob/034108a2a07e73e2de95824cb91f65275e3e3c85/src/components/Text/TextStylePropTypes.js#L42-L57).
And when you want to use this, you cannot just say "ok that will pass" (and that's a good thing - thanks reason). So you have to redefine the Text (in my case for example I called it `Link`) with a copy of the original component and the new props you want to pass. In native world, this will just be ignored so that's not a problem. But during the copy/paste/adjust, you will miss part using private apis.
I think I found a workaround (by adding the private folder to bsconfig dir) but that's not ideal as you have to add basically all the folder and remove the `open BsReactNative` otherwise, private are still private... Something less "private" will be a good thing, even if not super accessible in the first place.

saschatimme:
Maybe we could have an `__Internal` module which also gets exposed for such things? This will probably also help if you want port a new component / API without the hassle of creating a fork first etc.

arnarthor:
Yeah, I actually think there's no harm in exposing the events in this case. But having everything that's private exposed via __Internal might be a bit overboard 😄

wokalski:
Exposing `__Internal` sounds fine to me as long as the answer for "can I use these to circumvent the types" is no.

gunnigylfa:
I actually don't have a problem with some apis being private or haven't ran into walls yet having it like that maybe. But the purpose of this PR was to make the Events api public, because Android and iOS both expose these events. Maybe we can create an issue for the `Private/__Internal` discussion.

alliannas:
I agree that internal/private discussion should we discussed on a separate issue.

But regarding making RNEvent public again.

It does not seem to make a whole lot of sense for bs-react-native to hide away these events since they are publicly available, both in react native and natively (ios/android).

Note that its also very common to access these things when writing native UIs.

If we could get this in the next release it would save us from maintaining our own fork.
cc @arnarthor @wokalski

wokalski:
Fuck, I forgot about js sources. gosh

wokalski:
we need proper danger/CI so that this doesn't happen.

PR #106, opened 2017-12-07
Fix invalid json in bsconfig.json

MoOx:
Current json is not valid.
I think it's better to keep it valid.

I could set up [prettier](https://github.com/prettier/prettier) if you are interested, to prevent this kind of "issues".

### DIFF
--- a/bsconfig.json
+++ b/bsconfig.json
@@ -10,13 +10,13 @@
       "dir": "src",
       "subdirs": [
         {
-          "dir": "components",
+          "dir": "components"
         },
         {
           "dir": "private",
-          "public": [],
-        },
-      ],
+          "public": []
+        }
+      ]
     }
   ]
 }


wokalski:
I don't understand the pros of this one? Do you do anything particular with bsconfigs?

MoOx:
I just opened it in atom and saw syntax errors (json lint). Valid json does not accept trailing comma. I don't do anything special with it, but it could be incorrectly parsed "someday" by a valid json parser.

arnarthor:
It's not invalid syntax, bsconfig uses JSON5 so this is fine. I actually like trailing commas everywhere makes it easier to add more stuff later on. That said, I don't have a strong opinion on this, but I'd prefer to keep them.

wokalski:
Alrighty, closing this one then.

MoOx:
Make sense. Will remove my json lint plugin then haha.

PR #107, opened 2017-12-07
Fix master JS

MoOx:
As discussed in #97, current master has not js source up to date.


wokalski:
I have a fix locally on the style branch. The commits will land very soon.

PR #110, opened 2017-12-10
Fix Pct ".%" issue

### LINKED ISSUE #109: Pct(100.) is transformed to "100.%", should be "100%"
The JS code generated by `Pct(100.)` is actually working in react-native projects, but not when used in a browser, via react-native-web (browsers don't like the "100.%").
We should adjust this behavior (as discussed on Discord with @wokalski)

MoOx:
Closes #109

Note: I am new to ocaml/reason and would not have expected the need for the surrounding `string()` but I got an error during `npm run build`:

```
  We've found a bug for you!
  /Users/MoOx/Sync/Development/bs-react-native/src/style.re 12:17-31
  
  10 ┆ switch value {
  11 ┆ | Pt(px) => Encode.float(px)
  12 ┆ | Pct(pct) => Encode.pct(pct)
  13 ┆ };
  14 ┆ 
  
  This has type:
    string
  But somewhere wanted:
    Js.Json.t (defined as Js.Json.t)
``` 

With string() around the Printf.sprintf, seems ok.


wokalski:
In order to pack `opacity(Float(1.))` and `height(Pct(2.))` into one array we have to somehow encode them into one type. there are many options but because they are later on used by raw javascript in the raw float/string form we send them over saying "these are all Json values". `string` means that even though it doesn't exist in the compiled output!

MoOx:
Thanks for the explanation!

PR #113, opened 2017-12-12
[WIP] Add a script for consistent code formatting

wokalski:
I'm creating this PR in order to address formatting inconsistencies among people editing with various widths (including myself). If we configure a CI it should look at **each and every** commit in a PR so that if any of them contains wrong formatting, the commit is blocked and formatting needs to be fixed on per commit basis.

### DIFF
--- a/format
+++ b/format
@@ -1,0 +1,2 @@
+refmt --in-place --print-width 80 RNTester/src/**/*.re
+refmt --in-place --print-width 80 src/**/*.re


format
```diff
@@ -0,0 +1,2 @@
+refmt --in-place --print-width 80 RNTester/src/**/*.re
+refmt --in-place --print-width 80 src/**/*.re
```
bohdan-shulha: Will this cover `.rei` as well? 

BTW, there's a sense to add contributing guidelines with `precommit` hook instructions.

wokalski:
Closing since I won't have time to properly implement it soon :(

PR #114, opened 2017-12-13
Add accessibility labels

gunnigylfa:
In an effort to make our reason react native apps accessible for as many as possible we added:
- Accessibility label prop support for `TouchableHighlight`, `TouchableOpacity`, `TouchableWithoutFeedback` and `Text`. 
- Changed the type of accessibilityLabel in `View` from  `ReasonReact.reactElement` to string.


src/components/text.rei
```diff
@@ -2,6 +2,7 @@ module type TextComponent = {
   let make:
     (
       ~accessible: bool=?,
+      ~accessibilityLabel: string=?,
```
kristinns: I would remove `accessibilityLabel` from the `Text` element, it didn't seem to work when I tested it. 
I think if you need it, you should wrap it with `View` and add `accessibilityLabel` on that instead.
Also, it's not mentioned in the [documentation](https://facebook.github.io/react-native/docs/accessibility.html)

gunnigylfa: Nice catch! removed it removed in: ef88999

lib/js/src/components/text.js
```diff
@@ -55,9 +56,10 @@ function CreateComponent(Impl) {
 
 var Impl = /* module */[/* Text */ReactNative.Text];
 
-function make(accessible, allowFontScaling, ellipsizeMode, numberOfLines, onLayout, onLongPress, onPress, pressRetentionOffset, selectable, style, testID, selectionColor, textBreakStrategy, adjustsFontSizeToFit, minimumFontScale, suppressHighlighting, value, children) {
+function make(accessible, accessibilityLabel, allowFontScaling, ellipsizeMode, numberOfLines, onLayout, onLongPress, onPress, pressRetentionOffset, selectable, style, testID, selectionColor, textBreakStrategy, adjustsFontSizeToFit, minimumFontScale, suppressHighlighting, value, children) {
```
wokalski: You forgot to build after removing `accessibilityLabel` from `Text`. The rest is awesome. Good job

gunnigylfa: Ah nice catch, spaced out on that one. Fixed in: d0ccfc1

wokalski:
@gunnigylfa good job!

wokalski:
@kristinns @MoOx thanks for reviewing

PR #115, opened 2017-12-13
update Reason React native namespace

nhducit:


MoOx:
@wokalski hope a "Squash and merge" is fine (I like this, especially for single commit to avoid 2 commits for a simple thing)

nhducit:
woa, so fasssst

wokalski:
Yeah - single commits are an exception, I'm totally cool with squashing them.

PR #119, opened 2017-12-15
Fixing value mapping for flexbox align properties

gunnigylfa:
The values the variants for Flexbox align methods map to are not recognized by the react-native style api. They should be camelCased rather than snake cased (like the web implementation of flexbox does).


gunnigylfa:
This is a misunderstanding on my behalf. Sorry, disregard this PR.

PR #121, opened 2017-12-19
Upgrade examples following to the SDK changes

bohdan-shulha:
There were a few outdated examples in README and they just didn't work.


wokalski:
Thank you very much!

PR #123, opened 2017-12-26
Support non-standard Platform.OS identifiers & iOS idioms

ryanartecona:
This modifies the `Platform` module to support more OS's than only "ios" and "android". Namely, "web" for react-native-web and "sketch" for react-sketchapp are added as explicit variant cases, and an `Unknown(string)` case is added for other/future use.

Support is also added for distinguishing iPhone/iPad/tvOS within the `IOS` case by giving that constructor an argument accordingly.

Other notes:
- The `version` implementation also now defaults to `0` for undefined versions, since react-native-web [doesn't define a version](https://github.com/necolas/react-native-web/blob/master/src/apis/Platform/index.js#L13-L16). I thought it would be better to let it remain an `int` rather than changing it to an `option(int)`.
- `select` doesn't yet support any of the above, since I was unsure how best to address it. Extending the current signature `(~ios: 'a, ~android: 'a) => 'a` to include required named arguments for the new OS cases seemed counter to its existence as a convenience function. Alternatively, we could make those named arguments all optional instead of all required, and either add a `~default` argument or make the result type `option('a)`.

    Personally, I'd rather drop it from the module entirely. Since reason already has nice pattern matching, this `select(~ios: 'a, ~android: 'a)` doesn't really add anything you don't already get for free with the `os` variant. Also, as a user, I'd rather define my own helper for this in my app so I can require cases for platforms I decide to target, and ignore platforms I don't, but that's not something we can really do generically in this library.
- I've tested this on `IOS(Phone)` and `Web` platforms, but none of the others.
- I don't know how seriously this project is taking backwards API compatibility yet. This is of course a breaking change.
- I don't have much attachment to the API decisions made here. This is just what I found myself wanting after configuring a react-native app to work with react-native-web.


src/platform.rei
```diff
@@ -1,11 +1,19 @@
...
+  | Phone
+  | Pad
+  | TV;
+
 type os =
-  | IOS
-  | Android;
+  | IOS(iosIdiom)
+  | Android
+  | Web
```
wokalski: I generally don't think it's a good idea to mix `react-native-web` into the mix here because a number of little differences might make a few apis not as nice to use as they possibly could be.

That said, I do think we should provide escape hatches. IMO the best way to proceed is to remove web and sketch variants but leave `Unknown` as is.

---

EDIT: Alternatively both `os` and `version` could be changed to functions and throw exceptions in case of something unknown. The problem is that pattern matches should be exhaustive and the vast majority of apps uses one codebase only for the mobile. Therefore additional pattern matches for unused platforms might be a bit tedious for those apps.

src/platform.re
```diff
@@ -1,27 +1,68 @@
...
+external _select :
+  {
+    .
+    "ios": 'a,
+    "android": 'a
+  } =>
+  'a =
   "select";
 
 let select = (~ios, ~android) => _select({"ios": ios, "android": android});
```
wokalski: I'm also in favour of removing select.

ryanartecona:
Thanks for the review.

Agreed about pattern matching usage. I've made the suggested changes to make `os` and `version` functions instead of values so they can throw exceptions for non-standard values. Pattern matches in user code need only handle `IOS(_)` and `Android` cases in the common usage case, and can choose to e.g. also match on `exception (UnkownPlatform("web"))` as an escape hatch.

I've also removed `select`.

PR #124, opened 2017-12-27
Add Animated styles example to README

Thomas0c:
An addition to the README to illustrate how the `Animated` type can be used in `styles`.

Added additional methods to present the `Animated` type within a relevant context instead of just presenting it on it's own.

### DIFF
--- a/README.md
+++ b/README.md
@@ -115,7 +115,38 @@ let styles =
 <View style=styles##wrapper />;
 ```
 
-## Toubleshooting
+For animations:
+```reason
+open BsReactNative;
+
+[...]
+type state = {animatedValue: Animated.Value.t};
+let component = ReasonReact.reducerComponent("Example");
+
+initialState: () => {animatedValue: Animated.Value.create((-100.))},
+
+/* Start animation in method */
+Animated.CompositeAnimation.start(
+  Animated.Timing.animate(
+    ~value=state.animatedValue,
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -115,7 +115,39 @@ let styles =
...
+### Animations
+
+```reason
+open BsReactNative;
+
+[...]
+type state = {animatedValue: Animated.Value.t};
+let component = ReasonReact.reducerComponent("Example");
+
+initialState: () => {animatedValue: Animated.Value.create((-100.))},
```
gunnigylfa: I'm just wondering performance wise if it is a good thing to instruct users to place the animation value in the state. If you are animating the value from -100 to zero with a linear function would it not re-render 100 times since the state changes 100 times? I might be wrong and these re-renders do not happen. 
I personally just have an instance variable inside the make function.

wokalski: No, it won't. the internal value of the value is animated, not the value itself.

Thomas0c: Worth mentioning that React Native docs use `state` to store animation value - see: https://facebook.github.io/react-native/docs/animations.html. Thereby not saying that alternatives should not be mentioned/highlighted.

wokalski: They should not IMO, we just want to highlight the API. I'm unable to merge this PR (because I contributed) but please someone accept and merge.

PR #127, opened 2018-01-13
bumping react

dysinger:
The latest create-react-native-app combined with bs-react-native instructions wouldn't work

I had to bump the dependencies in package.json for bs-react-native before it would work


wokalski:
I  think we should bump to 0.52.0 right away but go through release notes and make updates accordingly.

wokalski:
Closing since there's no response. Feel free to ping me if you want to finish this one!

PR #130, opened 2018-01-17
creating accessory type that includes section to match react-native s…

bjornj12:
When using SectionList.re I hit the problem of RenderAccessoryView always having section as undefined.

This is the definition in react-native

```javascript
 renderSectionHeader?: ?(info: {section: SectionT}) => ?React.Element<any>,
```
They however don't define any object that only includes section, so I assume they are simply referring to the renderBag object, but discarding all items but section. 

I created a matching type to catch the Js object and return it, as I feel it increases readability of how to use it.

example use:
```
let renderHeader = (info: SectionList.renderAccessory(State.FriendRow.t)) => {
        let section = info.section;
        let sectionKey = section.key;
        switch sectionKey {
        | Some(key) => <View> <Text value=key /> </View>
        | None => ReasonReact.nullElement
        }
      };
<SectionList
      ....
      renderSectionHeader=(SectionList.renderAccessoryView(renderHeader)) 
/>
```

### DIFF
--- a/src/components/sectionList.re
+++ b/src/components/sectionList.re
@@ -12,7 +12,8 @@ and jsRenderBag('item) = {
   "index": int,
   "section": jsSection('item),
   "separators": {. "highlight": unit => unit, "unhighlight": unit => unit}
-};
+}
+and jsRenderAccessory('item) = {. "section": jsSection('item)};
 
 type jsSeparatorProps('item) = {
   .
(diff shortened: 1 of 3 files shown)


src/components/sectionList.re
```diff
@@ -12,7 +12,8 @@ and jsRenderBag('item) = {
   "index": int,
   "section": jsSection('item),
   "separators": {. "highlight": unit => unit, "unhighlight": unit => unit}
-};
+}
+and jsRenderAccessory('item) = {. "section": jsSection('item)};
```
arnarthor: I'd prefer to have this the same as everything else so not using `and` but just defining it as another type.

src/components/sectionList.rei
```diff
@@ -8,6 +8,9 @@ and section('item) = {
   data: array('item),
   key: option(string),
   renderItem: option((renderBag('item) => ReasonReact.reactElement))
+}
+and renderAccessory('item) = {
```
arnarthor: Same here

arnarthor:
LGTM. Tried digging through the react-native codebase to find this, but the type is never defined. I don't really get what's going on with this. But tested this myself and LGTM

arnarthor:
No wait OK, now I get it. It's referencing this object here https://github.com/facebook/react-native/blob/6b95c4fb142a7015b2afca50cc19eec0b8913d8c/Libraries/Lists/VirtualizedSectionList.js#L27

This is the same object apart from the missing KeyExtractor field, but I'm fine with skipping that for now since it was never defined before anyways. 

🚀

wokalski:
maybe someone wants to rewrite lists in reason :).

PR #136, opened 2018-01-28
Fix callback on Animated Value 

sscaff1:
I was adding an listener to an animated value and notice that the wrong type was defined in Animate.Value. This fixes the problem. I had to change this in the Animated source code to make this work for my project. Let me know if something is missing. This is my first PR to the reason community. 😄


wokalski:
Good catch. Thank you!

PR #140, opened 2018-02-01
Clipboard — bindings

mobily:
This PR adds `Clipboard` bindings.

API:
https://facebook.github.io/react-native/docs/clipboard.html

Usage examples:
```
Clipboard.setString("ReasonML");
Clipboard.getString()
|> Js.Promise.then_(text => {
   Js.log(text);
   Js.Promise.resolve();
 });
```


src/clipboard.re
```diff
@@ -0,0 +1,5 @@
+[@bs.module "react-native"] [@bs.scope "Clipboard"]
+external getString : unit => Js.Promise.t(string) = "";
```
wokalski: Does it ever throw?

src/clipboard.rei
```diff
@@ -0,0 +1,5 @@
+[@bs.module "react-native"] [@bs.scope "Clipboard"]
```
wokalski: Our convention is to use lets in the `rei`s. `external` is just an implementation detail

src/clipboard.re
mobily: it doesn't, and it returns an empty string (in Promise) if there's nothing in the clipboard

mobily:
@wokalski ✔️ fixed!

mobily:
thanks!

PR #141, opened 2018-02-01
CameraRoll — bindings

mobily:
This PR adds `CameraRoll` bindings.

API:
https://facebook.github.io/react-native/docs/cameraroll.html

Usage examples:
```
/* saveToCameraRoll */

CameraRoll.saveToCameraRoll(~uri="file:///sdcard/img.png", ())
|> Js.Promise.then_(result => {
     switch result {
     | Js.Result.Ok(assets) => Js.log(assets)
     | Js.Result.Error(error) => Js.log(error)
     };
     Js.Promise.resolve();
   })
|> ignore;

/* getPhotos */

CameraRoll.getPhotos(~first=20, ())
|> Js.Promise.then_(result => {
     result##edges |> Array.iter(photo => Js.log(photo##node##_type));
     Js.Promise.resolve();
   })
|> ignore;
```


src/cameraRoll.rei
```diff
@@ -0,0 +1,67 @@
...
+  | PhotoStream
+  | SavedPhotos;
+
+type assetType =
+  | All
+  | Videos
+  | Photos;
+
+let saveToCameraRoll:
+  (~uri: string, ~type_: fileType=?, unit) => Js.Promise.t(string);
```
wokalski: I think it throws, so I suggest wrapping it in some result type so that the user is forced to handle it explicitly.

src/cameraRoll.rei
```diff
@@ -0,0 +1,67 @@
...
+
+let getPhotos:
+  (
+    ~first: int,
+    ~after: string=?,
```
wokalski: This should be of an abstract type `cursor` `pageInfo { end_cursor }` (according to the [docs](https://facebook.github.io/react-native/docs/cameraroll.html#getphotos)). Thanks to that we can be sure that cursors are passed properly.

src/cameraRoll.rei
```diff
@@ -0,0 +1,67 @@
...
+    ~groupTypes: groupType=?,
+    ~groupName: string=?,
+    ~assetType: assetType=?,
+    ~mimeTypes: string=?,
+    unit
+  ) =>
+  Js.Promise.t(result);
```
wokalski: Maybe result is not the most fortunate name since `result` type is standard in OCaml (although in buckle script it's under Js.Result.t)

wokalski: Also, I think it throws, so I suggest wrapping it in `Js.Result.t` type so that the user is forced to handle it explicitly.

src/cameraRoll.rei
mobily: done! :) there is a mistake in the docs, it says `end/start_cursor` has `boolean` type

src/cameraRoll.rei
mobily: does `assets` sound better? 😉

src/cameraRoll.rei
mobily: hm, are you sure this is the correct way to handle Promise's errors?
I think the following code is correct (using `catch` in the Promise's chain):

```
CameraRoll.saveToCameraRoll(~uri="file:///sdcard/img.png", ())
|> Js.Promise.then_(uri => {
     Js.log(uri);
     Js.Promise.resolve();
   })
|> Js.Promise.catch(error => {
    /* do something with error */
    Js.Promise.resolve();
});
```

mobily:
@wokalski there is still one thing to discuss: https://github.com/reasonml-community/bs-react-native/pull/141#discussion_r165817497

src/cameraRoll.rei
wokalski: Yes, but I think this is something we need to do inside the library, not exposing it to the user. So that you `catch` it and then resolve to `Js.Result.Error SomeErrorOrPossiblyUnit`. otherwise you resolve to `Js.Result.Ok something`. Thanks to that our promise is safe and users won't suffer from unhandled exceptions.

mobily: @wokalski ok, I think I got it! please check this commit: https://github.com/reasonml-community/bs-react-native/pull/141/commits/e3b6825208ae83bb574cf926709e2bd8622542f2

mobily:
@wokalski it's ready for another review 😉

mobily:
thanks! 🎉

PR #142, opened 2018-02-02
Run RNTester 🎉

mobily:
After some changes, now I'm able to run the `RNTester` app on Android/iOS emulator. 

In short, this pull request removes legacy code (issue #96) and changes the configuration (the essential files is `rn-cli.config.js`). Right now we don't need the copy script, and also for `RNTester`  I use `reason-react` and `bs-platform` packages installed in the root node_modules (for consistency). Also, this PR updates RN (RNTester app) to the version `0.52.2`.

Edit: forgot to mention, please run first in the root directory either: `yarn`, `npm install`  then do the same thing in the `RNTester` directory.

Screenshots:

android emulator - nexus_5x_api_25 5554 2018-02-02 14-52-32
iphone 8 plus - 11 2 2018-02-02 14-54-33


RNTester/bsconfig.json
```diff
@@ -1,14 +1,17 @@
...
-            "dir": "examples"
-        }
-    }]
+  "name": "react-native-tester",
+  "reason": {
+    "react-jsx": 2
+  },
+  "refmt": 3,
+  "bs-dependencies": [
+    "../../node_modules/reason-react",
```
wokalski: Shouldn't it stay `reason-react`?

RNTester/rn-cli.config.js
```diff
@@ -0,0 +1,15 @@
+const path = require('path')
+
+const config = {
```
wokalski: what is this?

RNTester/src/RNTesterApp.re
```diff
@@ -31,41 +38,52 @@ let header = (~onBack=?, ~title, ()) =>
...
+      | ChangeCurrentExample(example) => Update({currentExample: example})
+      },
+    render: ({state, handle}) => {
       let components = ExampleList.components;
       switch state.currentExample {
       | None =>
         <View style=styles##exampleContainer>
           (header(~title="ReasonRNTester", ()))
-          <RNTesterExampleList components onPress=(update(onPress)) />
+          <RNTesterExampleList components onPress=(handle(onPress)) />
```
wokalski: ~~We should use `reduce` here.~~ I forgot it's deprecated. How about an inline function calling send?

RNTester/src/examples/ViewExample.re
```diff
@@ -15,33 +15,39 @@ let styles =
...
     initialState: () => {showBorder: false},
-    render: ({state, update}) =>
+    reducer: (action, _state) =>
+      switch action {
+      | UpdateBorderState(showBorder) => Update({showBorder: showBorder})
+      },
+    render: self =>
       Style.(
-        <TouchableWithoutFeedback onPress=(update(handlePress))>
+        <TouchableWithoutFeedback onPress=(self.handle(handlePress))>
```
wokalski: How about an inline function calling send?

RNTester/src/examples/ViewExample.re
```diff
@@ -78,24 +84,30 @@ module ViewBorderStyleExample = {
...
+          [|(-1), 0, 1, 2|];
         } else {
-          [|2, 1, 0, (-1)|]
+          [|2, 1, 0, (-1)|];
         };
-      let zIndexStr = (i) =>
+      let zIndexStr = i =>
         "ZIndex " ++ string_of_int(Array.unsafe_get(indices, i));
-      <TouchableWithoutFeedback onPress=(update(handlePress))>
+      <TouchableWithoutFeedback onPress=(handle(handlePress))>
```
wokalski: same as above

RNTester/transformer.js
```diff
@@ -0,0 +1,35 @@
+const path = require('path')
+const babel = require('babel-core')
+
+const moduleResolverPlugin = require('babel-plugin-module-resolver').default
+const reactNativePreset = require('babel-preset-react-native')
+
+const transform = ({ src, filename }) => {
```
wokalski: :O what is this? why do we need it?

RNTester/bsconfig.json
mobily: yep, you're right

RNTester/transformer.js
mobily: it turns out it's useless

RNTester/rn-cli.config.js
mobily: this file allows to override default configuration of the Metro packager

as you see, I defined two paths (roots), so we don't have to copy files (using `scripts/copy.js`) from the root directory each time we run the packager

RNTester/src/examples/ViewExample.re
mobily: ✔️ done

RNTester/src/examples/ViewExample.re
mobily: ✔️ done

RNTester/src/RNTesterApp.re
mobily: ✔️ done

however, is there any reason why we should use inline functions?

mobily:
@wokalski it's ready for the review 😉

RNTester/src/RNTesterApp.re
wokalski: The result JS is better because there's no currying. Also, I find the indirection with handle a little bit weird. For me handle says, this is a function which performs side effects. Although with `reduce` changed to `send` the difference is a bit blurry.

RNTester/package.json
```diff
@@ -3,21 +3,19 @@
...
-    "bs-platform": "^2.0.0",
-    "react": "^16.0.0-alpha.12",
-    "react-native": "^0.46.0",
-    "reason-react": "^0.3.0"
+    "react": "^16.2.0",
+    "react-native": "^0.52.2"
   },
   "devDependencies": {
     "babel-jest": "20.0.0",
+    "babel-plugin-module-resolver": "^3.0.0",
```
wokalski: one more thing, do we need babel in this case?

mobily: we don't, good catch, thanks! 👍

mobily:
I guess it's ready to merge ;)

wokalski:
Thank you for the huge amount of work you poured into this. It's been a major undertaking and I'm super happy that you stepped up to do this.

Issue #54, opened 2017-08-03
Add CI for RNTester


saschatimme:
Currently the RNTester app is not tested, since I'm not sure how to setup everything.

Would be appreciated if someone with a little bit travis-ci foo could help out :)

arnarthor:
We're using Lerna internally at my company with a React Native app and then a Reason web and it works fine in Travis. I could take a look at this sometime soon if no one else grabs it before me.

arnarthor:
Or we could check this out for monorepo structure https://yarnpkg.com/blog/2017/08/02/introducing-workspaces/

saschatimme:
One issue is probably that the RN package [doesn't support symlinks](https://github.com/facebook/metro-bundler/issues/1) 😞

wokalski:
Fixed by #143

PR #143, opened 2018-02-04
Add CI for RNTester

mobily:
This PR adds CI for `RNTester` (issue: #54).

Log:
https://travis-ci.org/reasonml-community/bs-react-native/builds/337167048?utm_source=github_status&utm_medium=notification

### DIFF
--- a/RNTester/package.json
+++ b/RNTester/package.json
@@ -3,10 +3,10 @@
   "version": "0.0.1",
   "private": true,
   "scripts": {
-    "build": "bsb -make-world",
-    "clean-build": "bsb -clean-world -make-world",
+    "build": "node ../node_modules/.bin/bsb -make-world",
+    "clean-build": "node ../node_modules/.bin/bsb -clean-world -make-world",
     "packager:start": "node node_modules/react-native/local-cli/cli.js start",
-    "watch": "bsb -make-world -w",
+    "watch": "node ../node_modules/.bin/bsb -make-world -w",
     "start": "npm run build && npm run packager:start",
     "test": "jest"
   },
--- a/RNTester/package.json
+++ b/RNTester/package.json
@@ -3,7 +3,8 @@
   "version": "0.0.1",
   "private": true,
   "scripts": {
-    "build": "bsb -clean-world -make-world",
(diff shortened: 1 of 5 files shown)


mobily:
@wokalski ready for review :)

RNTester/package.json
```diff
@@ -3,9 +3,10 @@
   "version": "0.0.1",
   "private": true,
   "scripts": {
-    "build": "bsb -clean-world -make-world",
+    "build": "node ../node_modules/.bin/bsb -make-world",
+    "clean-build": "node ../node_modules/.bin/bsb -clean-world -make-world",
     "packager:start": "node node_modules/react-native/local-cli/cli.js start",
-    "watch": "bsb -make-world -w",
+    "watch": "node ../node_modules/.bin/bsb -make-world -w",
```
wokalski: what is this?

RNTester/package.json
```diff
@@ -15,8 +16,7 @@
   },
   "devDependencies": {
     "babel-jest": "20.0.0",
-    "babel-preset-react-native": "1.9.1",
-    "fs-extra": "^3.0.1"
+    "babel-preset-react-native": "1.9.1"
```
wokalski: Why do we need babel transforms at all?

RNTester/package.json
mobily: ok, in the local environment `bsb` is usually is installed globally (according to this: https://reasonml.github.io/docs/en/quickstart-javascript.html), and here we must use `bsb` installed in the root `node_modules` (`bsb` belongs to the `bs-platform` package)

RNTester/package.json
mobily: we need this to build the iOS/Android app correctly :)

RNTester/package.json
wokalski: I get that but do we need to invoke node instead of just the path itself?

mobily: 👍 fixed!

wokalski:
Thank you for that. It's been broken for some time now.

mobily:
thanks! 🎉

PR #144, opened 2018-02-04
Add `NetInfo` bindings

mobily:
This PR adds `NetInfo` bindings (with examples).

I had some troubles with these bindings, and I'm looking forward to your feedback 😉 

Usage examples:

```
let handleConnectivityChange = connectionInfo => Js.log(connectionInfo);

NetInfo.addEventListener(handleConnectivityChange);

NetInfo.removeEventListener(handleConnectivityChange);

NetInfo.getConnectionInfo() 
|> Js.Promise.then_(
     (connectionInfo) => {
       Js.log(connectionInfo);
       Js.Promise.resolve()
     }
   );

NetInfo.IsConnected.fetch()
|> Js.Promise.then_(
     (isConnected) => {
       Js.log(isConnected);
       Js.Promise.resolve()
     }
   );
```


src/netInfo.re
```diff
@@ -0,0 +1,83 @@
+type infoType = string;
+
+type infoEffectiveType = string;
+
+type connectionType =
```
arnarthor: It might be against the convention in this project, but I think we should use the polymorphic generators here. This projects pre-dates the generators and I think it would make sense to use this for new features

https://bucklescript.github.io/docs/en/generate-converters-accessors.html#convert-between-js-string-enum-and-bs-polymorphic-variant

What do you think @wokalski ?

mobily: @arnarthor I was thinking about that too, and I believe it could simplify the bindings in some way

arnarthor: Definitely, we have a lot of these helper functions that only convert between JS and OCaml types.

wokalski: To be honest I don't think that the maintenance difficulty differs. It's slightly more convenient but bindings libraries like this are fairly static (in opposite to ever changing real world applications).

To give you a real argument compare [v0.34 docs](https://facebook.github.io/react-native/docs/0.34/netinfo.html) with [v0.53-RC docs](https://facebook.github.io/react-native/docs/netinfo.html). There's hardly any difference.

src/netInfo.re
```diff
@@ -0,0 +1,83 @@
...
+  external isConnected : t = "";
+  [@bs.send.pipe : t]
+  external _addEventListener :
+    ([@bs.as "connectionChange"] _, bool => unit) => unit =
+    "addEventListener";
+  [@bs.send.pipe : t]
+  external _removeEventListener :
+    ([@bs.as "connectionChange"] _, bool => unit) => unit =
+    "removeEventListener";
+  [@bs.send.pipe : t] external _fetch : unit => Js.Promise.t(bool) = "fetch";
```
wokalski: `fetch` [is deprecated](https://facebook.github.io/react-native/docs/netinfo.html#fetch) or `getConnectionInfo` doesn't exist depending which version you target. Managing that is major PITA with building bindings.

mobily: hm, according to the source `NetInfo.isConnected.fetch` is not deprecated, take a look here:
https://github.com/facebook/react-native/blob/master/Libraries/Network/NetInfo.js#L242

and `NetInfo.fetch` is deprecated 😉 
https://github.com/facebook/react-native/blob/master/Libraries/Network/NetInfo.js#L182

> Managing that is major PITA with building bindings.

is there any solution to handle this?

src/netInfo.re
arnarthor: No I agree, I was more so thinking about this in a way that it's less noise in the repo with these manual conversion functions. But either way is fine by me. Consistency might be better here since everything else is this way.

src/netInfo.re
wokalski: @mobily this is a good question :). I don't have a good answer for that. Feel free to share ideas.

wokalski:
Thank you @mobily. Stellar job 🙌

mobily:
thanks! 🎉

PR #146, opened 2018-02-04
Add Animated.Easing bindings

gilbert:


arnarthor:
Oh wait no, @gilbert could you add the compiled JS files for this change.

src/animatedRe.re
```diff
@@ -14,6 +14,32 @@ module CompositeAnimation = {
...
+
+  [@bs.module "react-native"] [@bs.scope "Easing"] external back : float => t = "";
+  [@bs.module "react-native"] [@bs.scope "Easing"] external elastic : float => t = "";
+
+  [@bs.module "react-native"] [@bs.scope "Easing"] external in_ : t => t = "";
+  [@bs.module "react-native"] [@bs.scope "Easing"] external inOut : t => t = "";
+  [@bs.module "react-native"] [@bs.scope "Easing"] external out : t => t = "";
+
+  [@bs.module "react-native"] [@bs.scope "Easing"]
+  external bezier: float => float => float => float => t = "";
```
mobily:
```
external bezier: (float, float, float, float) => t = "";
```

why not like this? 😉

arnarthor: Hmm refmt should take care of that. Maybe run refmt on the files as well

gilbert: I can run refmt, but the entire file changes. Is that ok?

gilbert:
Added compiled JS and updated bezier sig.

src/animatedRe.re
mobily: looks good! thanks :)

mobily:
@gilbert one more thing, could you please update the `STATUS.md` file?

gilbert:
Done

PR #147, opened 2018-02-13
making Colors be either string or interpolated

bjornj12:
### Color Interpolation
Introducing Interpolation to backgroundColor.

according to https://facebook.github.io/react-native/docs/animations.html

` interpolate() also supports mapping to strings, allowing you to animate colors as well as values with units.`
Colors should be able to be interpolated.

This PR is my proposal on how we add interpolation to colors, and have set it for backgroundColor.

I'm not entirely sure of all the colours this should apply to (shadow, text, ...),
so i wanted to start a discussion here of how we can move forward with this.

### Updating lib/js
I noticed when building my PR, that some components under lib had not been updated 
from ```reasonReact``` to ```ReasonReact```

I decided to slide it in, otherwise this will be modified by bsb for all future PR's untill someone checks it in.


bjornj12:
@wokalski can i get feedback from you on this? 😄

src/style.re
```diff
@@ -50,6 +50,16 @@ let encode_float_interpolated_animated = (value) =>
...
 
+type string_interpolated =
+  | String(string)
+  | Interpolated(AnimatedRe.Interpolation.t);
+
+let encode_string_interpolated = (value) =>
+  switch value {
+  | String(value) => Encode.string(value)
+  | Interpolated(value) => Encode.interpolatedValue(value)
+  };
```
mobily: you can use the `fun` operator (it's shorter and cleaner)

```
let encode_string_interpolated =
  fun
  | String(value) => Encode.string(value)
  | Interpolated(value) => Encode.interpolatedValue(value);
```

arnarthor:
The diff for this PR is unecessary big. Please update to the latest reason-cli so that refmt doesn't add all these unecessary parens changing formatting all over. When you've done that just update this commit instead of adding a new one.

arnarthor:
Apart from that comment, LGTM

wokalski:
@bjornj12 hey, thank you for your work. Two questions. 

1. could you separate out the refmt changes? i.e. please upgrade if you have an old version (or formatting at a non standard width (just realised we don't have anything about the width in the readme cc @arnarthor ). And then if there are any changes left, please commit them separately. 

2. Could you rebase your changes? You can simply start rebase and instead of resolving manually, try compiling the project because the diffs are in `lib`.

arnarthor:
Yeah, we should add some refmt info to the `README.md`. I'm using defaults myself, are you using anythings outside the defaults @wokalski?

wokalski:
@arnarthor I used to use defaults but the problem is that in the editor refmt looks at the window width. So I have it hard coded at 80. Also the problem is that defaults change :D

arnarthor:
lol good point! I'll dig up the current defaults and put them into the README 😄

bjornj12:
hey @wokalski, sorry that i haven't responed earlier, have been quite busy.
I have recommitted the changes, and split them up as asked, they are in 4 parts:
1. build with bs-platform 2.2,
2. refmt style.re/i and RNTester
3. update style.re/i with my changes
4. update RNTester to fullfill the new style color types

wokalski:
@bjornj12 thank you. #160 is merged. It should be easy to cherry pick 3 and 4 now.

bjornj12:
@wokalski  Awesome, i have rebased with master, seems like refmt still wanted to add a bit of trailing commas etc, its seperated in the first commit

mobily:
> seems like refmt still wanted to add a bit of trailing commas etc, its seperated in the first commit

great! I forgot to use `refmt` in the `RNTester` directory

wokalski:
Awesome. Thanks @bjornj12 !

PR #148, opened 2018-02-20
Update bs-platform to 2.2.0

flash-gordon:
This PR bumps version of bs-platform and replaces `Js.Undefined.from_opt/to_opt` with `Js.Undefined.fromOption/toOption`. All "additions" are a result of `refmt` solely.

I didn't change these two lines
https://github.com/reasonml-community/bs-react-native/blob/c466850346715c212bd31734d63d827d0a3a83b5/src/datePickerAndroid.re#L50-L51
because there are no `Js.Nullable.fromOption` in the newer version of BS (see https://github.com/BuckleScript/bucklescript/issues/2531).


gunnigylfa:
Nice! I Tested your branch against our project which is a fairly large one ( ~ 26k loc) And it appears to be working. So LGTM 👍

arnarthor:
Thank you so much for this @flash-gordon 👍 I'll try to get a new version out during the weekend. I want to see if we can get in #147 as well before bumping versions since these two will both be breaking changes.

flash-gordon:
Sure, thanks folks

PR #152, opened 2018-02-26
WebView Js.nullable instead of option

### LINKED ISSUE #151: WebView Js.nullable instead of option
It seems to me that in WebView.rei, we should change `option(t)` to `Js.nullable(t)`
```reason
~onNavigationStateChange: {
  .
  "url": option(string),
  "title": option(string),
  "loading": option(bool),
  "canGoBack": option(bool),
  "canGoForward": option(bool)
} => unit
```
to
```reason
~onNavigationStateChange: {
  .
  "url": Js.nullable(string),
  "title": Js.nullable(string),
  "loading": Js.nullable(bool),
  "canGoBack": Js.nullable(bool),
  "canGoForward": Js.nullable(bool)
} => unit
```

Indeed no conversion is done in `WebView.re`

maarekj:
Fix #151

### DIFF
--- a/src/components/webView.re
+++ b/src/components/webView.re
@@ -14,6 +14,30 @@ type iOSLoadRequestEvent = {
 
 type source;
 
+type callbackParams = Js.t({
+  .
+  url: option(string),
+  title: option(string),
+  loading: option(bool),
+  canGoBack: option(bool),
+  canGoForward: option(bool),
+});
+
+[@bs.obj]
+external makeCallbackParams: (~url:option(string)) => (~title:option(string)) => (~loading:option(bool)) => (~canGoBack:option(bool)) => (~canGoForward:option(bool)) => callbackParams = "";
+
+let optionalCallbackToJs = (callbackFunction) => Js.Undefined.fromOption(UtilsRN.option_map((callback) => {
+  (params) => {
+    callback(makeCallbackParams(
+      ~url=params##url |> Js.toOption,
+      ~title=params##title |> Js.toOption,
+      ~loading=params##loading |> Js.toOption |> UtilsRN.option_map(Js.to_bool),
+      ~canGoBack=params##canGoBack |> Js.toOption |> UtilsRN.option_map(Js.to_bool),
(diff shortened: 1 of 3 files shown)


maarekj:
Today, with option, it break.
Then changing to `Js.nullable` will not break compatibility because it never worked.

wokalski:
@maarekj You're right. Nevertheless we (quite) consistently expose `option` type and use `Nullalble` behind the scenes.

maarekj:
Obviously, it's best with the option type.
More difficult to implement, but better.

wokalski:
@maarekj let me know if you're planning to implement the suggestion.

maarekj:
@wokalski Done

src/components/webView.re
```diff
@@ -119,12 +119,62 @@ let make =
             "injectedJavaScript": fromOption(injectedJavaScript),
             "mediaPlaybackRequiresUserAction":
               fromOption(mediaPlaybackRequiresUserAction),
-            "onError": fromOption(onError),
-            "onLoad": fromOption(onLoad),
-            "onLoadEnd": fromOption(onLoadEnd),
-            "onLoadStart": fromOption(onLoadStart),
+            "onError": fromOption(UtilsRN.option_map((callback) => {
```
wokalski: This is repeated multiple times. Could you factor it out into a utility function?

maarekj:
It seems that the `onNavigationStateChange` callback takes an additional parameter compared to the others.
It is `jsEvaluationValue`(as you can see on [this link](https://medium.com/capriza-engineering/communicating-between-react-native-and-the-webview-ac14b8b8b91a))

wokalski:
@maarekj it makes little difference. Feel free to leave that one as is then.

maarekj:
ping @wokalski

wokalski:
@maarekj I'm slightly confused. Does `onNavigationStateChange` take the other argument?

maarekj:
No, I understood that it should not be added.
If we add it, we can no longer use the newly created function `optionalCallbackToJs` and the new type `callbackParams`

wokalski:
Hey, I'm closing this until there's further progress. The needed change is very small.

PR #154, opened 2018-02-27
Add optional style prop to README

peterpme:
:wave: 

This is something I struggled figuring out for awhile so I'd like to add it to the README.md

Thanks!

### DIFF
--- a/README.md
+++ b/README.md
@@ -114,6 +114,25 @@ let styles =
 
 <View style=styles##wrapper />;
 ```
+### Optional `style` prop:
+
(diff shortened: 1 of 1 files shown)


wokalski:
I think in this example, we need the trailing children argument and style to be optional. i.e. `~style=?`.

wokalski:
Closing this until there's any resolution/activity. To be honest I'm not a fan of putting various "quick tips" in the readme but if it's a common concern I'm happy to merge.

Issue #153, opened 2018-02-27
Make private types under TypesRN and Props public


jamieparkinson:
Apologies for opening an issue for a question, but either I've misunderstood how to do things (likely) or this can be resolved with a PR 🙃 

How come there's no public interface for `Props`? Specifically: I have a function that uses `PanResponder.panHandlers` and thus returns the `Props.touchResponderHandlers` type. It works fine, but it seems like I can't define a signature for the module in which said function lives without access to this type: is there a reason why it's not exposed and/or a way in which I can define the signature without having direct access to the type?

wokalski:
@jamieparkinson Could you show some snippets? It's hard to wrap your head around things given only a description.

Generally speaking, we leave types abstract to make the API as simple as possible.

jamieparkinson:
Sure! A minimal example would be that I'd like to write something like:

```
module type ExampleType = {
    let getHandler: unit => BsReactNative.Props.touchResponderHandlers; 
};

module Example = {
    let responder = PanResponder.create(...);
    let getHandler = () => PanResponder.panHandlers(responder);
};
```

But afaik I can't use `BsReactNative.Props.touchResponderHandlers` due to the `Prop` module's privacy. Is there a way round this (or just something I've missed)? I thought I might be able to use an abstract type in the interface but this seems to just move the issue to the module itself...

wokalski:
`touchResponderHandlers` should be public. But I think that `props` module should remain private. It's an implementation detail. Maybe we can move `touchResponderHandlers` somewhere?  Same applies to `TypesRN` module. I think all common types should live in one module. Either called `Common` or (IMO slightly better) in modules like `Insets` or `TouchResponder`. Then they can be properly grouped. It's a mess in that area, geez.

jamieparkinson:
I think I agree on the common types approach, but what do you think about (at least for the short term) putting `touchResponderHandlers` in the `PanResponder` module? The main issue I can see with this is that then `View` depends on `PanResponder`...

wokalski:
I think temporarily TypesRN should be renamed to Types/Common and `touchResponderHandlers` ought to be put there.

jamieparkinson:
Cool, that sounds good to me! :) Thanks for the help. I'm a bit busy atm but will try to PR in the next couple of days

wokalski:
@jamieparkinson fantastic, looking forward to it.

PR #155, opened 2018-02-28
Make private types under TypesRN and Props public

jamieparkinson:
Resolves https://github.com/reasonml-community/bs-react-native/issues/153

Let me know if you think an interface file is necessary too 🙂


wokalski:
@jamieparkinson awesome. One question, can we change `TypesRN` to `Types`?

jamieparkinson:
@wokalski hmmm, do you think doing that might cause some confusion when (as is likely/recommended) people are using `open BsReactNative`? `Types` is a generic enough name that I think there's an argument for including `RN`, perhaps as a prefix rather than a suffix though. Up to you!

wokalski:
I think that we have namespacing precisely in order to have precise names without suffixes and other things. I'm rooting for `Types`.

wokalski:
Also, since these are just types, only `rei` is needed IIRC. We don't need `.re`.

jamieparkinson:
Yeah, point taken re namespacing and good tip re `.rei` - have updated. (apologies for weird authorship on last commit, am using my old laptop with outdated git/ssh config)

PR #160, opened 2018-03-07
Update dependencies

mobily:
Summary:
- [x] update dependencies  to the latest versions
- [x] compile `javascript` files using the latest version of `bucklescript` (actually the PR changes only the auto-generated comment in each file)
- [x] format all `reason` files using the latest version of `refmt`

In short, I think this PR must be merged before new ones will be created. Let me know what's your opinion about that 😉


lib/js/src/components/image.js
```diff
@@ -96,7 +96,19 @@ function CreateComponent(Impl) {
                 tmp.body = prim$4[0];
               }
               if (prim$5) {
-                tmp.cache = prim$5[0];
+                tmp.cache = (function () {
```
ulrikstrid: Changed more than just the comment, is this intended?

ulrikstrid:
I went through the files and added a comment to the only change in .js files that didn't just have comment changes.
There is some lines where I think refmt made it go too narrow but if it's a default I don't really care.

wokalski:
while we're on it I'm advocating for updating `react-native`, too. I used to be naive in a sense that we can do it one day in such a way that we won't break anything but we simply don't have neither manpower nor infra to do that.

lib/js/src/components/image.js
ulrikstrid: From Bucklescript repo:

> @ulrikstrid it was a bug fix, maybe you did not hit it previously

mobily:
✔️ @wokalski done!

The `0.54.0` version of RN is available, but it uses `react@16.3.0-alpha` which is incompatible with the current `ReasonReact` version due to the latest API changes in React. That's why I decided to update RN to the `0.53.3` version.

ulrikstrid:
lib/js/src/components/image.js has a change other than comment

wokalski:
great!

arnarthor:
This LGTM. Should we merge this @wokalski?

mobily:
thanks 🎉

wokalski:
thank you @mobily 👍

PR #162, opened 2018-03-17
Add `Geolocation` bindings

mobily:
This PR adds `Geolocation ` bindings.

API:
https://facebook.github.io/react-native/docs/geolocation.html

Usage example (check: `RNTester/src/examples/GeolocationExample.re`):
```
/* Geolocation.getCurrentPosition */
type coords = {
  latitude: float,
  longitude: float,
};
type action =
  | UpdateCoords(coords);
type state = {coords};

Geolocation.getCurrentPosition(
  position =>
    ReasonReact.send(
      UpdateCoords({
        latitude: position##coords##latitude,
        longitude: position##coords##longitude,
      }),
    ),
  error => Js.log(error),
);
```


RNTester/src/examples/GeolocationExample.re
```diff
@@ -0,0 +1,144 @@
...
+        latitude: 0.0,
+        longitude: 0.0,
+      },
+    },
+    reducer: (action, _state) =>
+      switch (action) {
+      | UpdateCoords(coords) => Update({coords: coords})
+      },
+    render: ({state, handle}) =>
+      Style.(
```
peterpme: is it common practice to "open a local" in the middle of render instead of just doing Style. inline?

src/geolocation.re
```diff
@@ -0,0 +1,104 @@
+type rnConfig;
```
peterpme: do these last two files (re, rei) need to be capitalized? Geolocation.re?

arnarthor: They don't need to, and probably should stay lower cased in convention with everything else in the src folder.

RNTester/src/examples/GeolocationExample.re
arnarthor: It's just a preference, I use it a lot if I have more than 2 values from the same module and don't want to open it globally in the file. For smaller snippets I do it inline

src/geolocation.re
arnarthor: What is rnConfig?

src/geolocation.re
```diff
@@ -0,0 +1,104 @@
...
+    ~timeout: int=?,
+    ~maximumAge: int=?,
+    ~enableHighAccuracy: Js.boolean=?,
+    ~distanceFilter: int=?,
+    ~useSignificantChanges: Js.boolean=?
+  ) =>
+  watchPositionConfig =
+  "";
+
+[@bs.val "navigator.geolocation.setRNConfiguration"]
```
arnarthor: I'd change this to 
```reason
[@bs.val] [@bs.scope ("navigator", "geolocation")] external setRNConfiguration : rnConfig => unit = "";
let setRNConfiguration(makeRnConfig(~skipPermissionRequests?));
```

arnarthor: Same goes for all the nested bs.val bindings below

src/geolocation.re
```diff
@@ -0,0 +1,104 @@
...
+  _setRNConfiguration(makeRnConfig(~skipPermissionRequests?));
+
+[@bs.val "navigator.geolocation.requestAuthorization"]
+external requestAuthorization : unit => unit = "";
+
+[@bs.val "navigator.geolocation.stopObserving"]
+external stopObserving : unit => unit = "";
+
+[@bs.val "navigator.geolocation.getCurrentPosition"]
+external _getCurrentPosition :
```
arnarthor: These externals don't need to have the _ prefix. If you create a let binding directly after, this won't be accessible only the let binding.

src/geolocation.re
arnarthor: I also feel like this shouldn't be a part of the GeoLocation module. Maybe it should be in some sort of root module or even just Configuration module or something like that.

src/geolocation.re
mobily: it's an opaque type, look here: https://github.com/reasonml-community/bs-react-native/pull/162/files#diff-54cf7aeba2428bd523fa4fd98951effeR30

I guess the name is a bit confusing, do you have any suggestion?

src/geolocation.re
mobily: this is how the navigator global looks like:

react native debugger 2018-03-23 19-37-51

let's create `navigator.re` module, and add the `Geolocation` module there
we could add other bindings from the `navigator` global in next pull requests, what do you think about that? 😉

src/geolocation.re
mobily: this is how the navigator global looks like (while using the simulator):

react native debugger 2018-03-23 19-37-51

let's create `navigator.re` module, and add the `Geolocation` module there
we could add other bindings from the `navigator` global in next pull requests, what do you think about that? 😉

arnarthor: Oh wait, I misread, didn't notice that setRNConfiguration was under the geolocation scope.

Sorry about that

But I still want to change it to be `[@bs.val] [@bs.scope ("navigator", "geolocation") ...` instead of having the dots there 😄

src/geolocation.re
arnarthor: Yeah I get that it's an opaque type, just wondering what it's used for. It's just the name that's confusing me. Could you link me to the react native docs for this so I can see what it's intended for 😄

mobily:
@arnarthor ✔️ all done! thanks for the valuable feedback

mobily:
thank you! 🎉

PR #164, opened 2018-03-18
Add `KeyboardAvoidingView` bindings (component)

mobily:
This PR adds `KeyboardAvoidingView ` bindings.

Useful links:
https://facebook.github.io/react-native/docs/keyboardavoidingview.html
https://github.com/facebook/react-native/blob/master/Libraries/Components/Keyboard/KeyboardAvoidingView.js


peterpme:
:wave: anything I can do to help move this along?

grsabreu:
nice

mobily:
> One comment regarding the type of behavior

@arnarthor I don't see any comment 🤔

arnarthor:
How did github mess up my comment 🤔

src/components/keyboardAvoidingView.rei
```diff
@@ -0,0 +1,59 @@
...
+                              | `startsMedia
+                              | `summary
+                              | `text
+                            ],
+                          )
+                            =?,
+    ~accessibilityViewIsModal: bool=?,
+    ~shouldRasterizeIOS: bool=?,
+    ~keyboardVerticalOffset: float=?,
+    ~behavior: [ | `height | `position | `padding]=?,
```
arnarthor: Could we use the actual type defined in the module here instead of having to keep the two in sync.

mobily: I think I need an explanation because it's the interface file, and the `behavior` type is not defined here 🤔

arnarthor: Everything created in the re file should be available in the rei file as well. But I don't think it's a huge issue. This can always be generated so it will probably never be manually kept in sync anyways.

arnarthor:
LGTM

mobily:
thanks! 🎉

Issue #165, opened 2018-03-20
Style height(Js.Json.null) or width(Js.Json.null)


GTDev87:
`height` and `width` want to take a type `BsReactNative.Style.pt_pct_animated_interpolated`. Because of legacy react native behavior, `null` and `undefined` are useful options.  https://medium.com/the-react-native-log/tips-for-react-native-images-or-saying-goodbye-to-trial-and-error-b2baaf0a1a4d

If not an extension of properties for height and width, an unsafe style prop would be useful.

GTDev87:
https://github.com/reasonml-community/bs-react-native/pull/166

benadamstyles:
Was any solution found for this? Your PR was closed but it doesn't seem like anyone offered any alternative, or have I missed something? We still need to be able to pass `null`, currently I'm doing `[%raw "[\"width\", null]"]` but it would be great to have a typesafe solution!

MoOx:
See 
https://github.com/reasonml-community/bs-react-native/pull/166#issuecomment-377185382

benadamstyles:
@MoOx that PR was closed. Does that mean this is a won’t-fix?

MoOx:
This PR has been closed because we don't think the implementation offered by it wasn't a good fit (it's opening door to unsafe usages that you probably don't want - and if you really do weird things, just is js directly).

The comment I linked mentioned an alternative solution we would be open to that may resolve the initial issue.

benadamstyles:
Ah ok, thanks. I will PR!

MoOx:
This issue needs to be revisited with new bindings. Feel free to report with more details accurate to new bindings or directly offer a fix in a PR.

PR #166, opened 2018-03-21
added unsafeStyleElement for unrepresented styles

GTDev87:


arnarthor:
I'm really unsure about something like this. The reason we have typesafe language is to not need to have unsafe code and fast iteration loop and this is potentially a huge source of bugs in both of those regards.

What's your take on this @wokalski?

GTDev87:
Let me link back to the issue https://github.com/reasonml-community/bs-react-native/issues/165

arnarthor:
Yeah, I know that, but we should rather expose these method as heightAsUndefined or something like that to have it explicit. This is just a global escape hatch which is really dangerous

wokalski:
Not a fan, it's been mentioned several times. I agree with @arnarthor suggestion. heightAsUndefined (or maybe imageAutomaticHeight to reflect its purpose) is what I'd suggest here, too.

wokalski:
Hm, looks like everyone is on the same page.

PR #172, opened 2018-04-19
Upgrade bs-platform and fix Js.bool related fixes

Gregoirevda:
I did a `yarn add -D bs-platform` which is now on `3.0.0` and it has support for Js.booleans natively.
I now have a lot of warnings of unnecessary Js.bool casting functions


src/private/encode.re
```diff
@@ -8,7 +8,7 @@ external float : float => Js.Json.t = "%identity"; /* [float n] makes a JSON num
 external int : int => Js.Json.t = "%identity"; /* [int n] makes a JSON number of the [int] [n] */
 
 /** [boolean b] makes a JSON boolean of the [Js.boolean] [b] */
-external boolean : Js.boolean => Js.Json.t = "%identity"; /* [boolean b] makes a JSON boolean of the [Js.boolean] [b] */
+external boolean : bool => Js.Json.t = "%identity"; /* [boolean b] makes a JSON boolean of the [Js.boolean] [b] */
```
Gregoirevda: Not sure if the comment is still ok

src/components/textInput.re
```diff
@@ -1,9 +1,8 @@
 [@bs.module "react-native"]
 external view : ReasonReact.reactClass = "TextInput";
 
-[@bs.send] external _isFocused : ReasonReact.reactRef => Js.boolean = "";
+[@bs.send] external isFocused : ReasonReact.reactRef => bool = "";
 
-let isFocused = UtilsRN.(Js.to_bool << _isFocused);
```
Gregoirevda: Please double check this one, not sure what UtilsRN. is doing

grsabreu: I think it is just doing pointfree composition

src/private/encode.re
wokalski: It should be [boolean b] makes a JSON boolean of the [bool] [b]

grsabreu:
Should we bump the bindings to 1.0.0? As this is a breaking change for previous bs-platform versions
cc @wokalski

wokalski:
yes, a bindings release is long overdue. I think we should 😀. I'm sort of bummed out because we planned to have migration scripts for all breaking changes but we still don't have infrastructure for that @grsabreu .

gunnigylfa:
Would it be possible to use the reason-react migration script? https://github.com/chenglou/upgrade-reason-react
If that works then we could just add that step to the release notes.

vnovick:
It's nice to have a migration script, but does not having it really blocks us from migrating the change?
Js.boolean change to bool according to @chenglou and @jaredly are not a breaking change. 
@chenglou , @jaredly , What's your insights on this?

arnarthor:
I agree. It would be great to have a migration script, but I don't think it should be a blocker for updating this.

chenglou:
The migration script would be going the extra mile, which is great! This one’s simple though. Could be a few regexes. The heavier migration infra is also in place, and we’ve used it twice for ReasonReact already, so no worries. But again, this can just be a regex change. 

The Boolean changes isn’t breaking (all previous helpers are now no-ops).

wokalski:
@chenglou thankfully we don't need your permission ;).

In any case, let's merge it.

chenglou:
I’m just chiming in since this seems to have stalled a bit

wokalski:
My comment was unnecessary.

I got biased by your controlling behaviour in person and this was an impolite intuitional answer. Anyway, I need to set the standards higher.

chenglou:
All good. I commented because another member from the community asked me to

PR #178, opened 2018-04-27
change JS name of Animated.Easing.in_ function

mcanobbio:
I've been having problems to use the Animated.Easing.in_ and I realized there is an error on your bindings, you are trying to call the JS function like this  `in_` and should be `in`


arnarthor:
LGTM thanks!

PR #182, opened 2018-05-04
Fix incorrect skew bindings

### LINKED ISSUE #181: Incorrect skew bindings
image

medson10:
Closes #181


grsabreu:
Works and is compliant with the React Native docs https://facebook.github.io/react-native/docs/transforms.html#transform

PR #185, opened 2018-05-14
Adds bindings for WebView methods

kristinns:
Adds bindings for `goForward()`, `goBack()`, `reload()` and `stopLoading()` on WebView.
These methods are still missing in the React-Native documentation but here is a PR to fix that https://github.com/facebook/react-native-website/pull/363


arnarthor:
LGTM

PR #187, opened 2018-05-16
Add sourceHtml and rename source to sourceUri in WebView bindings

kristinns:
Renames `WebView.source` method to `WebView.sourceUri` and adds missing `WebView.sourceHtml`

### DIFF
--- a/src/components/webView.re
+++ b/src/components/webView.re
@@ -15,12 +15,22 @@ type iOSLoadRequestEvent = {
 type source;
 
 [@bs.obj]
-external source :
+external sourceUri :
   (
     ~uri: string=?,
     ~method: string=?,
     ~headers: Js.t('a)=?,
     ~body: string=?,
+    unit
+  ) =>
+  source =
+  "";
+
+[@bs.obj]
+external sourceHtml :
+  (
+    ~html: string=?,
+    ~baseUrl: string=?,
     unit
   ) =>
   source =
--- a/src/components/webView.re
+++ b/src/components/webView.re
@@ -36,6 +36,8 @@ external sourceHtml :
   source =
   "";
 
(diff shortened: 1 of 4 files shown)


wokalski:
@kristinns would using `bs.deriving abstract` work here? It's a bit clearer to read.

kristinns:
@wokalski that would be great but I'm not sure how I could do that.
Mainly because `WebView.sourceUri` and `WebView.sourceHtml` create different JS types but we are telling the reason compiler that this is the same type.
Do you know how this could be done with `bs.deriving abstract`?

src/components/webView.rei
```diff
@@ -1,8 +1,11 @@
 type source;
 
-let source:
```
wokalski: This is a breaking change. We can make it non breaking by aliasing `source` to `sourceUri` and marking it as deprecated

kristinns: Good point, fixed in dec94b7

src/components/webView.re
```diff
@@ -26,6 +26,19 @@ external source :
...
+external sourceHtml :
+  (
+    ~html: string=?,
+    ~baseUrl: string=?,
+    unit
+  ) =>
+  source =
+  "";
+
+[@deprecated "Please use WebView.sourceUri instead"]
```
wokalski: This is only needed in the `rei`. Also did u test it with the RNTester to verify? I never trust reason deprecated annotations :P

kristinns: Ah ok, fixed in dec94b7 (amended to the commit)
And yes the deprecation worked, got warning message
`deprecated: BsReactNative.WebView.source, Please use WebView.sourceUri instead`

PR #189, opened 2018-05-21
Fix WebView types

### LINKED ISSUE #186: Not sound WebView `onNavigationStateChange` bindings
The bindings declare that `onNavigationStateChange` receives options https://github.com/reasonml-community/bs-react-native/blob/master/src/components/webView.rei#L121 from JS and there is no conversion in the code, so they should be `Js.Null_undefined.t`. But I also think they are never null/undefined.
I'm going to send a PR fixing this.

grsabreu:
Closes #186 

Bindings following https://github.com/facebook/react-native/blob/master/ReactAndroid/src/main/java/com/facebook/react/views/webview/ReactWebViewManager.java and https://github.com/facebook/react-native/blob/master/React/Views/RCTWebView.m

### DIFF
--- a/src/components/webView.re
+++ b/src/components/webView.re
@@ -1,15 +1,22 @@
 [@bs.module "react-native"] external view : ReasonReact.reactClass = "WebView";
 
-type iOSLoadRequestEvent = {
+type navigationType = [
+  | [@bs.as "click"] `click
+  | [@bs.as "formsubmit"] `formsubmit
+  | [@bs.as "backforward"] `backforward
+  | [@bs.as "reload"] `reload
+  | [@bs.as "formresubmit"] `formresubmit
+  | [@bs.as "other"] `other
+];
+
+type iOSOnShouldStartLoadWithRequestEvent = {
   .
-  "target": int,
-  "canGoBack": bool,
-  "lockIdentifier": int,
-  "loading": bool,
-  "title": string,
-  "canGoForward": bool,
-  "navigationType": string,
   "url": string,
+  "title": string,
+  "loading": bool,
+  "canGoBack": bool,
+  "canGoForward": bool,
+  "navigationType": navigationType,
 };
 
 type source;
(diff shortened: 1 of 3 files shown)


grsabreu:
I'm going to fix the conflicts soon

grsabreu:
I'm having some troubles generating the interface files

```
./node_modules/.bin/bsc -pp 'refmt --print binary' -i -I lib/bs/src/ -I lib/bs/src/components -I ./node_modules/reason-react/lib/bs/src/ -impl src/components/webView.re
```

getting this output

```
File "src/components/webView.re", line 4, characters 6-11:
Warning 101: Unused BuckleScript attribute: bs.as
File "src/components/webView.re", line 5, characters 6-11:
Warning 101: Unused BuckleScript attribute: bs.as
File "src/components/webView.re", line 6, characters 6-11:
Warning 101: Unused BuckleScript attribute: bs.as
File "src/components/webView.re", line 7, characters 6-11:
Warning 101: Unused BuckleScript attribute: bs.as
File "src/components/webView.re", line 8, characters 6-11:
Warning 101: Unused BuckleScript attribute: bs.as
File "src/components/webView.re", line 9, characters 6-11:
Warning 101: Unused BuckleScript attribute: bs.as
File "src/components/webView.re", line 172, characters 8-24:
Error: Unbound module Props
```

src/components/webView.re
```diff
@@ -1,15 +1,22 @@
 [@bs.module "react-native"] external view : ReasonReact.reactClass = "WebView";
 
-type iOSLoadRequestEvent = {
+type navigationType = [
+  | [@bs.as "click"] `click
+  | [@bs.as "formsubmit"] `formsubmit
+  | [@bs.as "backforward"] `backforward
+  | [@bs.as "reload"] `reload
+  | [@bs.as "formresubmit"] `formresubmit
+  | [@bs.as "other"] `other
```
arnarthor: The warnings you're seeing in refmt are because these `@bs.as` are not used without `[@bs.deriving jsConverter]` on the type.

src/components/webView.re
```diff
@@ -1,15 +1,22 @@
...
-  "lockIdentifier": int,
-  "loading": bool,
+  "url": string,
   "title": string,
+  "loading": bool,
+  "canGoBack": bool,
   "canGoForward": bool,
-  "navigationType": string,
-  "url": string,
+  "navigationType": navigationType,
```
arnarthor: I think it would be better to inline the poly variant `navigationType` here, then you should be able to use `[@bs.string]` if I remember correctly.

src/components/webView.re
```diff
@@ -60,10 +67,63 @@ let make =
...
+         option(
+           {
+             .
+             "url": string,
+             "title": string,
+             "loading": bool,
+             "canGoBack": bool,
+             "canGoForward": bool,
+             /* iOS only */
+             "domain": Js.Null_undefined.t(string),
```
arnarthor: `Null_undefined` is AFAIK deprecated. We should probably use `Js.Nullable` instead

src/components/webView.re
grsabreu: hmm I tried bs.string and I still get the `unused` warning

arnarthor: I _think_, but I might be mistaken here, that `"navigationType": [@bs.string] [ | `reload | `formresubmit ...]` should work.

But then you need to inline it, and you don't need the `[@bs.as "..."]` for each variant since the values are the same. `@bs.string` defaults to converting it to the exact string without the back tick.

grsabreu:
@arnarthor afaik bs.string will only work together with `external`, so I left `navigationType` as string and also let the user call `navigationFromJs` helper from  bs.deriving

PR #191, opened 2018-05-23
Fixes event types for WebView

kristinns:
This broke with the bs platform update ( because of the Js.Bool change )
So this fixes the bindings for the WebView


kristinns:
Nvm, this doesn't seem to be a problem. We were using old bs-platform version, that's why it didn't work

PR #195, opened 2018-05-31
Fix whileEditing to use kebab-case

shritesh:
`clearButtonMode` in TextInput uses kebab-case.
https://facebook.github.io/react-native/docs/textinput.html#clearbuttonmode


nhducit:
thanks @shritesh

PR #200, opened 2018-06-27
Fixes webview event types

Rikkiabb:
Fixes bindings for event types in WebView

### DIFF
--- a/lib/js/src/components/webView.js
+++ b/lib/js/src/components/webView.js
@@ -9,6 +9,8 @@ var ReactNative = require("react-native");
 var Props$BsReactNative = require("../private/props.js");
 var UtilsRN$BsReactNative = require("../private/utilsRN.js");
 
+var EventTypes = /* module */[];
+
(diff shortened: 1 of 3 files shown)


ulrikstrid:
Should this be modelled with deriving abstract instead?

Rikkiabb:
Sure, how would you do that?

ulrikstrid:
The type would look something like this:
```ocaml
[@bs.deriving abstract]
type t = {
  [@bs.option] url: string,
  [@bs.option] title: string,
  [@bs.option] loading: bool,
  [@bs.option] canGoBack: bool,
  [@bs.option] canGoForward: bool
}
```
and then it can be used like this: `let config = t(~url: "https://google.com", ());`

ulrikstrid:
I think this looks good cc @wokalski

wokalski:
I think `/lib` hasn't been updated.

wokalski:
Other than that, it's super mergeable 👍

wokalski:
Thanks!

PR #202, opened 2018-07-05
ImagePickerIOS bindings

reinvanimschoot:
Bindings for ImagePickerIOS API

- [x] **canRecordVideos**
- [x] **canUseCamera**
- [x] **openSelectDialog**
- [ ] **open CameraDialog**

I got some problems with **openCameraDialog**. When trying to open the dialog on a simulator I got:

screen shot 2018-06-28 at 15 38 26

Now, this could be due to the app being run on a simulator. However, when I try to run it on my phone, I get another error in Xcode:

screen shot 2018-06-28 at 17 48 54

So not sure what is going on. If someone would like to take a look, that would be appreciated!


src/imagePickerIOS.rei
```diff
@@ -0,0 +1,15 @@
+let canUseCamera: (unit => unit) => unit;
+
+let canRecordVideos: (unit => unit) => unit;
+
+[@bs.deriving abstract]
+type error = {
```
wokalski: For the rei, would it be possible to expose only the abstract `type error` and the getters?

src/imagePickerIOS.rei
```diff
@@ -0,0 +1,15 @@
+let canUseCamera: (unit => unit) => unit;
```
wokalski: `can*` functions' callbacks here, have booleans as the first arg (see the [implementation here](https://github.com/facebook/react-native/blob/master/Libraries/CameraRoll/RCTImagePickerManager.m#L42))

RNTester/src/examples/ImagePickerIOSExample.re
```diff
@@ -0,0 +1,168 @@
...
+  let make = _children => {
+    ...component,
+    initialState: () => {canUseCamera: false},
+    reducer: (action: action, _state: state) =>
+      switch (action) {
+      | UpdateCanUseCamera(canRecordVideos) =>
+        Update({canUseCamera: canRecordVideos})
+      },
+    render: ({handle}) =>
+      <View style=styles##pageStyle key="1">
```
wokalski: Are those keys necessary?

wokalski:
regarding the simulator issue: https://stackoverflow.com/questions/13564027/source-type-1-not-available
Regarding the latter issue: https://stackoverflow.com/questions/39465687/nscamerausagedescription-in-ios-10-0-runtime-crash

grabbou:
Closing as ImagePickerIOS has already landed in master as a part of Rebolt merging. Thank you for your contributions!

PR #207, opened 2018-08-17
Adds Animated.FlatList

kristinns:
- Adds `CreateComponent` functor to FlatList so that it can be used with both `FlatList` and `Animated.FlatList`
- Adds `FlatList` to `AnimatedComponents` module and to `Animated` module

FlatList has not yet been added to the Animated library in React-Native but I guess it will be added in the next release since this [commit](https://github.com/facebook/react-native/commit/daa7c78055857cd2d9ea650de0c4b0f72d3f2acf#diff-35653b5f6360716ec796e5ffdd483841) is already merged.

If you don't like having it in the `AnimatedComponents` module until it's released in React-Native I can remove it from there and only add the `CreateComponent` functor for `FlatList`

### DIFF
--- a/lib/js/src/animated.js
+++ b/lib/js/src/animated.js
@@ -49,6 +49,8 @@ var View = 0;
 
 var ScrollView = 0;
 
+var FlatList = 0;
+
 exports.Animation = Animation;
 exports.CompositeAnimation = CompositeAnimation;
 exports.Easing = Easing;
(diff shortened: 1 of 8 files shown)


wokalski:
I think we’re good to merge. We don’t have tools to sync with JS lib releases so whatever we do is potentially wrong. For that reason I prefer to merge things early.

kristinns:
I agree, should we merge this then @wokalski ?

PR #209, opened 2018-08-22
Fix FlatList scrollToOffset

phated:
I've been using these changes in my project otherwise `scrollToOffset` wasn't working.


grabbou:
What do you mean by `it wasn't working`? You were not able to provide float values? Or the method wasn't working at all

grabbou:
After chatting with the core team, we figured that offset can have float values (at least on iOS)

phated:
To clarify "doesn't work" - the method being called didn't exist and when I changed it, the onScroll handler receives a float in the event that can't be passed directly to `scrollToOffset`.

Regarding floats - I'm using floats on android 7.1.2 and it works perfectly.  If I round the float, the view doesn't scroll to the exact position again.

grabbou:
Going to merge it as I confirmed float to be the correct payload.

wokalski:
Thank you guys!

grabbou:
Thanks @wokalski! Wanted to merge it, but eventually realized it was impossible from my mobile device :(

PR #210, opened 2018-08-27
Fixes onmessage type

Rikkiabb:
Was using OnMessage in a WebView and noticed it is wrongly typed. This fixes it.

### DIFF
--- a/src/components/webView.rei
+++ b/src/components/webView.rei
@@ -106,7 +106,7 @@ let make:
     ~injectJavaScript: string => unit=?,
     ~injectedJavaScript: string=?,
     ~mediaPlaybackRequiresUserAction: bool=?,
-    ~onMessage: RNEvent.NativeEvent.t => unit=?,
+    ~onMessage: RNEvent.t => unit=?,
     ~onNavigationStateChange: EventTypes.t => unit=?,
     ~scalesPageToFit: bool=?,
     ~startInLoadingState: bool=?,


ulrikstrid:
I'm not sure this is correct when reading the docs here: https://facebook.github.io/react-native/docs/webview#onmessage

But I think that this line: https://github.com/reasonml-community/bs-react-native/blob/master/src/RNEvent.rei#L14 should just be a string.

I might be wrong as I haven't done any react-native lately this is just from reading the docs.

Rikkiabb:
I have tested this out and I think that the type on OnMessage is incorrect. I believe this should be the type: https://github.com/reasonml-community/bs-react-native/blob/master/src/RNEvent.rei#L50

Rikkiabb:
I have tested this out and I think that the type on OnMessage is incorrect. 
I think the type should be: https://github.com/reasonml-community/bs-react-native/blob/master/src/RNEvent.rei#L50 
And then you can use OnMessage like this: 
``` 
          onMessage={
            event =>
              send(
                OnMessageReceived(
                  event->RNEvent.nativeEvent,
                ),
              )
          }
``` 
If you would have the same function and have it as it is then you get the type back but everything on it comes back as `undefined`. But with this PR I get values on it. 
But of course I could be wrong but with the current implementation it is not working and it doesn't work with changing `data` on the `NativeEvent` to string because everything comes back as `undefined`.

PR #211, opened 2018-08-29
Rebolt: Animated

grabbou:
This PR contains few changes that align with Animated module better, mainly better types for different Value and ValueXY and allowing math operations on already interpolated values.

It is backwards compatible with depreciation warnings.


grabbou:
CC: @wokalski this is first PR, so it may require some additional work from our sides to make sure its smooth. Rest should be just "misc", this was the biggest change in our code.

src/animatedRe.rei
```diff
@@ -1,46 +1,29 @@
+type calculated;
+
+type regular;
+
+type value('a);
+
+type valueXY('a);
```
wokalski: woohoo! Finally. Great change.

src/animatedRe.rei
```diff
@@ -225,35 +215,102 @@ module ValueXY: {
...
 
-let loop: (~iterations: int=?, ~animation: CompositeAnimation.t, unit) => CompositeAnimation.t;
+[@deprecated "Please use Easing module instead"]
+module Easing = Easing;
 
-let createAnimatedComponent : ReasonReact.reactClass => ReasonReact.reactClass;
+[@deprecated
+  "Please use Animated.start, Animated.stop or Animated.reset instead"
+]
+module CompositeAnimation = Animation;
```
wokalski: Is it backwards compatible to some extent? (if not, that's cool, too)

src/animatedRe.rei
```diff
@@ -225,35 +215,102 @@ module ValueXY: {
...
 
+[@deprecated "Please use Animated.timing instead"]
 module Timing = Value.Timing;
 
 module TimingXY = ValueXY.Timing;
```
wokalski: Do we want the deprecation comments on `*XY` modules, too?

src/easing.rei
```diff
@@ -0,0 +1,18 @@
+type t = float => float;
```
wokalski: That's cool.

src/animatedRe.rei
grabbou: Yeah, it's the same set of methods, so should be fine.

src/animatedRe.rei
grabbou: Good call.

grabbou: In this particular case, I ended up not deprecating the `*XY` modules because `spring`, `timing` and `decay` (top-level methods on this module) are for `Value`, not for `ValueXY` (as per React Native interface itself). To be honest, the `XY` to my knowledge after chatting with Facebook team is considered being deprecated in the future, so I wouldn't worry too much about it.

grabbou:
Thank you @wokalski for your review - I am really happy that, overall, you like the changes we have made! Can't wait to migrate all our apps to use `bs-react-native` from now on!

PR #213, opened 2018-08-29
AlertIOS

knowbody:
This fixes the current bindings for `AlertIOS`, as the RN API doesn't have `options` I removed it and also there were typos in the `type` variant to string mapper.


grabbou:
To be merged after #214

ulrikstrid:
I'm trying to go through the Rebolt changes and get them merged

@grabbou or @knowbody would you mind fixing the conflict? It looks like it should be straight forward but I'm not sure

knowbody:
@ulrikstrid done

PR #214, opened 2018-08-29
Rebolt: Update all remaining components

grabbou:
This is catch-all PR that does not contain any major changes to the existing APIs/components. Just adding new types and making sure the remaining ones are matching what RN ships with. If removal of `optional` is breaking your codebase, it most likely means that you might be using RN API in a wrong way.

We also updated `bs-react-native` to 4.x and performed necessary changes.

Due to the nature of the update, it is easier to ship all updates in a single PR.


grabbou:
Thank you!

PR #225, opened 2018-08-30
Replace functions duplicating Bucklescript std lib

sgny:
Redefined functions in `utilsRN.re` as aliases to functions in `Js.Dict` in case anybody wants to keep referencing them, and replaced the calls to them in `style.re`.

I closed another pull request based on an earlier version of `style.re`.

### DIFF
--- a/src/private/utilsRN.re
+++ b/src/private/utilsRN.re
@@ -15,48 +15,10 @@ external objAssign2 : (Js.t({..}), Js.t({..}), Js.t({..})) => Js.t({..}) =
 
 let (<<) = (f, g, x) => f(g(x));
 
-/***
- * The following is taken from bs-json (https://github.com/BuckleTypes/bs-json) converted to reason syntax
- */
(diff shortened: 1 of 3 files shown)


src/private/utilsRN.re
```diff
@@ -15,48 +15,10 @@ external objAssign2 : (Js.t({..}), Js.t({..}), Js.t({..})) => Js.t({..}) =
...
-let dictFromArray = entries => {
-  let dict = Js.Dict.empty();
-  let l = Js_array.length(entries);
-  for (i in 0 to l - 1) {
-    let (key, value) = Array.unsafe_get(entries, i);
-    Js.Dict.set(dict, key, value);
-  };
-  dict;
-};
+let dictFromArray = Js.Dict.fromArray;
```
wokalski: Since utils are private, we should remove those aliases I think.

wokalski:
superseded by #235

PR #227, opened 2018-08-30
Clean up references to optBoolToOptJsBoolean since that is the identity function

sgny:
Cleaning up `bool-Js.boolean` conversion artifacts.


wokalski:
Thanks for those PRs @sgny is there a chance you could submit them together (and possibly squash the commits)?

sgny:
Sure, I'll take care of that later today.

PR #228, opened 2018-08-31
Adds imagePickerIOS component bindings from Rebolt.

Trancever:
Brings imagePickerIOS component bindings from Rebolt.


src/imagePickerIOS.rei
```diff
@@ -0,0 +1,14 @@
...
+
+let canRecordVideos: (unit => unit) => unit;
+
+type error = {
+  .
+  "code": int,
+  "message": string,
+};
+
+let openCameraDialog: (~videoMode: bool, unit => unit, error => unit) => unit;
```
wokalski: I think all these args should be labelled. WDYT?

Trancever: Yeah, that's good idea. Are onSuccess and onError good labels?

wokalski: yes! It'd be cool do it for both open* functions

Trancever: Yeah, it should be done right now.

wokalski:
Nice, thanks.

PR #229, opened 2018-08-31
Add ImageStore component bindings from Rebolt.

Trancever:
Adds ImageStore component bindings from Rebolt.


src/imageStore.rei
```diff
@@ -0,0 +1,9 @@
+type error;
+
+let hasImageForTag: (string, bool => unit) => unit;
+
+let removeImageForTag: string => unit;
+
+let addImageFromBase64: (string, string => unit, error => unit) => string;
+
+let getBase64ForTag: (string, string => unit, error => unit) => unit;
```
wokalski: For the sake of documentation we could alias string to `base64Image` like this: `type base64Image = string`.

so we would get:

```
let getBase64ForTag: (string, base64Image => unit, error => unit) => unit;
```

WDYT?

wokalski:
Looks great! Thanks.

PR #231, opened 2018-08-31
Add bindings to settings module from Rebolt.

Trancever:
Add bindings to settings module from Rebolt.


src/settings.rei
```diff
@@ -0,0 +1,7 @@
+let get: string => string;
+
+let set: Js.Dict.t(string) => unit;
+
+let watchKeys: (list(string), unit => unit) => int;
+
+let clearWatch: int => unit;
```
wokalski: This is a great place to assure more safety. By making `int` an abstract type in `rei` we can make sure that only `watchToken` or whatever we call it is cleared. So we can do:
```
type watchToken;

let watchKeys: (list(string), unit => unit) => watchToken;

let clearWatch: watchToken => unit;
```

Trancever: That's awesome idea, I totally agree with you 👍

wokalski:
Awesome, thanks!

PR #232, opened 2018-08-31
Add `Share` module

Trancever:
Add bindings for share module from Rebolt.


src/share.rei
```diff
@@ -0,0 +1,29 @@
...
+  (~title: string=?, ~message: string, ~url: string) => makeContent;
+let makeOptions:
+  (
+    ~subject: string=?,
+    ~tintColor: string=?,
+    ~excludedActivityTypes: array(string)=?,
+    ~dialogTitle: string=?
+  ) =>
+  makeOptions;
+let _share: (makeContent, makeOptions) => Js.Promise.t(bool);
```
wokalski: This wasn't intended was it?

src/share.rei
```diff
@@ -0,0 +1,29 @@
...
+let share:
+  (
+    ~content: content,
+    ~subject: string=?,
+    ~tintColor: string=?,
+    ~excludedActivityTypes: array(string)=?,
```
wokalski: Are activity types a well defined set? If yes, maybe this could be a variant.

grabbou:
Updated with better API. @wokalski, answering one of your requested changes, we cannot unfortunately type `excludedActivityTypes` prop as its meant to be literally an array of strings. You use it to define a list of `share providers` to disable when iOS is about to show its share menu (e.g. no `post to twitter`).

wokalski:
👍

PR #234, opened 2018-08-31
Add toastAndroid module bindings from Rebolt.

Trancever:
Add toastAndroid module bindings from Rebolt.


src/toastAndroid.rei
```diff
@@ -0,0 +1,17 @@
+type duration =
+  | BOTTOM
+  | CENTER
```
wokalski: is duration `CENTER` really a thing :D ?

src/toastAndroid.re
```diff
@@ -0,0 +1,22 @@
...
+
+[@bs.module "react-native"] [@bs.scope "ToastAndroid"]
+external show : (string, duration) => unit = "";
+
+[@bs.module "react-native"] [@bs.scope "ToastAndroid"]
+external showWithGravity : (string, duration, gravity) => unit = "";
+
+[@bs.module "react-native"] [@bs.scope "ToastAndroid"]
+external showWithGravityAndOffset :
+  (string, duration, gravity, int, int) => unit =
```
wokalski: do we really pass the variants as is to JS?

src/toastAndroid.rei
Trancever: Fixed :D

src/toastAndroid.re
Trancever: Yeah, bindings were totally broken, fixed them already so you can check once again :)

wokalski:
It looks good to me although I believe we should use labels for the x/y offsets.

grabbou:
I have updated the interface. Unfortunately, due to really bad API from the React Native side (multiple `show` methods with no clear default parameter), I was unable to provide a single, unified, entry-point with optional labels.

Decided to follow the Animated API where the first argument is provided and rest that follows are labels. I think it matches the overall API pretty nice.

grabbou:
We could provide a unified `show` method but in case the user provides offset and no gravity, there's no native React Native method to call. In that case, we would have to assume the default value for "gravity" which is unclear and not described in React Native docs. Question is, whether we should make such default for better API sake?

Essentially, it would be a single switch/case statement checking on the arguments provided and calling the appropriate native method.

grabbou:
Here's the method with alternative API:
```reason
let betterShow = (message, ~duration, ~gravity=?, ~offset=?, ()) =>
	switch ((gravity, offset)) {
	| (None, None) => _show(message, getDuration(duration))
	| (Some(gravity), None) => _showWithGravity(message, getDuration(duration), getGravity(gravity))
	| (Some(gravity), Some((x, y))) => _showWithGravityAndOffset(message, getDuration(duration), getGravity(gravity), x, y)
	| (None, Some((x, y))) => _showWithGravityAndOffset(message, getDuration(duration), getGravity(CENTER), x, y)
	}
```
Note that the last case assumes "`CENTER`" is the default when offset is set. We don't have `showWithOffset` method on the native side and so, we need to put a default.

PR #233, opened 2018-08-31
Add timePickerAndroid module bindings from Rebolt.

Trancever:
Add timePickerAndroid module bindings from Rebolt.

PR #237, opened 2018-09-05
Update datePickerAndroid bindings

rodan-lewarx:
datePickerAndroid bindings now use bs.deriving abstract and bs.deriving jsConverter


src/datePickerAndroid.re
```diff
@@ -19,48 +17,45 @@ type action =
...
-    "minDate": Js.Nullable.fromOption(minDate),
-    "maxDate": Js.Nullable.fromOption(maxDate),
-    "mode":
-      switch (mode) {
-      | Default => "default"
-      | Calendar => "calendar"
-      | Spinner => "spinner"
-      },
-  })
+let open_ = (~date: Js.Date.t, ~minDate=?, ~maxDate=?, ~mode=`default, ()) =>
```
wokalski: cool!

src/datePickerAndroid.rei
```diff
@@ -8,10 +8,7 @@ type action =
   | Dismissed
   | Set(response);
 
-type mode =
-  | Calendar
-  | Spinner
-  | Default;
+type mode = [ | `calendar | `spinner | `default];
```
wokalski: I'm not sure about this change. I don't think it's justified to change the public API in this case.

rodan-lewarx: @wokalski Sorry I didn't think about that. I am just using abstract and jsConverter everywhere. 

I just wanted to do the same I did for timePickerAndroid https://github.com/reasonml-community/bs-react-native/pull/233 If we are not using jsConverter here, should I change timePickerAndroid to use the same pattern as used here?

wokalski: Hmm, ok, that's a good point. Let's make this breaking change then!

wokalski:
Thanks a lot @rodan-lewarx 👍

PR #241, opened 2018-09-17
Release v0.10.0-rc.0

grabbou:
Can't make a direct push, so making a PR. 

Pushed 0.10.0-rc.0 and tagged it as `next`. We will use this one to test `Rebolt` integration and many new features before promoting it to stable (similar to React Native lifecycle).

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
 	"name": "bs-react-native",
-	"version": "0.9.0",
+	"version": "0.10.0-rc.0",
 	"scripts": {
 		"build": "bsb -make-world",
 		"start": "bsb -make-world -w",


wokalski:
the built has failed 😢

grabbou:
Oh :(

What's the `push` build supposed to do? Can't find any log information in there. The `pr` one seems to be passing.

wokalski:
btw, we dont' want to have direct pushes allowed. We are accepting everybody to have write access so this gives us at least some security.

grabbou:
Yeah, I noticed. You could probably add me to `whitelisted` people with
push access to master, I am sure no-one would mind xd Either way, it's
fine. I can still push tags and make a release, so it's not that disturbing
as you'd think.

PR #243, opened 2018-09-19
Adds types to textinput

Rikkiabb:
I had to create bindings against the native input and thought it kind off sucked. So this PR is a suggestion of how it could be better to work with.
I think it is cleaner and simpler to maintain. 
Not sure if there is a better way of doing this, if so then I would love feedback on it.

So before I had to do this: 
screen shot 2018-09-19 at 14 43 06

But after I can simply do this:
screen shot 2018-09-19 at 14 42 21


grabbou:
Thanks for this pull request. Would you mind rebasing on top of master? We have changed quite a few things (including CircleCI support) and it's necessary for this PR to be updated before we can merge it.

src/components/textInput.re
```diff
@@ -9,6 +9,141 @@ external view : ReasonReact.reactClass = "TextInput";
...
 
+module AutoCapitalize = {
+  type t = [
+    | `none
+    | `sentences
+    | `words
+    | `characters
+  ];
+let to_string = x => {
+        switch (x) {
```
rodan-lewarx: You could use bs.deriving jsconverter instead of writing to_string methods everywhere https://bucklescript.github.io/docs/en/generate-converters-accessors

lib/js/src/components/textInput.js
```diff
@@ -1,4 +1,4 @@
-// Generated by BUCKLESCRIPT VERSION 4.0.5, PLEASE EDIT WITH CARE
+// Generated by BUCKLESCRIPT VERSION 3.1.5, PLEASE EDIT WITH CARE
```
wokalski: Please update your bucklescript versioning regenerate this file.

PR #244, opened 2018-09-19
Create React Native template

### LINKED ISSUE #35: Create bsb theme
Now that Bucklescript supports themes / templates, we should pitch one for React Native :)

grabbou:
Fixes #35 

You can test it by running:
```
react-native init HelloWorld --template reason
```

It has some limitations (all of them described in the project's readme) that we are working on in the React Native.

### DIFF
--- a/generator/README.md
+++ b/generator/README.md
@@ -1,0 +1,47 @@
+# HelloWorld
+
+> This is a React Native application developed in Reason.
+
+## Installation
+
+To get started, add the following scripts to your `package.json`:
+
+```json
+"scripts": {
+  "start": "node node_modules/react-native/local-cli/cli.js start",
+  "build": "bsb -make-world -clean-world",
+  "watch": "bsb -make-world -clean-world -w"
+}
+```
+
+Next, run `watch` and `start` scripts from the command line so that both processes are running at the same time. This can be done with `yarn` (e.g. `yarn watch`) or `npm` (e.g. `npm run watch`).
+
+The first script compiles you Reason code in a watch mode, meaning that every change is instantly recompiled.
+
+Second one is a React Native packager that serves the compiled Reason code to the React Native client.
+
+## Hiding `.bs.js` files
+
+Bucklescript compiles your Reason files to Javascript `in source`, meaning that `App.re` will result in `App.bs.js` file generated next to it.
+
+These are build files and are by default ignored from the git repository.
+
+If you are using VSCode, you can add the following to your workspace settings to hide them automatically:
+
+```json
+{
+  "files.exclude": {
+    "**/*.bs.js": true
+  }
+}
+```
+
+## Breakdown of existing files
+
+The `react-native init` command generates a lot of boilerplate with files that might be not needed for your particular use case. Below is a breakdown of all top-level files to help you decide:
+
+- `.babelrc` - required, for transpiling Javascript files to supported by the mobile engine Javascript version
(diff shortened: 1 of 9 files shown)


grabbou:
In the future, we need to update the `template` command in a way that allows us to remove "default" files and do even more than that.

Right now, the `template` works by running `react-native init` with default Javascript / Flow settings and then, it applies our configuration on top of it. As a result, we don't have control over what the content of `package.json` script will be etc.

The `init` command uses Yeoman, so I am going to look into tweaking it to give us a bit more functionality.

grabbou:
Previously published as `rebolt-generator`, this repository is now being moved here.

generator/dependencies.json
```diff
@@ -0,0 +1,5 @@
+{
+  "bs-platform": "^4.0.0",
+	"reason-react": "^0.4.2",
```
Trancever: There is a small indentation issue

generator/README.md
```diff
@@ -0,0 +1,47 @@
...
+"scripts": {
+  "start": "node node_modules/react-native/local-cli/cli.js start",
+  "build": "bsb -make-world -clean-world",
+  "watch": "bsb -make-world -clean-world -w"
+}
+```
+
+Next, run `watch` and `start` scripts from the command line so that both processes are running at the same time. This can be done with `yarn` (e.g. `yarn watch`) or `npm` (e.g. `npm run watch`).
+
+The first script compiles you Reason code in a watch mode, meaning that every change is instantly recompiled.
```
Trancever: Should be "your" instead of "you"

generator/README.md
```diff
@@ -0,0 +1,47 @@
...
+
+Second one is a React Native packager that serves the compiled Reason code to the React Native client.
+
+## Hiding `.bs.js` files
+
+Bucklescript compiles your Reason files to Javascript `in source`, meaning that `App.re` will result in `App.bs.js` file generated next to it.
```
Trancever: How about adding information that user can change bsconfig.json to generate .js files in lib directory?

grabbou: Oh, then... how about we just get rid of that setting by default and remove this notice?

generator/dependencies.json
grabbou: Right. For some reason, my VSCode keeps falling back to tabs.

generator/README.md
Trancever: I am ok with both options :D

grabbou:
Fixed and tested. Going to merge it now once the build is green.

grabbou:
@Trancever need your approval once again

generator/package.json
```diff
@@ -0,0 +1,4 @@
+{
+  "name": "react-native-template-reason",
+  "version": "0.0.8"
```
czystyl: Check the second section of https://semver.org/#faq

grabbou:
True. To be honest, `semver` doesn't really matter in this case, but I
think we can make it a 1.0.0.

generator/dependencies.json
```diff
@@ -0,0 +1,5 @@
+{
+  "bs-platform": "^4.0.0",
+  "reason-react": "^0.4.2",
```
MoOx: Any reason to stick to 0.4 as 0.5+ is out there?

grabbou:
Tried to match what's in bs-react-native's own package.json. We can update
them both.

MoOx:
Oh right, good catch :)
We can let it like that and upgrade both after.

PR #248, opened 2018-09-21
Improve `Style` bindings with `Animated`

### LINKED ISSUE #246: BsReactNative.AnimatedRe.calculated
Was updating my project to the latest commit and ran into problems with the new type `AnimatedRe.calculated`. I don't get why the change to `AnimatedRe.calculated` and `AnimatedRe.regular`, since `React native` just worries about if the variable is animated or not?
I would like a way to change a `AnimatedRe.calculated` to `AnimatedRe.regular` or have this as the same type.

grabbou:
This PR updates Style module to reduce complexity when working with `Animated` values. Fixes #246 

It's a breaking change. There's no `Interpolated(value)` now to be used when defining a style. Use `Animated(value)` instead.

### DIFF
--- a/lib/js/src/components/flatList.js
+++ b/lib/js/src/components/flatList.js
@@ -33,7 +33,7 @@ function CreateComponent(Impl) {
     return /* () */0;
   };
   var scrollToOffset = function (ref, offset, animated, _) {
-    ref._scrollToIndex({
+    ref.scrollToOffset({
           offset: Js_undefined.fromOption(offset),
           animated: Js_undefined.fromOption(animated)
         });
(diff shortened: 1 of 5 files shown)


grabbou:
We should also check what values accept Animated and what not to protect against runtime errors.

grabbou:
It is a breaking change and removes `Interpolated` from styles in favor of using Animated everywhere.

wokalski:
@grabbou don't be discouraged by my use of `Request Changes`. It has this scary red look but I don't mean it 😄. I just try to communicate clearly.

grabbou:
Thanks for the review! You are totally right. Just updated the PR.

grabbou:
Yeah, I auto-generated this file one more time after changing types here and there. I would leave it as is unless there's a way to add white-spaces automatically. Otherwise, might be a bit tricky to maintain in the future.

PR #252, opened 2018-09-25
Easy to run example

grabbou:
Decided to replace the current example app with Expo one to lower the entry barrier for people to try out Reason and BsReactNative.


grabbou:
Going to work on few small styling issues before it's ready to merge.

example/package.json
```diff
@@ -0,0 +1,18 @@
+{
+  "name": "empty-project-template",
+  "main": "node_modules/expo/AppEntry.js",
+  "private": true,
+  "scripts": {
+    "start": "expo start",
+    "android": "expo start --android",
+    "ios": "expo start --ios",
+    "eject": "expo eject",
+    "build": "../node_modules/.bin/bsb -make-world -w",
```
ulrikstrid: Should this be a `watch` instead?

grabbou:
Right!

example/package.json
grabbou: You are right! :)

README.md
```diff
@@ -10,6 +10,8 @@ Currently supported React Native version: `^0.53.3`
 $ react-native init HelloWorld --template reason
 ```
 
+You can run the [playground](./example) with Expo: https://expo.io/@grabbou/bs-react-native
```
arnarthor: Does this still work on iOS? I vaguely recall Apple banning the ability to run apps that you don't own in Expo. 

If that's the case it should probably be documented that this only works on android

czystyl: Agree, I put information here

grabbou:
@arnarthor, decided to revert that change in one of my PRs because Expo website we link mentions that limitation explicitly. More over, it gives users choice to run on Android via QR code or run in-browser. Having such information in readme will prevent iOS users from even opening the link itself.

arnarthor:
Makes sense. :shipit:

PR #259, opened 2018-09-27
Fix links in docs

ostera:
Since https://github.com/ocaml/odoc/pull/204 got in `odoc`, I figured I'd have a look at what's breaking the links around here.

Turns out the autonamespacing is messing this up because it rewrites every module name under the hood. From `PermissionsAndroid` to `PermissionsAndroid-BsReactNative`.

I tried making the docs aware of this and reference modules by their namespaced name (such as `PermissionsAndroid-BsReactNative` instead of just `PermissionsAndroid`), but it's not actually possible since `PermissionsAndroid-BsReactNative` is not a valid OCaml module identifier and `odoc` is using standard OCaml parsing libraries to read the typed trees.

Manual namespacing is really not that big of a deal, and OCaml/Erlang/others have been doing it for 20+ years without much problem. Renaming a module to `RN_` like `RN_PermissionsAndroid` still lets you have a top-level `ReactNative` module that does the appropriate re-exports as you see fit.

Just my 2 cents on the issue 🙏 

cc/ @grabbou

### DIFF
--- a/bsconfig.json
+++ b/bsconfig.json
@@ -4,7 +4,7 @@
   "reason": {"react-jsx": 2},
   "refmt": 3,
   "bs-dependencies": ["reason-react"],
-  "namespace": true,
+  "namespace": false,
   "sources": [
     {
       "dir": "src",


wokalski:
Yes. I also think we can do manual name spacing if it saves us the headache. Another possibility is fixing this in bucklescript which might be as simple as replacing `*-PackageName` with `*_PackageName`.

grabbou:
I agree with that. Thank you for digging into this issue and finding out the root cause of the problem. Going to merge this PR which should deploy the website.

Once that's finished, I will send another PR to rename files (unfortunately, that can affect Github history) and will expose a single `BsReactNative` entry point.

PR #262, opened 2018-10-01
Provide API reference

### LINKED ISSUE #14: API Reference
The APIs are getting more and more different from the JS version. It'd be great to have some format of documentation. What are the best options? `ocamldoc`?

cc @chenglou @jordwalke

grabbou:
Closes #14

CC: @wokalski

### DIFF
--- a/docs/pages/page-status.mld
+++ b/docs/pages/page-status.mld
@@ -4,83 +4,83 @@ Please find the list of available components and APIs below.
 
 {3 Components}
 - {!ART}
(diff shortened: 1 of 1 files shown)


grabbou:
Not much to see now as we don't build documentation on PRs. Anyway, the links work and the ones w/o are APIs and components we are yet to support.

wokalski:
All pages are 404.

grabbou:
Ah, snap. Base-url is not customizable. Let me fix that.

PR #266, opened 2018-10-03
Documentation for ActivityIndicator

grabbou:
I am working on migrating documentation of ActivityIndicator from Rebolt to demonstrate the way we should handle this with `odoc`.

### DIFF
--- a/docs/highlight.pack.js
+++ b/docs/highlight.pack.js
@@ -16,4 +16,11 @@ document.addEventListener("DOMContentLoaded", function(event) {
       node.textContent = node.textContent.replace("-BsReactNative", "");
     });
   }
+
+  var make = document.querySelector("dl > dt#val-make");
+  if (make) {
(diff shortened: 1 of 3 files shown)


grabbou:
This is ready to be reviewed. I have updated the documentation for ActivityIndicator as well as added documentation for a View component.

That way, we don't have to repeat all the "common" props for every component (like we did in Rebolt), but link to the View component itself.

I decided to hide `let make` signature every time it has a corresponding documentation block. I am hoping that one day, `odoc` will do a better job of formatting that (there's a PR open).

On that note, is there a better way to put documentation / description of every parameter of the function instead of making description manually, like we did for ActivityIndicator? @ostera

grabbou:
screencapture-localhost-8080-bsreactnative-activityindicator-bsreactnative-2018-10-05-16_08_51

PR #269, opened 2018-10-08
Add documentation server

czystyl:
I added the HTTP server to host documentation locally, also added small steps what's need to be done to work with documentation locally.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -55,9 +55,9 @@ Documentation is generated from the Reason interface and comments that are inclu
 To start working with documentation:
 
 - Ensure that have the odoc installed [odoc](https://github.com/ocaml/odoc)
-  > if you have the problem during installing the odoc with OCaml dependencies, try: `opam pin add odoc.dev git+https://github.com/ocaml/odoc`
+  > if you have a problem during installing the odoc with OCaml dependencies, try: `opam pin add odoc.dev git+https://github.com/ocaml/odoc`
 - build code with: `yarn clean-build` or if you want to build with watch then use: `yarn start`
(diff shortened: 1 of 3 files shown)


CONTRIBUTING.md
```diff
@@ -48,6 +48,18 @@ Next, start Expo and follow on-screen instructions:
...
 ```
 
+## Running docs
+
+Documentation is generated from the Reason interface and comments that are included there.
+
+To start working with documentation:
+
+- Ensure that have the odoc installed [odoc](https://github.com/ocaml/odoc)
+  > if you have the problem during installing the odoc with OCaml dependencies, try: `opam pin add odoc.dev git+https://github.com/ocaml/odoc`
```
grabbou: *a problem

grabbou: Do we need `>`? :D

CONTRIBUTING.md
```diff
@@ -48,6 +48,18 @@ Next, start Expo and follow on-screen instructions:
...
+- build code with: `yarn clean-build` or if you want to build with watch then use: `yarn start`
+- start docs server using: `yarn start-docs`. Script will run the HTTP server and reload after something will change.
```
grabbou: something changes

package.json
```diff
@@ -23,6 +24,7 @@
   "devDependencies": {
     "all-contributors-cli": "^5.4.0",
     "bs-platform": "^4.0.0",
+    "live-server": "^1.2.0",
```
grabbou: Nice! I've always used `http-server`, but this one is nice as it has the live-reload capabilities.

CONTRIBUTING.md
czystyl: It looks better with `>` (imho) :P

Issue #267, opened 2018-10-08
Docs migration from rebolt to bs-react-native 


vnovick:
There was an extensive work done on [rebolt](https://rebolt.callstack.com/docs/intro.html) project for documenting lots of react-native components and APIs.  It will be great to migrate those to bs-react-native by using odoc auto generation with macros

Check the following example how docs should look like:
[Activity Indicator](https://reasonml-community.github.io/bs-react-native/BsReactNative/ActivityIndicator-BsReactNative/)
And this was the corresponding [PR](https://github.com/reasonml-community/bs-react-native/pull/266) by @grabbou 

Currently most of the docs are not migrated and formatted.

I will start opening PRs on docs migration and comment on this issue with PR links.
If you have additional suggestions how to improve docs beyond simple migration from Rebolt, please comment here too.

czystyl:
I'm going to move `Button`

MoOx:
@vnovick has everything being migrated? Can this issue be closed?

MoOx:
Looks like #294 closed this

PR #271, opened 2018-10-09
Add date picker doc

czystyl:
127 0 0 1_8080_bsreactnative_datepickerios-bsreactnative_

#267


grabbou:
CC: @vnovick this is a good example of a component documentation with small hacks. Might be worth checking how it compares with your work.

vnovick:
@grabbou  I don't see much of hacks here. Mostly just documenting above make function will get everything above make function. So there was no need to document types or modules. There when real problems arise. Especially when documenting modules as odoc generates modules as collapsible sections at the top of the file

PR #272, opened 2018-10-09
Migrating FlatList docs from rebolt

vnovick:
Migrated FlatList from rebolt. 

Due to the fact odoc puts comments below definitions there are custom css hacks which I don't really like 😕 So suggest other ways of putting docs section above module.

Also CreateComponent and FlatListComponent modules are hidden from the doc for that same reason.

screencapture-127-0-0-1-8080-bsreactnative-flatlist-bsreactnative-2018-10-09-23_50_01

#267

### DIFF
--- a/docs/css/custom.css
+++ b/docs/css/custom.css
@@ -4,6 +4,15 @@ dt#type-calculated {
   display: none;
 }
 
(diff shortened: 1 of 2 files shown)


grabbou:
Hopefully this will get less hacky once the JSON output is implemented and we come up with our own template. 

Great work!

Issue #85, opened 2017-10-26
Treat warnings as errors on CircleCI


wokalski:
We need to set `-warn-error` in the `bucklescript` flags. Additionally we might consider adding `@A` which turns on all warnings and makes them errors.

grabbou:
Updated description to better reflect current CI. Good first task.

MoOx:
I guess #277 closed this, so I am closing this issue. Feel free to reopen if I did something wrong :)

PR #277, opened 2018-10-18
adding warning as error flag to circleCI

vnovick:
So I added treat all warnings as errors to Circle CI  (issue #85) (In fact for every build) but build breaks because tons of errors.

Errors that were encountered:
- disambiguation errors for all panResponder handlers
```
Warning number 42
  /Users/vladimirnovick/dev/CallStackOpenSource/bs-react-native/src/panResponder.re 143:5-33

  141 ┆ onResponderTerminate:
  142 ┆   Js.Undefined.toOption(jsHandlers##onResponderTerminate),
  143 ┆ onResponderTerminationRequest:
  144 ┆   Js.Undefined.toOption(jsHandlers##onResponderTerminationRequest),
  145 ┆ onStartShouldSetResponder:

  this use of onResponderTerminationRequest required disambiguation.
```

- disambiguation errors for `netInfo.re` `connectionType` and `effectiveConnectionType` variants. Can be solved by wrapping them in separate modules
- disambiguation errors for `cameraRoll.re` for `groupType`
- polymorphic comparison error
```
  Warning number 102
  /Users/vladimirnovick/dev/CallStackOpenSource/bs-react-native/src/timePickerAndroid.re 50:34-47

  48 ┆ | Some(v) =>
  49 ┆   switch (v) {
  50 ┆   | _good when v <= valueTo && v >= valueFrom => Some(v)
  51 ┆   | _ =>
  52 ┆     failwith(

  polymorphic comparison introduced (maybe unsafe)
```
- platform.re pattern matching fragile
```
Warning number 4
  /Users/vladimirnovick/dev/CallStackOpenSource/bs-react-native/src/platform.re 36:3-41:3

  34 │
  35 │ let equals = targetOs =>
  36 │   switch (os(), targetOs) {
  37 │   | (IOS(_), IOS(_)) => true
   . │ ...
  40 │   | _ => false
  41 │   };
  42 │
  43 │ [@bs.module "react-native"] [@bs.scope "Platform"]

  this pattern-matching is fragile.
It will remain exhaustive when constructors are added to type os.
```

What are your ideas what is better to do with panResponders and polymorphic comparison errors
Other disambiguation errors can be solved with wrapping in modules

cc @grabbou @czystyl @wokalski


grabbou:
I think we could change the whole Platform module to have less conditionals. 

As far as the others, I guess we should just update/change the interface to eliminate them. I am not sure I understand where's the pan responder issue coming from. Mind explaining a bit?

vnovick:
This is the whole list of problems if we switch warnings to be errors. 

screen shot 2018-10-22 at 13 46 15

Most of the problems are disambiguation error as you can see. @grabbou, we need to think of best practice how we deal with these warnings. It's possible to use submodules al over the place, but we need to define best practices for this I guess.

As for pan responder I am also not sure what is the problem there. So trying to figure this out.

vnovick:
This is the whole list of problems if we switch warnings to be errors. 

screen shot 2018-10-22 at 13 46 15

Most of the problems are disambiguation error as you can see. @grabbou, we need to think of best practice how we deal with these warnings. It's possible to use submodules al over the place, but we need to define best practices for this I guess.

As for pan responder I am also not sure what is the problem there. So trying to figure this out.

vnovick:
This is the whole list of problems if we switch warnings to be errors. 

screen shot 2018-10-22 at 13 46 15

Most of the problems are disambiguation error as you can see. @grabbou, we need to think of best practice how we deal with these warnings. It's possible to use submodules al over the place, but we need to define best practices for this I guess.

As for pan responder I am also not sure what is the problem there. So trying to figure this out.

vnovick:
This is the whole list of problems if we switch warnings to be errors. 

screen shot 2018-10-22 at 13 46 15

Most of the problems are disambiguation error as you can see. @grabbou, we need to think of best practice how we deal with these warnings. It's possible to use submodules al over the place, but we need to define best practices for this I guess.

As for pan responder I am also not sure what is the problem there. So trying to figure this out.

wokalski:
IMO we should feel free to ignore some of those warnings.

vnovick:
@wokalski ok I ignored some warnings. Some of them are ignored by default btw (41,42,102) cause default ignore pattern is +a-4-6-7-9-27-29-32..39-41..42-44-45-48-50-102
Please review PR again and if everything is fine approve it and it will close issue #85

vnovick:
@wokalski, with 40 we are getting this error:
```bash
Warning number 40
  /Users/vladimirnovick/dev/CallStackOpenSource/bs-react-native/src/private/responderUtils.re 15:39-26:1

  13 │       (),
  14 │     )
  15 │     : Types.touchResponderHandlers => {
  16 │   onMoveShouldSetResponder,
   . │ ...
  25 │   onStartShouldSetResponderCapture,
  26 │ };

  this record of type BsReactNative.Types.touchResponderHandlers contains fields that are
not visible in the current scope: onMoveShouldSetResponder onMoveShouldSetResponderCapture onResponderGrant onResponderMove onResponderReject onResponderRelease onResponderTerminate onResponderTerminationRequest onStartShouldSetResponder onStartShouldSetResponderCapture.
They will not be selected if the type becomes unknown.

```
I've asked @grabbou and yours advise on this one. Do you know when it comes from?

wokalski:
You need to open `Types.{ }` on line 15.

vnovick:
@wokalski Thanks. I've updated PR. please review

MoOx:
I authorised myself to delete the branch associated with this PR since it has been merged. Hope that ok.

Issue #279, opened 2018-10-24
imageEditor API - PR


codinglawyer:
Hi, I created a PR with Reason bindings for imageEditor API :)

MoOx:
Closed by #368

PR #280, opened 2018-10-24
Image editor API

codinglawyer:
#279


vnovick:
LGTM, @czystyl, @grabbou, What do you think?

MoOx:
@codinglawyer just add the doc to the .rei and it will be perfect!
You can see example on how to document here https://github.com/reasonml-community/bs-react-native/blob/e9ae1ea77957f90bd75269305b008aff21a0ece5/src/actionSheetIOS.rei
More about the doc format here http://caml.inria.fr/pub/docs/manual-ocaml/ocamldoc.html#sec357

sgny:
@codinglawyer has contributed solid bindings and I'm already using them (so thanks are in order), but there is room for a few refinements. I would prefer that only the types `imageSource`, `offset`, `size`, `resizeModeT` and `error` should be exposed in the interface file, in addition to the `cropImage` function. Since `error` objects may only be returned and handled with the callback, the creation function for `error` could be disabled with the `pri` keyword.

I have already implemented these changes and prepared documentation and I could create a PR. While @codinglawyer may not finish off his work as he has not returned to submit documentation in the past two months, he has already taken care of most of it, so I would like to hold off until his work gets merged.

ulrikstrid:
That sounds really good @sgny, are you okay with that @MoOx?

MoOx:
@ulrikstrid Indeed. I am currently trying to get things back to normal with the CI (since almost all PRs aren't building)...
@sgny If I understand correctly, the right move here would be to merge this in its current state & poke you to get your work?

sgny:
@MoOx Yes that seems fair to me.

src/imageEditor.re
```diff
@@ -0,0 +1,100 @@
...
+type size = {
+  width: int,
+  height: int,
+};
+
+[@bs.deriving jsConverter]
+type resizeModeT = [
+  | [@bs.as "contain"] `Contain
+  | [@bs.as "cover"] `Cover
+  | [@bs.as "stretch"] `Stretch
```
MoOx: All those bs.as are unnecessary

```
Warning number 101
  /home/opam/bs-react-native/src/imageEditor.rei 15:7-11
  
  13 │ [@bs.deriving jsConverter]
  14 │ type resizeModeT = [
  15 │   | [@bs.as "contain"] `Contain
  16 │   | [@bs.as "cover"] `Cover
  17 │   | [@bs.as "stretch"] `Stretch
  
  Unused BuckleScript attribute: bs.as

  Warning number 101
  /home/opam/bs-react-native/src/imageEditor.rei 16:7-11
  
  14 │ type resizeModeT = [
  15 │   | [@bs.as "contain"] `Contain
  16 │   | [@bs.as "cover"] `Cover
  17 │   | [@bs.as "stretch"] `Stretch
  18 │ ];
  
  Unused BuckleScript attribute: bs.as

  Warning number 101
  /home/opam/bs-react-native/src/imageEditor.rei 17:7-11
  
  15 │   | [@bs.as "contain"] `Contain
  16 │   | [@bs.as "cover"] `Cover
  17 │   | [@bs.as "stretch"] `Stretch
  18 │ ];
  19 │ 
  
  Unused BuckleScript attribute: bs.as
```

MoOx: I guess bs automatically transform those according to the name of the variant?!

sgny:
Without `bs.as`, the first letter will be uppercase which will cause an error with RN.

In the interface, however, there is no need to expose the `jsConverter` function, so only the variants should suffice.

src/imageEditor.re
sgny: Without `bs.as`, the first letter will be uppercase which will cause an error with RN.

In the interface, however, there is no need to expose the `jsConverter` function, so only the variants should suffice (`imageEditor.rei` copied below, excluding documentation):

```
type imageSource = [ | `URI(string) | `Required(Packager.required)];

type offset = {
  x: int,
  y: int,
};

type size = {
  width: int,
  height: int,
};

type resizeModeT = [ | `Contain | `Cover | `Stretch ];

[@bs.deriving abstract]
type error = pri {
  code: int,
  message: string,
};

let cropImage:
  (
    ~uri: imageSource,
    ~offset: offset,
    ~size: size,
    ~displaySize: size=?,
    ~resizeMode: resizeModeT=?,
    ~success: string => unit,
    ~failure: error => unit,
    unit
  ) =>
  unit;
```

MoOx: Oh right, I didn't read correctly the filename in the error. My bad.

src/imageEditor.rei
```diff
@@ -0,0 +1,73 @@
...
+  width: int,
+  height: int,
+};
+
+[@bs.deriving jsConverter]
+type resizeModeT = [
+  | [@bs.as "contain"] `Contain
+  | [@bs.as "cover"] `Cover
+  | [@bs.as "stretch"] `Stretch
+];
```
MoOx: Needs a change, see 
https://github.com/reasonml-community/bs-react-native/pull/280/files#r256710619

sgny: Could you clarify please? Changes in line with what I've proposed (removing `bs.as` annotations, etc.) or anything beyond that?

MoOx: I guess anything that will makes compiler happy? :)
I am still new to this bs.* directives

sgny: Alright then, I was confused because with my changes in place, I get no errors or warnings from the compiler and the module seems to work as expected in the app.

MoOx:
We are going to revisit the bindings in the next rewrite, so we won't merge this as it is. Thanks anyway for the work, we will surely reuse it somehow!

PR #281, opened 2018-10-25
Fix share action sheet options converting

sck-v:
There was error converting options for share action sheet between Reason's variant type and object, containing `url` or `message` fields required for native function


src/actionSheetIOS.re
```diff
@@ -1,6 +1,8 @@
 type actionSheetConfig;
 
 type shareActionSheetConfig;
+external shareActionSheetConfig: Js.t('a) => shareActionSheetConfig =
+  "%identity";
```
czystyl: `%` is added intentionally?

MoOx: Probably, see 
https://bucklescript.github.io/docs/en/intro-to-external#special-identity-external

MoOx:
@sck-v can you confirm that the change is not a breaking change for the bs binding?

src/actionSheetIOS.re
czystyl: Nice, I didn't know about it. Thanks for the explanation ;)

sck-v:
> @sck-v can you confirm that the change is not a breaking change for the bs binding?

Yep, it just fixes internal implementation

PR #283, opened 2018-10-26
Exposed ReactNative's Platform.select function.

SandyChapman:


MoOx:
Interested in having this!
Could you rebase/resolve conflicts please?

SandyChapman:
@MoOx : Rebased. Please review.

MoOx:
We need to figure out why ci doesn't want to test this pr... (like lot's of others)

src/platform.re
```diff
@@ -51,3 +51,25 @@ let version = () =>
...
   };
+
+type selectOptions('a) = {
+  ios: 'a,
+  android: 'a,
+};
+
+[@bs.deriving abstract]
+type _selectOptions('a) = {
+  [@bs.as "ios"]
```
czystyl: I think that we don't need to annotate fields if we don't want to change the fields name

SandyChapman: Good point. Will remove the annotations.

SandyChapman: @czystyl : Updated. Please re-review.

MoOx:
I am having second thoughts about this binding. Please take a look to 
https://github.com/reasonml-community/bs-react-native/issues/316

PR #284, opened 2018-10-26
add trackColor prop to Switch

sgny:
`onTintColor` is deprecated starting with RN 0.57, I have added the replacement property.

The property takes as argument an object of type `{false: color, true: color}` where `color` is either a `string` or `null`. With the function defined in this commit, `Js.Dict.t(Js.nullable(string))` is passed  to handle the `null` case.

To use from Reason, an `array((bool, string))` argument is required.

```
<Switch
  trackColor=[|(true, "green"), (false, "red")|]
  ... 
  value=false
/>
```

Providing only one tuple is properly handled, that is an argument `[|(true, "green")|]` will yield the appropriate object `{false: null, true: "green"}`.

For use with earlier versions of RN, `onTintColor` remains available.


sck-v:
As for me the declarative interface will look better:

```
<Switch
  trackColor=Switch.trackColor(~true="green", ~false="red", ())
  value=false
/>
```

Partial option will work nice tooo:

```
<Switch
  trackColor=Switch.trackColor(~true="green", ())
  value=false
/>
```

sgny:
@sck-v Thank you for the feedback. I did have some back-and-forth regarding interface design.

There is one disadvantage to your suggestion, however. `true` and `false` are reserved and cannot be used as names for function parameters. I did consider using `true_` and `false_` (along with `[@bs.as "true"]` and `[@bs.as "false"]`) and using the `bs.deriving` construct, but passing the argument as `array(tuple)` or `list(tuple)` allowed keeping `true` and `false`. Once I had the array, I found it appealing to use standard library functions to create a `Js.Dict.t('a)` object from that.

I will create another PR based on the general sentiment, I suppose we should wait to see if there is more feedback.

sck-v:
> however. true and false are reserved and cannot be used as names for function parameters

Yeah, sure. I gave that example just for reference

src/components/switch.re
```diff
@@ -1,8 +1,18 @@
 [@bs.module "react-native"] external view: ReasonReact.reactClass = "Switch";
 
+let makeTrackColor = (trackColor: option(array((bool, string)))) => {
+  let arr = switch (trackColor) {
+  | None => [||]
+  | Some(x) => x
+  };
+  let f = (s: option(string)): Js.nullable(string) => {if (s == None) Js.Nullable.null else Js.Nullable.fromOption(s)};
+  Js.Dict.fromArray(Array.map(y => (Js_string.make(y), f(UtilsRN.option_map(snd, Js.Array.find((x) => {fst(x) == y}, arr)))),  [|true, false|]))
```
MoOx: Could you run refmt on this to make it more easy to read? Thanks!

sgny: I ran refmt, but also made a few trivial changes.

• replaced `UtilsRN.option_map` with `Belt.Option.map`. As the former is simply another implementation of the latter with reversed order of arguments, perhaps it should be replaced throughout the repo. However, I can undo that change if it is preferred to avoid using `Belt`.
• added the prop to documentation

I also force pushed the commit to reduce the number of commits based on feedback from my first accepted PR into this repo.

sgny:
I made further use of the `Belt.Option` module to remove redundant implementations of `getWithDefault` and `mapWithDefault`.

MoOx:
A problem I see with the array solution is that it can leads errors. For example if you make an array with multiple `"true"` (or false)... And that's not safe imo.

So maybe indeed we should go with something more declarative like @sck-v suggested, even if we need to use true_ & false_... Or trueColor & falseColor.

sgny:
The resultant object would not be affected because Js.Array.find will only match one tuple with true or false as the first element. Of course this does not save users from themselves, so another design might be preferred on that account.

Perhaps two props, one each for true and false might be even cleaner.

MoOx:
> Perhaps two props, one each for `true` and `false` might be even cleaner.

Why not? `trackTrueColor` & `trackFalseColor`?

sgny:
> > Perhaps two props, one each for `true` and `false` might be even cleaner.
> 
> Why not? `trackTrueColor` & `trackFalseColor`?

I avoided that to keep only one prop akin to the JS side, but it is definitely far more straightforward.
Only, given the prop is `trackColor` in JS, perhaps they should be named `trackColorTrue` and `trackColorFalse`. I'll force push a revised version in a short while.

sgny:
I revised the PR to split `trackColor` into two props and implemented the `ios_backgroundColor` and `thumbColor` props.

I have also updated documentation accordingly.

src/components/switch.re
```diff
@@ -3,9 +3,13 @@
 let make =
     (
       ~disabled: option(bool)=?,
+      ~trackColorTrue: option(string)=?,
+      ~trackColorFalse: option(string)=?,
+      ~ios_backgroundColor: option(string)=?,
```
MoOx: I know that this is the RN name, but what about `iosBackgroundColor` ? The mix of _ & uppercase is a bit awkward to me.

sgny: You're right, looks quite ugly. I've revised accordingly.

MoOx:
@sgny nice!
Could you run

- `yarn build` (& commit changes to js)
- `yarn bsrefmt --in-place` on each .re files you changed (on also you could thumbs up this issue #312)

Thanks!

PR #287, opened 2018-10-29
Add command to easily run refmt in place on all files

MoOx:
Convenient, don't you think?
Maybe there is a quicker way that I am not aware of?

After this PR, I could do another one to actually run this command (minor changes expected, just some newlines at EOF and 2 or 3 files actually with diff). Maybe we could even run this as a precommit hook?

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,6 +2,7 @@
   "name": "bs-react-native",
   "version": "0.10.0",
   "scripts": {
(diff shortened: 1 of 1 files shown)


ulrikstrid:
As this PR doesn't break anything I will go ahead and merge it. I think this should be run every time we update bs-platform version.

Issue #273, opened 2018-10-11
Reason creates an object with keys that have value undefined, crashing RN's Animated


adamjuhasz:
I'm very new to Reason so I'm not sure if this needs to go the the core Reason project or this is the best place for it...

When using `ValueXY.Spring.animate()`, the parameters that are not passed in get sent to RN's Animated framework as `undefined` values in the config object. When the framework attempts to destructure the undefined value, an exception is thrown. The react native project would rather not fix this with an undefined check (https://github.com/facebook/react-native/pull/21717), where would it best be to fix this on the Reason side? I assume the keys with `undefined` values should be removed from the object before being passed along?

```Reason
let commonAnimatedSpring = ValueXY.Spring.animate(
  ~value=state.coord,
  ~velocity={"x": 0., "y": 1.},
  ~toValue=`raw({"x": 0., "y": 0.})
);
commonAnimatedSpring()->Animation.start();
```
this creates a config object:
```javascript
const config = {
  bounciness: undefined,
  damping: undefined,
  velocity: {x: 0, y: 1},
//...
}
```
this fails with an exception: `TypeError: TypeError: Cannot read property 'x' of undefined` because of the following code `const {x, y} = config[key];` at https://github.com/facebook/react-native/blob/6eeff75849c9b8bf91592c1b7906b4dab8fba518/Libraries/Animated/src/AnimatedImplementation.js#L113

Example project: https://github.com/paytgthr/reason-drawer/blob/RNAPP-5/src/GestureHandler.re

sgny:
I had also ran into this issue myself, but rolled back to a previous commit because I could not look into it at the time.

I added your sample code to my project to test and reproduced the error. I believe the error was caused by incorrect type annotations, so I created PR #288, which does fix the error for me. 

If you have not resolved the issue yet, please take a look into the PR.

adamjuhasz:
@sgny Thank you!

PR #288, opened 2018-11-03
Correct type annotations for arguments in AnimatedRe.ValueAnimations.Spring module

sgny:
Type annotations for the `Spring` module lead to creation of a config object where keys for optional props have `undefined` as their values. This causes runtime `Type error` exceptions with React Native. 

Redefining type annotations in line with other AnimatedRe modules (`Timing`, `Decay`, `Interpolation`) resolves this issue.

Issue was also reported in #273.


MoOx:
@sgny I understand (as I encountered similar issues with others modules) that this is purely a bugfix (no breaking change in anyway?) Thanks for the confirmation!

sgny:
@MoOx That's right, this is purely a bugfix.

PR #289, opened 2018-11-06
Fix Animated.delay example in docs

MoOx:
Just because it's wrong on 2 parts :)

### DIFF
--- a/src/animatedRe.rei
+++ b/src/animatedRe.rei
@@ -553,15 +553,14 @@ let event: (array('a), 'b) => animatedEvent;
 let animation =
   Animated.sequence(
     [|
-      Animated.delay(500),
+      Animated.delay(500.),
       Animated.timing(
         ~value=barValue,
         ~toValue=`raw(0.0),
         ~duration=100.0,
         (),
       ),
-    |],
-    {"stopTogether": false},
+    |]
   );
 
 Animated.start(animation, ());]}


src/animatedRe.rei
```diff
@@ -740,4 +739,4 @@ let stop: Animation.t => unit;
 let start: (Animation.t, ~callback: Animation.endCallback=?, unit) => unit;
 
 /** See {!Animation.reset} for details. */
-let reset: Animation.t => unit;
\ No newline at end of file
+let reset: Animation.t => unit;
```
MoOx: This change is because I used GitHub UI for the fix :/

src/animatedRe.rei
```diff
@@ -553,15 +553,14 @@ let event: (array('a), 'b) => animatedEvent;
 let animation =
   Animated.sequence(
     [|
-      Animated.delay(500),
+      Animated.delay(500.),
```
MoOx: Error 1.

src/animatedRe.rei
```diff
@@ -553,15 +553,14 @@ let event: (array('a), 'b) => animatedEvent;
...
       Animated.timing(
         ~value=barValue,
         ~toValue=`raw(0.0),
         ~duration=100.0,
         (),
       ),
-    |],
-    {"stopTogether": false},
```
MoOx: Error 2: stopTogether is for parallel :)

src/animatedRe.rei
ulrikstrid: I think this is correct anyway

Issue #240, opened 2018-09-17
Use new @bs.deriving abstract 


czystyl:
Some of the components and API's uses the old object API that is deprecated by [BuckedScript doc](https://bucklescript.github.io/docs/en/object-deprecated). 
Instead of using deprecated object use the new [`[@bs.deriving abstract]`](https://bucklescript.github.io/docs/en/object) and [`[@bs.deriving jsConverter]`](https://bucklescript.github.io/docs/en/generate-converters-accessors)

List of component contain depricated API:
* [ ] flatList
* [ ] animatedRe
* [ ] cameraRoll
* [ ] scrollView
* [ ] sectionList
* [ ] textInput
* [ ] touchableHighlight
* [ ] touchableOpacity
* [ ] viewPagerAndroid
* [ ] imagePickerIOS
* [ ] linking
* [ ] panResponder
* [ ] utilsRN
* [ ] RNEvent
* [ ] events
* [ ] permissionsAndroid

Feel free to update that list.

czystyl:
The issue is moved from Rebolt repository: https://github.com/callstackincubator/rebolt/issues/47

wokalski:
Do you think it's worth it? I'm not convinced. `Js.t` API is not deprecated anymore. It's been undeprecated. What's the reasoning here? Maybe we should go about it case by case?

czystyl:
We created that list when `Js.t` was deprecated, but I haven't realized that it's undeprecated now. As you mentioned that not all component from the list need changes but in some cases moving to `@bs.deriving` could improve usage of API. I'd keep this list of a possibility of improving and create separate issue/PR for a single component/API.

MoOx:
Does this issue make sense to be opened? Could we open an issue per compo if still relevant?

MoOx:
No need for this anymore, we are going to follow another approach with new rewrite

PR #292, opened 2018-11-12
use bs.deriving abstract in extendView

### LINKED ISSUE #291: KeyboardAvoidingView does not work
When adding a KeyboardAvoidingView it does nothing because the onLayout that sets the frame-ref is never called.

ChristophGr:
This fixes  #291
and is related to
#240


src/private/props.re
```diff
@@ -138,9 +173,11 @@ let extendView =
...
-      "shouldRasterizeIOS": shouldRasterizeIOS,
-    },
-    moreProps,
-    serialize(responderHandlers),
-  );
+      ~accessibilityViewIsModal?,
+      ~shouldRasterizeIOS?,
+      (),
+    )
+    ->Obj.magic;
```
MoOx: Are you sure about this `Obj.magic`? I guess a comment would make sense on top of it to avoid loosing the rational behind it.

MoOx:
No need for this anymore, we are going to follow another approach with new rewrite

PR #303, opened 2019-01-13
Moving common polymorphic variants to types.re

sgny:
In many components, permitted values for common props (`pointerEvents`, `accessibilityTraits`, etc.) have been defined as unbound, polymorphic variants. Since permitted values for those props are identical across RN components, this was a lot of repetition - which I felt could be removed. Further, functions to convert constructors to strings were identically and duplicately defined across various components and those duplicates could be removed.

I have defined relevant polymorphic variants in `src/private/types.re` and updated components to use those types, instead of adhoc definitions in each. I decided to take advantage of the`@bs.deriving jsConverter` annotation for necessary converters, but as that would not work within the interface-only `types` (`types.rei`), I changed that be a regular module (`types.re`).

`touchableOpacity` and other `touchable*` components might also be updated to make use of `Props.extendView`, but such further changes might be better left until #292 can be merged.

Note: I had originally defined the types in `props.re` but had to reconsider since that meant the constructors were outside the scope of other projects (oversight due to changing `private` directory locally to no longer be `private` in my local copy). `types.re` is arguably the more appropriate location for the types, anyway. However, if keeping `types` inferface-only is preferred, I'll update the PR accordingly.


sgny:
> If I interpret this correctly, I understand it's not a change for the end user.
> Looks good to me.

Right, no changes for end users.

MoOx:
Maybe we should keep a `src/types.rei` version in addition to the new `types.re` (which handle the bs.* I guess)?

sgny:
All the JS converters have to be exposed and the interface would be identical to the module, so I chose to omit it. If the types were moved to `props.re` and `touchable*` components were updated to make use of `Props.extendView` then exposing the converters would no longer be necessary and a more minimal interface may be defined.

Should I still add the interface or implement said changes in a new PR?

MoOx:
@sgny lets wait for another review or 2 before, I am still new to reasonml so not entirely confident with my feedback haha :)

czystyl:
I like it, let me go through all the codebase and merge :) 
image

@sgny awesome work 🚀

MoOx:
@sgny could you run `yarn build` & commits changes of `lib/js/src`?

sgny:
OK, I handled that. Since almost all source files were touched, I also ran refit on everything to take to fix remaining formatting inconsistencies.

MoOx:
@sgny sorry to bother you again, but it seems you aren't using `yarn` as required per our project, which means files where builded using bs-platform 4.0.18 instead of [4.0.5](https://github.com/reasonml-community/bs-react-native/blob/8127fe03359d70ee72d45b0eabd02bb578546e59/yarn.lock#L263). Could you use yarn so we have a small & relevant changes only please? Thanks!

sgny:
I realized afterwards that `package.json` is not on `.gitignore` so that got changed to require BS version 4.0.5. Is that alright or should I fix that?

MoOx:
@sgny I still see too many files changes :/

sgny:
@MoOx I’ll check again, but in the meanwhile if any file in particular has caught your attention, please do let me know.

MoOx:
@sgny Nevermind I wan't paying enough attention. A file rolled back a incorrect bs-platform version due to a previous incorrecte change.

PR #310, opened 2019-02-10
Fix odoc install on Circle-CI

MoOx:
`[I don't know what I am doing.gif]`

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -31,13 +31,12 @@ defaults: &defaults
 
 version: 2
 jobs:
-
   # Set up a Node environment for downstream jobs
   checkout_code:
     <<: *defaults
     steps:
       - checkout
-      
+
       # Install Node.js dependencies
       - restore-cache: *restore-yarn-cache
       - run: yarn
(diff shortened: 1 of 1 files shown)


MoOx:
Seems I fixed it. Any thoughts about how I did it? @wokalski @grabbou 
Note that I used another docker image (found this in facebook/reason repo) and it seems to be faster (at least to get the opam/odoc ready to go).

wokalski:
This seems great @MoOx. I'm giving an approval, but I'd like to hear what @ulrikstrid thinks since he's most knowledgable when it comes to odoc here.

ulrikstrid:
I only used odoc with dune and esy, but @ostera might be interested in this repo

MoOx:
I used this docker image following a discussion with him (dm on discord). Not that he advised this precise solution but he pointed out that the issue we were facing was because of the opam version (1.2.2 before this pr). We need opam 2.x apparently.
This docker image might be a good choice as it's fast & reliable.
I would like @grabbou feedback as it seems he setup the previous circle-ci config.

ulrikstrid:
I think this is good as it is. I hope we can get better integrating of odoc and bsb soon so we shouldn't spend more time than necessary on this imo.

MoOx:
Going to merge then. If we want to improve we can :)

PR #314, opened 2019-02-12
Add opam requirement (and the right version) to contributing notes

MoOx:
Because it's a missing information I struggled to find (I found this in CI configuration).


czystyl:
👏

PR #317, opened 2019-02-14
Bind stickyHeaderIndices

Arnarkari93:


lib/js/src/style.js
```diff
@@ -250,7 +250,7 @@ function justifyContent(v) {
     case 4 : 
         tmp = "space-between";
         break;
-    case 6 : 
+    case 5 :
```
MoOx: I think this change should not be here

MoOx: Oh I think I see, it's a missing update. Will handle this in master.

MoOx: I cannot commit directly into master, so I made #318

PR #318, opened 2019-02-14
Fixup js output for 9bc1d32 (because of 3081fac)

MoOx:
#307 made the js generated by #305 incorrect.
This PR fixes that.


MoOx:
@czystyl thanks :)

PR #324, opened 2019-02-20
Remove UtilsRN (option_map -> Belt.Option.map)

MoOx:
This PR remove `UtilsRN.option_map` and replace all usage by `Belt.Option.map`.
While I was on this change, I use `->` to send props to map function. Feel more natural imo. When it made sense, I also move `fromOption` at the end using `->` too.
Also, I remplaced som `x => switch(x) {` by `fun`.
I replaced the only usage of `objAssign2` by a simpler solution (2 `Js.Obj.assign`)


grsabreu:
nice clean up

MoOx:
My idea here is to minimize private code so I can override bs-react-native more easily to write bs-react-native-web

sgny:
Thank you for creating this PR, I've been thinking `option_map` should be replace with `Belt.Option.map` for a while now.

> My idea here is to minimize private code so I can override bs-react-native more easily to write bs-react-native-web

Have you already started on BS bindings for `react-native-web`? I see there was some work, but it's long been deprecated. I would like to test them out if you feel like making them available somewhere.

MoOx:
@sgny I have started a thing, but just with props I used (which is not a lot). Will communicate about it via a PR soon :)

PR #345, opened 2019-03-21
Add .bsb.lock to .gitignore.

cknitt:


.gitignore
```diff
@@ -7,6 +7,7 @@ node_modules
 package-lock.json
 
 # ocaml/reason build artifacts
+.bsb.lock
```
MoOx: shouldn't we add `**/`?

cknitt: It's sufficient like this, it matches .bsb.lock in subdirectories, too.

MoOx: I was thinking the same, but the other day, I had to make some changes for patterns below (cause it wasn't enough...). Weird, maybe I am buggy lol

PR #346, opened 2019-03-21
Move cca-bs-react-native to bs-react-native-next

cknitt:
Move https://github.com/cca-io/cca-bs-react-native to bs-react-native-next.


cknitt:
As discussed on Discord: What remains to be cleaned up/completed is mostly AnimatedRe (to be renamed to Animated) and Event (which I already renamed from RNEvent).

And some components/APIs are not bound yet at all.

MoOx:
I am accepting this in it's current state and we can iterate and discuss if needed with issues/PRs :)

Thanks for this work 🙏

PR #347, opened 2019-03-21
Zero cost bindings for style merge (array/list)

MoOx:
RN accept works out of the box array of styles (object|int from StyleSheet entries|false|null|undefined).
It also recursively resolve the array of styles, so list should work too (as list are just recursive array).

I tried to find a "smart" name but at the end, using merge/flatten/combine/concat just make this more confusing...

This PR
- deprecate all previous confusing names method
- add array/list free binding
- add also array/listOption as those are free to, & convenient when you conditionally mix styles(since RN accepts undefined & that None is encoded as undefined...)
- deprecate combine/merge as this is totally unsafe IRL (you cannot merge object of style with a stylesheet entry, which works with current types...)

Thoughts? Idea for better naming?

### DIFF
--- a/bs-react-native-next/src/Style.re
+++ b/bs-react-native-next/src/Style.re
@@ -252,11 +252,6 @@ external style:
   t =
   "";
 
-[@deprecated "Use Style.array([|t|]) instead"]
-[@bs.module "react-native"]
-[@bs.scope "StyleSheet"]
-external flatten: array(t) => t = "";
-
 external array: array(t) => t = "%identity";
 external arrayOption: array(option(t)) => t = "%identity";
 external list: list(t) => t = "%identity";
--- a/bs-react-native-next/src/Style.re
+++ b/bs-react-native-next/src/Style.re
@@ -252,6 +252,12 @@ external style:
   t =
   "";
 
-// Duplicated from StyleSheet.re for convenience.
-[@bs.module "react-native"] [@bs.scope "StyleSheet"]
+[@deprecated "Use Style.array([|t|]) instead"]
+[@bs.module "react-native"]
+[@bs.scope "StyleSheet"]
 external flatten: array(t) => t = "";
+
+external array: array(t) => t = "%identity";
+external arrayOption: array(option(t)) => t = "%identity";
+external list: list(t) => t = "%identity";
+external listOption: list(option(t)) => t = "%identity";
--- a/bs-react-native-next/src/Style.re
+++ b/bs-react-native-next/src/Style.re
@@ -252,7 +252,39 @@ external style:
   t =
   "";
 
+/*
+ <View style=array([|
+   styles##thing,
+   styles##whatever,
+ |])>
+   */
 external array: array(t) => t = "%identity";
+
+/*
+ <View style=arrayOption([|
+   Some(styles##thing),
+   Some(styles##whatever),
+   optionalStyle,
+   cond ? Some({something:"dynamic"}) : None
+ |])>
+ */
 external arrayOption: array(option(t)) => t = "%identity";
+
+/* list works too since RN accept recursive array of styles (list are just recursive arrays)*/
+/*
+ <View style=list([
+   styles##thing,
+   styles##whatever,
+ ])>
+   */
 external list: list(t) => t = "%identity";
+
+/*
+ <View style=listOption([
+   Some(styles##thing),
+   Some(styles##whatever),
+   optionalStyle,
+   cond ? Some({something:"dynamic"}) : None
+ ])>
+ */
 external listOption: list(option(t)) => t = "%identity";
--- a/bs-react-native/src/style.re
+++ b/bs-react-native/src/style.re
@@ -81,11 +81,16 @@ let objectStyle = (key, value) => (key, Internals.Encoder.object_(value));
 let arrayStyle = (key, value) => (key, Internals.Encoder.array(value));
 
 let style = sarr => sarr |> Js.Dict.fromList |> to_style;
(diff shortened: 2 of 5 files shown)


bs-react-native/src/style.re
```diff
@@ -81,27 +81,23 @@ let objectStyle = (key, value) => (key, Internals.Encoder.object_(value));
 let arrayStyle = (key, value) => (key, Internals.Encoder.array(value));
 
 let style = sarr => sarr |> Js.Dict.fromList |> to_style;
-let emptyStyle = Js.Dict.empty()->to_style;
 
-external fromArray: array(t) => t = "%identity";
-let fromList = styles => styles->Belt.List.toArray->arrayOfStyle;
```
MoOx: I am just removing here things I previously added, but never released.

bs-react-native/src/style.re
```diff
@@ -81,27 +81,23 @@ let objectStyle = (key, value) => (key, Internals.Encoder.object_(value));
...
 };
-let mergeOptional = (s, so) =>
-  so->Belt.Option.map(so => s->merge(so))->Belt.Option.getWithDefault(s);
-let optional = s => s->Belt.Option.getWithDefault(emptyStyle);
-
-/* deprecated */
-let flatten = fromArray;
-let concat = fromList;
-let combine = merge;
-let combineOptional = mergeOptional;
```
MoOx: Removed without notice as never released

bs-react-native/src/style.re
```diff
@@ -81,27 +81,23 @@ let objectStyle = (key, value) => (key, Internals.Encoder.object_(value));
...
   let entries =
     Array.append(
       Js.Dict.entries(style_to_dict(a)),
       Js.Dict.entries(style_to_dict(b)),
     );
   Js.Dict.fromArray(entries) |> to_style;
 };
-let mergeOptional = (s, so) =>
-  so->Belt.Option.map(so => s->merge(so))->Belt.Option.getWithDefault(s);
-let optional = s => s->Belt.Option.getWithDefault(emptyStyle);
```
MoOx: Removed without notice as never released

bs-react-native-next/src/Style.re
```diff
@@ -252,6 +252,12 @@ external style:
   t =
   "";
 
-// Duplicated from StyleSheet.re for convenience.
-[@bs.module "react-native"] [@bs.scope "StyleSheet"]
+[@deprecated "Use Style.array([|t|]) instead"]
+[@bs.module "react-native"]
+[@bs.scope "StyleSheet"]
 external flatten: array(t) => t = "";
```
cknitt: I think we can just remove this instead of deprecating it as -next is still WIP/unreleased anyway.

bs-react-native-next/src/Style.re
```diff
@@ -252,6 +252,12 @@ external style:
...
+
+external array: array(t) => t = "%identity";
+external arrayOption: array(option(t)) => t = "%identity";
+external list: list(t) => t = "%identity";
+external listOption: list(option(t)) => t = "%identity";
```
cknitt: It's great that this works out of the box. I wonder whether there is any performance difference to using StyleSheet.flatten.

bs-react-native-next/src/Style.re
MoOx: Done.

bs-react-native-next/src/Style.re
MoOx: StyleSheet.flatten (per doc) seems to have a cost since it's resolve stylesheets references (int) to the styles associated etc (see 
https://facebook.github.io/react-native/docs/stylesheet#flatten).

So passing styles from array should be better as ids are directly passed to the bridge, with eventual dynamic styles.

bs-react-native/src/style.rei
```diff
@@ -25,20 +25,48 @@ type deg_animated('a) =
...
 let style: list(styleElement) => t;
-external fromArray: array(t) => t = "%identity";
-let fromList: list(t) => t;
-let merge: (t, t) => t;
-let mergeOptional: (t, option(t)) => t;
-let optional: option(t) => t;
 
-[@deprecated "use Style.fromArray([|t|]) instead"]
+/*
+ <View style=mix([|
```
cknitt: Should be `array`, not `mix`.

bs-react-native/src/style.rei
```diff
@@ -25,20 +25,48 @@ type deg_animated('a) =
...
+   styles##whatever,
+ |])>
+   */
+let array: array(t) => t;
+/*
+ <View style=arrayOption([|
+   Some(styles##thing),
+   Some(styles##whatever),
+   optionalStyle,
+   cond ? {something:"dynamic"} : None
```
cknitt: `{something: "dynamic"}` is not a valid `option(Style.t)`. Maybe better write

```cond ? Some(conditionalStyle) : None```

File: bs-react-native/src/style.rei [hunk 8]
```diff
@@ -25,20 +25,48 @@ type deg_animated('a) =
...
+   styles##whatever,
+ ])>
+   */
+let list: list(t) => t;
+/*
+ <View style=listOption([
+   Some(styles##thing),
+   Some(styles##whatever),
+   optionalStyle,
+   cond ? {something:"dynamic"} : None
```
cknitt: `{something: "dynamic"}` is not a valid `option(Style.t)`. Maybe better write

```cond ? Some(conditionalStyle) : None```

MoOx:
All fixed @cknitt,  thanks for your review. I also duplicated fixed comments in next version. Would help for upcoming docs :)

PR #350, opened 2019-03-23
Fix accessibilityLabel type.

cknitt:


MoOx:
@cknitt can you explain the rational behind this change?
I noticed RN docs mention type "node" in most place (except button, why? no idea).
But in practise, it seems codebase has this typed as string, see https://github.com/facebook/react-native/search?p=2&q=accessibilitylabel&unscoped_q=accessibilitylabel
You can notice that ios & android implementation explicitly use string (same for the (react-native-)web).

The only place we see node in the codebase, is in the deprecated props, see https://github.com/facebook/react-native/blob/408207b356849ca3cefad0ce91a3889e73bd8682/Libraries/DeprecatedPropTypes/DeprecatedImagePropType.js#L34 so I am not sure this make change to change the bindings...

cknitt:
I saw that the docs said `node` and assumed that that was correct. 😕

Thanks for checking in the source, let's leave `string`.

MoOx:
FYI I tried to fix that on RN website https://github.com/facebook/react-native-website/pull/850

PR #356, opened 2019-04-04
bs-react-native-next: Add Image methods

fhammerschmidt:
Also fixed the cache enums with [@bs.as].

Concerning `resolveAssetSource`: You can look up in the [RN code](https://github.com/facebook/react-native/blob/1151c096dab17e5d9a6ac05b61aacecd4305f3db/Libraries/Image/resolveAssetSource.js#L88), that this method does not do anything for a source that isn't a number (or `Packager.required` here) and does not make sense in a strongly typed language (it is just there if you do not know whether your source is from a `require` statement or a source object (for instance when you want to write some image handling library).

### DIFF
--- a/bs-react-native-next/src/components/Image.re
+++ b/bs-react-native-next/src/components/Image.re
@@ -1,31 +1,34 @@
+type uriSource;
+
+[@bs.obj]
+external uriSource:
+  (
+    ~uri: string,
+    ~bundle: string=?,
+    ~method: string=?,
+    ~headers: Js.t('a)=?,
+    ~body: string=?,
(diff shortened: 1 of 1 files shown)


MoOx:
@fhammerschmidt should we do something similar for defaultSource? extract uriSource maker & use it in DefaultSource submodule?

fhammerschmidt:
@MoOx I would not do that, because `DefaultSource` has two creation methods (`fromUri` & `fromRequired`). Also `DefaultSource` is more comparable to `Source` than to `uriSource` because it wraps different types of sources.

PR #359, opened 2019-04-09
bs-platform 5.0.1-beta.1

cknitt:
bs-platform 5.0.1-beta.1 with `"react-jsx": 3` and `[@bs.inline]` support.


bs-react-native-next/src/apis/ActionSheetIOS.bs.js
```diff
@@ -1,2 +1 @@
-// Generated by BUCKLESCRIPT VERSION 5.0.0, PLEASE EDIT WITH CARE
```
MoOx: I guess that's new? Is this cause the name is .bs.js or does bsb just dropping this "feature"?

bs-react-native-next/src/apis/NativeModules.bs.js
```diff
@@ -1,4 +1,4 @@
-// Generated by BUCKLESCRIPT VERSION 5.0.0, PLEASE EDIT WITH CARE
+
```
MoOx: why is there new line in some cases & some cases not?

bs-react-native-next/src/apis/Platform.rei
```diff
@@ -0,0 +1,9 @@
+type os = pri string;
```
MoOx: what does `pri string` means?

bs-react-native-next/src/apis/ActionSheetIOS.bs.js
cknitt: I added `-bs-no-version-header` to bsconfig.json. ;-)

MoOx: oh I missed that :D

bs-react-native-next/src/apis/NativeModules.bs.js
cknitt: Different in bs-platform 5.0.1 for some reason unknown to me.

bs-react-native-next/src/apis/Platform.rei
cknitt: `private` in OCaml (as suggested by Hongbo in https://github.com/BuckleScript/bucklescript/issues/3472#issuecomment-480190956) becomes `pri` in Reason.

Just `type os;` alone won't work as the inline constants are of type `string`.

MoOx: Thanks for the explanation.

bs-react-native-next/bsconfig.json
```diff
@@ -16,6 +16,9 @@
   "ppx-flags": [
     "./node_modules/reason-react/_build/default/jsx_ppx/reactjs_jsx_ppx_v3.exe"
   ],
-  "bsc-flags": ["-bs-super-errors", "-warn-error @a", "-w -41..42-102"],
```
cknitt: `-bs-super-errors` is the default now AFAIK, so I removed it. I added `-bs-no-version-header` to get rid of the version headers in future diffs.

MoOx: Make sense!

MoOx:
Feel free to merge, will fix CI this evening or tomorrow morning (fixed on the website PR - well, failing but for another reason - I have that fixed locally)

PR #362, opened 2019-04-10
bs-react-native-next: bs-platform 5.0.1 & reason-react 0.7.0 🥳

MoOx:
Here we go https://twitter.com/rickyvetter/status/1115964552921927680

I blindly replaced `React.Ref.domRef` by `ReactDOMRe.Ref.t` but I guess we should make something better on the long run.

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,9 +46,8 @@ risk._
 
 Things you need
 
-- `package.json` dep: `"bs-platform": "^5.0.1-beta.1",`
-- `package.json` dep:
-  `"reason-react": "https://github.com/reasonml/reason-react.git#0.7.0",`
+- `package.json` dep: `"bs-platform": "^5.0.1"`
+- `package.json` dep: `"reason-react": "^0.7.0"`
 - `bsconfig.json`: `"reason": { "react-jsx": 3 }`
 
 ---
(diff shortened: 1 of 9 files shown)


cknitt:
I think we could use `ReactDOMRe.domRef` instead of `ReactDOMRe.Ref.t`.

MoOx:
That's indeed the same... What's bothering me is that it's referring to the dom.

cknitt:
`ReactDOMRe.Ref.t` is referring to the DOM, too. I'll ask Ricky about that at ReasonConf. :-)

If you prefer to keep the `ReactDOMRe.Ref.t`, it's fine with me, too.

MoOx:
Yes, they are the same. I didn't express my self correct.

What I mean is that we could probably make our own `React.Ref.t`. This type expect an argument (for dom, well it's easy, `Dom.element`, that's what `ReactDOMRe.domRef` / `ReactDOMRe.Ref.t` are using (the former is an alias for the first). 

For RN I am not sure what we should provide. That's why I've used the dom thing, but I guess we should not use a dom ref, as those are native refs?

cknitt:
Yes, native ref would be better. But we can do this in a separate PR after ReasonConf. :-)

MoOx:
I created https://github.com/reasonml-community/bs-react-native/issues/363 for this

PR #367, opened 2019-04-13
Add zero-cost bindings for the ImageBackground component

sgny:


bs-react-native-next/src/components/ImageBackground.re
```diff
@@ -0,0 +1,57 @@
...
+                   | `center
+                   | `contain
+                   | `cover
+                   | `repeat
+                   | `stretch
+                 ]
+                   =?,
+    ~source: Image.Source.t,
+    ~style: Style.t=?,
+    ~imageStyle: t=?,
```
MoOx: Could you add imageRef too ? https://facebook.github.io/react-native/docs/imagebackground#imageref

sgny: Sure, will take care of that later tonight

PR #368, opened 2019-04-13
Add zero-cost bindings for the ImageEditor API

sgny:


MoOx:
@sgny reminder: you have write access on the repo & using your fork is not required if you want :) (pr are still required to get at least one review from another contributor)

PR #369, opened 2019-04-14
Correct types for Gesture Responder props in components

### LINKED ISSUE #203: Incorrect bindings for gesture handlers
It seems like the bindings for gesture responder handlers are incorrect. They assume the input is `NativeEvent` when it is in fact an object like `{nativeEvent: NativeEvent}`. I'll try to whip up a PR soon but I thought I'd create an issue just to track it.

sgny:
This should correct #203. A separate PR will follow to add the `PanResponder` API. 

`nativeEvent` can be accessed using `Event.nativeEvent` for each handler function and subsequently each key (such as `locationX`) in the `nativeEvent` object can be accessed using `Event.NativeEvent.[name of key]` as below:

```
~onResponderMove=
    (e: Event.t) =>
        e
        |> Event.nativeEvent
        |> Event.NativeEvent.locationX
        |> someFunction
```

where `someFunction` is of type `float => unit` as this handler requires the function to return `unit`.


cknitt:
I think we can merge this, but it did not build.

@MoOx I have no permissions to restart the failed build. Can you restart it? Can you give me permissions?

For further work/discussion on event bindings, I have created #374.

MoOx:
@cknitt are you correctly logged in on circle-ci via your github account? It should be automatic...
I don't think I can't do anything as it's supposed to be synced with github rights...

cknitt:
@MoOx Thanks, you are right, it seems I was not logged in correctly. 🤦‍♂️

PR #370, opened 2019-04-14
Zero-cost Keyboard binding.

cknitt:


bs-react-native-next/src/apis/Keyboard.bs.js
```diff
@@ -1,13 +1,7 @@
...
 
 
-var Easing = /* module */[
-  /* easeIn */"easeIn",
-  /* easeInEaseOut */"easeInEaseOut",
-  /* easeOut */"easeOut",
-  /* linear */"linear",
-  /* keyboard */"keyboard"
-];
+var Easing = /* module */[];
```
MoOx: Is using a module necessary here? Or could bs avoid this empty modules?

cknitt: I asked @bobzhang on Discord about this, but couldn't convince him. Maybe we should open an issue in the BuckleScript repo.

(But although it doesn't look like it, the binding is still zero cost this way, as the JS file will never get imported.)

PR #372, opened 2019-04-14
Redefine FontVariants using [@bs.inline]

sgny:


sgny:
Moved FontVariant out of Style to avoid having to create an interface file for Style as well.

PR #373, opened 2019-04-14
Event.pressEvent defined for touch events

sgny:
This PR updates NativeLayoutEvent and NativeScrollEvent to be more consistent with NativeEvent, defining accessors with the `[@bs.get]` method.

It also better reflects the read-only nature of these objects as all types are abstract with no constructors defined.


sgny:
Sorry pushed a commit to the wrong branch, will return to this PR in a while.

cknitt:
How exactly does it not work?

```re
type event('a) = {. "nativeEvent": 'a};

type pressEventPayload = {
  .
  "changedTouches": array(pressEventPayload),
  "force": float,
  "identifier": int,
  "locationX": float,
  "locationY": float,
  "pageX": float,
  "pageY": float,
  "target": Js.Nullable.t(float),
  "timestamp": int,
  "touches": array(pressEventPayload),
  "data": Js.Nullable.t(string),
};

let payload: pressEventPayload = {
  "changedTouches": [||],
  "force": 1.,
  "identifier": 1,
  "locationX": 0.,
  "locationY": 0.,
  "pageX": 0.,
  "pageY": 0.,
  "target": Js.Nullable.null,
  "timestamp": 0,
  "touches": [||],
  "data": Js.Nullable.null,
};
```

compiles fine for me (at least with bs 5.0.2).

cknitt:
BTW, we only need `Js.Nullable.t` if the value can actually be `null`. If it can be `undefined` only, `option` is fine. (I don't know whether target or data may be `null` though.)

MoOx:
For circular types, you can do

```reason
type a = {
  b: b
} 
and b = {
  a: a
}
```

Instead of using several `type` keywords, you chain type using  `and`.

MoOx:
Oh maybe I write down something too quickly, seems irrelevant. Sorry :)

sgny:
> How exactly does it not work?
> 
> ```c++
> type event('a) = {. "nativeEvent": 'a};
> 
> type pressEventPayload = {
>   .
>   "changedTouches": array(pressEventPayload),
>   "force": float,
>   "identifier": int,
>   "locationX": float,
>   "locationY": float,
>   "pageX": float,
>   "pageY": float,
>   "target": Js.Nullable.t(float),
>   "timestamp": int,
>   "touches": array(pressEventPayload),
>   "data": Js.Nullable.t(string),
> };
> 
> let payload: pressEventPayload = {
>   "changedTouches": [||],
>   "force": 1.,
>   "identifier": 1,
>   "locationX": 0.,
>   "locationY": 0.,
>   "pageX": 0.,
>   "pageY": 0.,
>   "target": Js.Nullable.null,
>   "timestamp": 0,
>   "touches": [||],
>   "data": Js.Nullable.null,
> };
> ```
> 
> compiles fine for me (at least with bs 5.0.2).

It would compile fine, but something like `(e##nativeEvent##touches |. Belt.Array.getUnsafe(0))##locationX` would cause the compiler to complain, not recognising the type of the expression in (), then return nothing (or perhaps undefined, can't remember) whereas the other implementation worked. 

Then defined the types for `touches` and `changedTouches` as some abstract type and then cast them as type `pressEventPayload`. Then the compiler threw an error which I felt implied the cyclicality was a problem (should have noted it).

Anyway, now both seem to work fine. I'll clean up and update the PR.

sgny:
I removed the modules `Event.NativeEvent` (and `Event.NativePressEvent`) and type `t` in favor of the type `Event.pressEvent` and updated all components accordingly.

I did not update `scrollEvent` and `layoutEvent` as there are quite a few changes already; I will handle those with a later PR.

MoOx:
And refmt should normalize those... And git pre-commit hook should run refmt... :)

sgny:
This should be fine now. I had setup yarn for this repo, but reverted to npm a while back and I missed the `refmt` pre-commit hook.

cknitt:
Thanks! I looked at the flow types again and noticed that `PressEvent` is a `ResponderSyntheticEvent` which means that it should have `touchHistory`, too.

We can handle that in a separate PR though. `event` itself (`SyntheticEvent` in the flow types) is missing fields, too.

PR #378, opened 2019-04-15
Add SourceCode.scriptURL (undocumented, but useful).

cknitt:


cknitt:
Moved it to a subfolder called "private".

PR #379, opened 2019-04-15
Zero cost bindings to PanResponder API

sgny:
This will most likely have to be revised before getting merged. Everything works, but it's missing some *magic*.

 `PanResponder.create` creates an object which has the gesture responder `props` (and their relevant values) under the key `panHandlers`. That object is passed to a `View` using the spread notation (i.e. `{...panHandlers}`) under JS, but I am not sure how to handle that under Reason. Spread used to work, but I'm not sure if there is equivalent functionality anymore. I tried on Discord and @rickyvetter suggested it should just work, but I've yet to figure out how.

Obviously, one can make use of the accessors and do something like
```
onResponderStart={panHandlers |> PanResponder.onResponderStart}
```
but that is hardly convenient.


MoOx:
We already have this situation with plenty of props (eg View props available almost everywhere, ScrollView props for all kind of lists etc) and spread isn't possible I am afraid. We will have to be explicit about it...

bs-react-native-next/src/apis/PanResponder.re
```diff
@@ -0,0 +1,76 @@
+type gestureState;
+
+[@bs.obj]
+external gestureState:
```
cknitt: As per https://github.com/reasonml-community/bs-react-native/blob/master/bs-react-native-next/CONTRIBUTING.md#js-objects-returned-in-callbacks, this should be a `Js.t` type.

bs-react-native-next/src/apis/PanResponder.re
```diff
@@ -0,0 +1,76 @@
...
+[@bs.module "react-native"] [@bs.scope "PanResponder"]
+external create: config => t = "";
+
+type panHandlers;
+
+[@bs.get] external panHandlers: t => panHandlers = "";
+
+type callback('a) = Event.t => 'a;
+
+[@bs.get ] external onMoveShouldSetResponder: panHandlers => callback(bool) = "";
```
cknitt: The code is not formatted using `refmt`, even though we have a pre-commit hook that should be doing that.

MoOx: @sgny what are you using to commit? Current repo should automatically setup git hooks as soon as you run "npm install" or "yarn", transparently (so refmt should be run when you commit)

sgny: Sorry I've been using my scripts, but I will update those.

sgny:
> Please/add update .md file.
> Btw if you want to write some docs about it (how to use this), feel free to do it, even it that's just a sentence or 2 :)

I do intend to write documentation, but for now I simply duplicated the code as placeholder.

MoOx:
@cknitt does this look good to you?

PR #382, opened 2019-04-16
Fix refs.

cknitt:
We added a TextInput to our app and discovered that I had modelled refs incorrectly. 😱

Having

```re
type element;
```

and

```re
[@react.component] [@bs.module "react-native"]
external make:
  (
    ~ref: ref=?,
    ...
```

we need to have

```re
[@bs.send] external focus: element => unit = "";
```

not

```re
[@bs.send] external focus: ref => unit = "";
```


MoOx:
Nice catch!

PR #383, opened 2019-04-16
Zero-cost bindings for decay, spring and timing [Animated]

sgny:
With this PR, bindings reflect JS API more closely with the exception of named arguments, i.e.. 
```
Animated.timing(~value, ~config)
``` 
`config` is the output of a `makeConfig` function as before, however, it is intended to be called directly and not indirectly by the `animate` function.

I would appreciate feedback on whether arguments should be named or not.


sgny:
> In the other bindings, we named the make function the same as the type. So I think we should follow that here and name `makeConfig` just `config`.
> 
> Regarding labelled arguments, personally I would only use those when necessary for disambiguation, e.g. when you have two string or two float params that the user might confuse. @MoOx What do you think?

I brought this up, because I don't think they are necessary in `Animated` either, but I wasn't sure if we should break the API beyond what's necessary and kept them (likewise for the `makeConfig`). I do feel breaking at this stage should be OK as it is already a major transition for all users.

Thanks for your comments, it seems we are in agreement. @MoOx what's your take?

cknitt:
I would not worry about breaking changes. Due to the zero cost approach, the `-next` API is very different from the old one anyway. We should rather strive for consistency within the `-next` APIs.

sgny:
> There is a leftover abc.re now?

Sorry I was investigating what went wrong as refmt did not run before.

MoOx:
@sgny looks good to me!

PR #389, opened 2019-04-17
Event.scrollEvent defined for scrollable components

sgny:
I left `LayoutEvent` untouched for an easier-to-mange PR.


MoOx:
Nice :)

PR #394, opened 2019-04-18
Zero-cost binding for Animated.loop

sgny:
I did not want to implement this without investigation the default value provided (-1). Without the config object, the default value is already assumed to be -1 (which means infinite loop), so there is no need for the binding to provide the default value.

### DIFF
--- a/bs-react-native-next/src/apis/Animated.md
+++ b/bs-react-native-next/src/apis/Animated.md
@@ -244,10 +244,9 @@ external parallel:
 external stagger: (float, array(Animation.t)) => Animation.t = "";
 
 [@bs.module "react-native"] [@bs.scope "Animated"]
-external _loop: (Animation.t, {. "iterations": int}) => Animation.t = "loop";
-
-let loop = (~iterations=(-1), ~animation, ()) =>
-  _loop(animation, {"iterations": iterations});
+external loop:
+  (Animation.t, ~config: {. "iterations": int}=?, unit) => Animation.t =
+  "";
 
 type animatedEvent;
 
@@ -270,5 +269,6 @@ let start = Animation.start;
 let stop = Animation.stop;
 
 let reset = Animation.reset;
+
 
 ```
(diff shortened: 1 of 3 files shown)


bs-react-native-next/src/apis/Animated.md
```diff
@@ -244,10 +244,9 @@ external parallel:
 external stagger: (float, array(Animation.t)) => Animation.t = "";
 
 [@bs.module "react-native"] [@bs.scope "Animated"]
-external _loop: (Animation.t, {. "iterations": int}) => Animation.t = "loop";
-
-let loop = (~iterations=(-1), ~animation, ()) =>
-  _loop(animation, {"iterations": iterations});
+external loop:
+  (Animation.t, ~config: {. "iterations": int}=?, unit) => Animation.t =
```
cknitt: Thanks!

Two remarks:
1. Can we use `[@bs.obj]` for the config object (like for the other configs in this file and also following https://github.com/reasonml-community/bs-react-native/blob/master/bs-react-native-next/CONTRIBUTING.md#js-objects-passed-as-props)?

2. For easier usage, I think we should have `loop` with just the `Animation.t` parameter and `loopWithConfig` with the config parameter, too. Then the user does not need to remember the unit at the end.

MoOx: I am not sure about loop & loopWithConfig but I totally understand the reason.

bs-react-native-next/src/apis/Animated.md
```diff
@@ -243,11 +243,12 @@ external parallel:
...
-[@bs.module "react-native"] [@bs.scope "Animated"]
-external _loop: (Animation.t, {. "iterations": int}) => Animation.t = "loop";
+type loopConfig;
+
+[@bs.obj] external loopConfig: (~iterations: int) => loopConfig = "";
 
-let loop = (~iterations=(-1), ~animation, ()) =>
-  _loop(animation, {"iterations": iterations});
+[@bs.module "react-native"] [@bs.scope "Animated"]
+external loop: (Animation.t, loopConfig) => Animation.t = "";
```
cknitt: What I meant was

```re
[@bs.module "react-native"] [@bs.scope "Animated"]
external loop: Animation.t => Animation.t = "loop";

[@bs.module "react-native"] [@bs.scope "Animated"]
external loopWithConfig: (Animation.t, loopConfig) => Animation.t = "loop";
```

sgny: Sorry read your comment too quickly.

I implemented that change, but making config non-optional might have been cleaner.

PR #395, opened 2019-04-18
Zero-cost bindings to ImagePickerIOS

sgny:
I discovered there were some errors with the legacy implementation, namely `canUseCamera` and `canRecordVideos` should take functions of type `bool => unit`.

Likewise `onSuccess` should take functions of type `string => unit` so that the image selected from the library or captured on camera can be acted upon.

Finally, perhaps it's due to only testing on the simulator,  but the type of the error is unclear. Hitting the `Cancel` button calls the `onError` callback, but I only get `undefined`. It's been previously specified as
```
type error = {
  .
  "code": int,
  "message": string,
};
```
but I cannot tell if that is or ever was correct.

Should the documentation also mention that `RCTCameraRoll` has to be linked and appropriate permissions be set?

### DIFF
--- a/bs-react-native-next/src/apis/ImagePickerIOS.bs.js
+++ b/bs-react-native-next/src/apis/ImagePickerIOS.bs.js
@@ -1,0 +1,1 @@
+/* This output is empty. Its source's type definitions, externals and/or unused code got optimized away. */
--- a/bs-react-native-next/src/apis/ImagePickerIOS.md
+++ b/bs-react-native-next/src/apis/ImagePickerIOS.md
@@ -13,10 +13,30 @@ external canUseCamera: (bool => unit) => unit = "";
 [@bs.module "react-native"] [@bs.scope "ImagePickerIOS"]
 external canRecordVideos: (bool => unit) => unit = "";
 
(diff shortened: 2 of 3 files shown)


sgny:
I understand the point about consistency, but with these changes BS API became quite cumbersome to use

cknitt:
I meant that the fields of the config objects should be optional. The way you made the config itself optional will not work (it will result in partial application when the config is omitted).

@MoOx What do you think, is the API too cumbersome to use now?

sgny:
> I meant that the fields of the config objects should be optional. The way you made the config itself optional will not work (it will result in partial application when the config is omitted).
> 
> @MoOx What do you think, is the API too cumbersome to use now?

Again, sorry I read your comment too quickly. However, this is how the select dialog can be opened now with no config options:

```
ImagePickerIOS.openSelectDialog(
	~config=ImagePickerIOS.selectDialogConfig(),
	~onSuccess=successCallback,
	~onError=errorCallback
);
```
I feared omitting () when calling `selectDialogConfig` would cause issues, but that seems fine. Still, the inline type was easier to use, IMO.

PR #400, opened 2019-04-18
bs-react-native-next: PermissionsAndroid

fhammerschmidt:
Saw that the PermissionsAndroid bindings were still missing and decided to give it a try.

It is a bit hard to keep them zero-cost, the usage will be different to the old ones, especially for requestMultiple.

I will try to use them in a project and add examples to the .md in a later PR.

### DIFF
--- a/bs-react-native-next/src/components/WebView.bs.js
+++ b/bs-react-native-next/src/components/WebView.bs.js
@@ -1,0 +1,13 @@
+'use strict';
+
+
+var Source = /* module */[];
+
+var DataDetectorTypes = 0;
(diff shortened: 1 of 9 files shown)


MoOx:
@fhammerschmidt small details, but it seems your commit isn't attached to your github profile. You may have used an email that isn't attached to your github profile to commit. You should have a look there https://github.com/settings/emails & in your `~./gitconfig` in order to be correctly credited for you contribution :) This also might help https://help.github.com/en/articles/setting-your-commit-email-address-in-git

PR #403, opened 2019-04-18
NetInfo bindings

sgny:
Other than using `[@bs.deriving jsConverter]` I'm not familiar with any means to get a variant back from its string representation. I figured that similar to `Platform` users could compare the object (or individual keys using accessors) they create with what is returned in the callback.

An alternative could be to make use of `[@bs.inline]` but that will require additional modules for each type, which comes with its own downsides.


sgny:
Since there is only one possible value for the event (`connectionChange`), adding a function that wraps `addEventListener`, etc. already providing that as the argument will simply shift the cost from the call site to here.

Should we simply do that?

bs-react-native-next/src/apis/NetInfo.md
```diff
@@ -0,0 +1,66 @@
...
+              | `wimax
+            ]
+              =?,
+    unit
+  ) =>
+  info =
+  "";
+
+[@bs.get] external effectiveType: info => string = "";
+[@bs.get] external _type: info => string = "type";
```
fhammerschmidt: Also here: Use `type_` instead.

cknitt: type_ won’t work automagically with bs.obj like _type does.

So I would keep it as it is.

sgny: > Also here: Use `type_` instead.

I transitioned from `type_` to `_type`, etc. a while ago, but forgot why. According to [BuckleScript documentation](https://bucklescript.github.io/bucklescript/Manual.html#_object_label_translation_convention) 

> Else if the rest happens to be a keyword, drop the first '_'

So, for keywords I believe it's better to use `_type` everywhere.

sgny:
I pushed a new commit using `[@bs.inline]` to convert to string. Despite polluting the tree with additional modules, the actual interface is nicer.

I'd appreciate feedback on which is preferred.

cknitt:
> Since there is only one possible value for the event (`connectionChange`), adding a function that wraps `addEventListener`, etc. already providing that as the argument will simply shift the cost from the call site to here.
> 
> Should we simply do that?

This could be done with zero cost, see https://bucklescript.github.io/docs/en/function#fixed-arguments.

But I am fine with keeping things the way it is so that it matches the JS API and won't change in case a new event type is added in the future.

bs-react-native-next/src/apis/NetInfo.md
```diff
@@ -0,0 +1,43 @@
...
+id: apis/NetInfo
+title: NetInfo
+wip: true
+---
+
+```reason
+type info = {
+  .
+  "_type": NetInfoType.t,
+  "effectiveType": NetInfoEffectiveType.t,
```
cknitt: Can we name the modules `ConnectionType` and `EffectiveConnectionType` to match the RN docs?

The actual file/module names can be `NetInfo_ConnectionType` and `NetInfo_EffectiveConnectionType`, and then we can do:

```re
module ConnectionType = NetInfo_ConnectionType;
module EffectiveConnectionType = NetInfo_EffectiveConnectionType;
```

sgny: I doubt we would have name collision if we simply renamed the modules `ConnectionType` and `EffectiveConnectionType`, would that be alright?

bs-react-native-next/src/apis/NetInfo.md
```diff
@@ -0,0 +1,43 @@
...
+};
+
+[@bs.module "react-native"] [@bs.scope "NetInfo"]
+external addEventListener:
```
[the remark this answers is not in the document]
sgny: Thanks, I wasn't aware of this. I'll also review other bindings.

sgny:
> > Since there is only one possible value for the event (`connectionChange`), adding a function that wraps `addEventListener`, etc. already providing that as the argument will simply shift the cost from the call site to here.
> > Should we simply do that?
> 
> This could be done with zero cost, see https://bucklescript.github.io/docs/en/function#fixed-arguments.
> 
> But I am fine with keeping things the way it is so that it matches the JS API and won't change in case a new event type is added in the future.

That's a neat trick I wasn't aware of. Thanks!

bs-react-native-next/src/apis/NetInfo.md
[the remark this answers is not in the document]
sgny: `NetInfo.addEventListener` and `NetInfo.isConnected.addEventListener` return an object with key `remove` of type `unit => unit`. Likewise with `BackHandler`.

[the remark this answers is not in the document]
sgny: @cknitt Should we pretend that object is of type `EventSubscription.t` and define a getter? Then we'd have (quoting only some of the module)
```
type s = unit => unit;

[@bs.get] external remove: EventSubscription.t => s = "";

[@bs.module "react-native"] [@bs.scope "NetInfo"]
external addEventListener:
  ([@bs.string] [ | `connectionChange], info => unit) => EventSubscription.t =
  "";
```
Is there a way to express this more compactly, i.e. something like
```
[@bs.get] external remove: EventSubscription.t => (unit => unit) = "";
```

bs-react-native-next/src/apis/NetInfo.md
```diff
@@ -0,0 +1,48 @@
...
+---
+
+```reason
+type info = {
+  .
+  "type_": ConnectionType.t,
+  "effectiveType": EffectiveConnectionType.t,
+};
+
+type remove = {
```
cknitt: Why did you define your own `remove` when we already have `EventSubscription.t` in a common `EventSubscription` module?

sgny: I checked the implementation in RN code and some addEventListener functions have output type `EmitterSubscription` whereas NetInfo and BackHandler define `{remove: ....}`.

cknitt: Ah, ok, interesting. Then let's keep it as it is!

bs-react-native-next/src/apis/NetInfo.md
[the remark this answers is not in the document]
cknitt: Sorry, didn't see this earlier. Let's keep the remove stuff as it is.

bs-react-native-next/src/apis/NetInfo.md
cknitt: IMHO they should be in the NetInfo scope (accessible as `NetInfo.ConnectionType` and `NetInfo.EffectiveConnectionType`). If you want to have them in separate files, I think we should follow the convention I described above, same as I did for `WebView` and related types.

bs-react-native-next/src/apis/NetInfo.rei
```diff
@@ -0,0 +1,62 @@
...
+    let net3G: t;
+
+    [@bs.inline "4g"]
+    let net4G: t;
+
+    [@bs.inline "unknown"]
+    let unknown: t;
+  };
+type info =
+    {. "effectiveType": EffectiveConnectionType.t, "_type": ConnectionType.t};
```
cknitt: Formatting looks weird in this file - did this go through refmt?

Could use a few newlines, too. :wink:

sgny: Missed that due to the file type, fixed now.

Sorry about the assault on unsuspecting eyes!

PR #405, opened 2019-04-18
Zero-cost bindings to TimePickerAndroid

sgny:


bs-react-native-next/src/apis/TimePickerAndroid.re
```diff
@@ -0,0 +1,29 @@
...
+  ) =>
+  options =
+  "";
+
+type response = {
+  .
+  action: string,
+  hour: int,
+  minute: int,
+};
```
MoOx: I am surprised that quote aren't necessary here 😮

bs-react-native-next/src/apis/TimePickerAndroid.md
```diff
@@ -0,0 +1,39 @@
...
+type response = {
+  .
+  action: string,
+  hour: int,
+  minute: int
+};
+
+
+[@bs.module "react-native"] [@bs.scope "TimePickerAndroid"]
+external _open: options => Js.Promise.t(response) = "open";
```
fhammerschmidt: I guess you want `open_` here, because it is not unused and not only used internally.

bs-react-native-next/src/apis/TimePickerAndroid.re
```diff
@@ -0,0 +1,29 @@
...
+
+[@bs.module "react-native"] [@bs.scope "TimePickerAndroid"]
+external _open: options => Js.Promise.t(response) = "open";
```
MoOx: shouldn't we do `open_` instead ? I remember that reason-react change some thing `_likethat` in favor of things `likethat_` not sure why.

sgny: fixed

bs-react-native-next/src/apis/TimePickerAndroid.re
sgny: I'm more surprised that they weren't.

fixed

bs-react-native-next/src/apis/TimePickerAndroid.re
sgny: Further to my comment in #403, perhaps this should really be `_open`?

bs-react-native-next/src/apis/TimePickerAndroid.md
sgny: I thought `open_` and `_open` were equally fine, but probably not, see #403.

bs-react-native-next/src/apis/TimePickerAndroid.re
cknitt: Right, for object labels (`f##_open` in your link) and for bs.obj labelled arguments we need to have the underscore prefix, it won't work any other way.

For externals we are free to decide, and as the leading `_` usually means "unused", I am also in favor of `open_` in this case.

bs-react-native-next/src/apis/TimePickerAndroid.md
```diff
@@ -0,0 +1,39 @@
...
+  "action": string,
+  "hour": int,
+  "minute": int
+};
+
+
+[@bs.module "react-native"] [@bs.scope "TimePickerAndroid"]
+external open_: options => Js.Promise.t(response) = "open";
+
+[@bs.module "react-native"] [@bs.scope "TimePickerAndroid"] [@bs.val]
```
cknitt: I think the [@bs.val] are superfluous.

bs-react-native-next/src/apis/TimePickerAndroid.re
MoOx: Would be nice to have bs.obj to work with suffix instead of prefix...
I would prefer `open_` too... :'(

bs-react-native-next/src/apis/TimePickerAndroid.re
cknitt: Quotes **are** required.

PR #406, opened 2019-04-18
Zero-cost bindings to DatePickerAndroid

sgny:


bs-react-native-next/src/apis/DatePickerAndroid.md
```diff
@@ -0,0 +1,44 @@
...
+  "action": string,
+  "year": int,
+  "month": int,
+  "day": int,
+};
+
+[@bs.module "react-native"] [@bs.scope "DatePickerAndroid"]
+external open_: options => Js.Promise.t(response) = "open";
+
+[@bs.module "react-native"] [@bs.scope "DatePickerAndroid"] [@bs.val]
```
cknitt: Thanks!

I think the `[@bs.val]` are superfluous.

PR #407, opened 2019-04-18
Zero-cost bindings to ToastAndroid

sgny:


bs-react-native-next/src/apis/ToastAndroid.md
```diff
@@ -0,0 +1,38 @@
...
+title: ToastAndroid
+wip: true
+---
+
+```reason
+type message = string;
+type duration;
+type gravity;
+
+[@bs.module "react-native"] [@bs.scope "ToastAndroid"] [@bs.val]
```
cknitt: Nice work! 😀

Here, too, I think the [@bs.val] are superfluous though.

sgny: Thanks, they were.

PR #410, opened 2019-04-18
Platform docs + add web as a safe platform

MoOx:
Feedback welcome

### DIFF
--- a/bs-react-native-next/src/apis/Platform.md
+++ b/bs-react-native-next/src/apis/Platform.md
@@ -1,17 +1,73 @@
 ---
 id: apis/Platform
 title: Platform
-wip: true
+officialDoc: https://facebook.github.io/react-native/docs/platform-specific-code
 ---
 
+## `Platform.os`
+
+React Native provides a module that detects the platform in which the app is
+running. You can use the detection logic to implement platform-specific code.
+Use this option when only small parts of a component are platform-specific.
+
 ```reason
-type os = pri string;
-
-[@bs.module "react-native"] [@bs.scope "Platform"] external os: os = "OS";
-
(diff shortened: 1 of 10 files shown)


bs-react-native-next/src/apis/Platform.md
```diff
@@ -1,17 +1,73 @@
...
 
-[@bs.module "react-native"] [@bs.scope "Platform"] external os: os = "OS";
+- `Platform.ios` when running on iOS
+- `Platform.android` when running on Android.
+- `Platform.web` when running on the web (via
+  [`react-native-web`](https://github.com/necolas/react-native-web)).
 
-[@bs.inline "ios"]
-let ios: os;
+If you need unsupported platform, you can use `Platform.unsafe(string)`.
```
cknitt: ... *an* unsupported platform ...

bs-react-native-next/src/apis/Platform.md
```diff
@@ -1,17 +1,73 @@
...
-let android: os;
+For conditional style depending on the platform, you can do the following:
+
+```reason
+let styles =
+  StyleSheet.create(
+    Style.{
+      "wrapper":
+        style(
+          ~width=(Platform.os == Platform.ios ? 100. : 200.)->Size.pct,
```
cknitt: Beginners might not be familiar with the pipe first operator yet, maybe just write

```re
          ~width=Size.pct(Platform.os == Platform.ios ? 100. : 200.),
```

MoOx: adjusted but I think in practice it's definitely nicer to educate people, especially when the operator make sense like if that's a unit (eg: we write 100% & not %100 :D )

PR #412, opened 2019-04-18
Complete implementation of CoreEventTypes

sgny:
Documentation does not list if any keys with types marked with ? can become `null` or not, so I chose to define all such keys to be of type `Js.Nullable.t('a)`.

Likewise, I chose to define all keys of type `number` as `float` until we can determine if they are in fact integers.

Finally, `responderSyntheticEvent` is exactly `syntheticEvent` with the additional key `touchHistory`. It might help to add a comment indicating such, so that if `syntheticEvent` is updated, any future commits mirror any such changes in `responderSyntheticEvent`.

### DIFF
--- a/bs-react-native-next/src/apis/Event.md
+++ b/bs-react-native-next/src/apis/Event.md
@@ -29,6 +29,7 @@ type syntheticEvent('a) = {
 
 type responderSyntheticEvent('a) = {
   .
+  // synthethicEvent keys
   "bubbles": Js.Nullable.t(bool),
   "cancelable": Js.Nullable.t(bool),
   "currentTarget": float,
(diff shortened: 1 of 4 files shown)


bs-react-native-next/src/apis/Event.md
```diff
@@ -7,7 +7,92 @@ wip: true
...
+  "isPropagationStopped": unit => bool,
+  "isTrusted": Js.Nullable.t(bool),
+  "nativeEvent": 'a,
+  "persist": unit => unit,
+  "target": Js.Nullable.t(float),
+  "timeStamp": float,
+  "_type": Js.Nullable.t(string),
+};
+
+type responderSyntheticEvent('a) = {
```
cknitt: Please add a comment like you suggested.

sgny: Added

PR #417, opened 2019-04-19
reason-react-native: add Animated components

MoOx:
We cannot add them into as Animated.* otherwise we run into a cycle dep error (style->animated->style)

This PR is just a copy/paste of exisiting compo with `[@bs.scope "Animated"]`.

I used include for Image & ScrollView method, not sure if we should do it or not.

### DIFF
--- a/reason-react-native/src/components/AnimatedImage.bs.js
+++ b/reason-react-native/src/components/AnimatedImage.bs.js
@@ -1,0 +1,11 @@
+'use strict';
+
+var Image$ReactNative = require("../Image.bs.js");
+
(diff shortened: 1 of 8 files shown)


sgny:
Thank you for adding these, I was simply using `React.createElementVariadic` and `Animated.createAnimatedComponent` in place of `Animated.View`.

However, I'm not sure if we should target the interface of `Animated.View` as you have or replicate it as I have (please see #418).

BTW it turned out that `include Image` was a great inspiration. Without that, the JS output becomes huge.

PR #418, opened 2019-04-19
Animated Components (Image, ScrollView, Text, View)

sgny:
As far as I know, these components are simply the regular component of the same name, wrapped in `Animated.createAnimatedComponent`.

If that is indeed the case, I propose we reimplement the components and not simply create bindings.

PR that @MoOx contributed has the advantage that it would be fine even if RN implementation changes and this would need to be revisited.

### DIFF
--- a/reason-react-native/src/apis/Animated.bs.js
+++ b/reason-react-native/src/apis/Animated.bs.js
@@ -2,6 +2,7 @@
 
 var Caml_option = require("bs-platform/lib/js/caml_option.js");
 var ReactNative = require("react-native");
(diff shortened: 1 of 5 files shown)


cknitt:
Even though this needs to be revisited if RN implementation changes, I think I like this better than #417 because it avoids duplication.

Let's see what @MoOx thinks though.

BTW, React Native also has AnimatedFlatList and AnimatedSectionList: https://github.com/facebook/react-native/tree/master/Libraries/Animated/src/components

PR #430, opened 2019-04-22
Fix Types.edgeInsets to match Flow types.

cknitt:
According to the flow types, `insets` should be `edgeInsets`,  and its fields should be optional.

I am not too happy with the module name `Types` either BTW, but couldn't think of anything better.


MoOx:
> I am not too happy with the module name Types either BTW, but couldn't think of anything better.

Same here, but since it's kind of internal, we could rename/move this later

PR #433, opened 2019-04-22
reason-react-native: fixes & improvements on Touchables

MoOx:
- TouchableWithoutFeedback doesn't have `style` (it's cloning its child)
- I reordered props so we can easier track what props are from TouchableWithoutFeedback
- I added `tvParallaxProperties` type & a magic to have defaults (RN accept object or `true` to get default values
- Fixed accessibility props for all touchables

### DIFF
--- a/reason-react-native/src/types/TV.re
+++ b/reason-react-native/src/types/TV.re
@@ -1,0 +1,18 @@
+type parallax;
+
+[@bs.obj]
+external parallax:
+  (
+    ~shiftDistanceX: float,
+    ~shiftDistanceY: float,
+    ~tiltAngle: float,
+    ~magnification: float,
+    ~pressMagnification: float,
+    ~pressDuration: float,
+    ~pressDelay: float,
+    unit
+  ) =>
+  parallax =
+  "";
+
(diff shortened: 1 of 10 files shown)


reason-react-native/src/types/TV.re
```diff
@@ -0,0 +1,12 @@
+type parallax = {
```
cknitt: For stuff that we pass in to components, our convention is to use bs.obj, not Js.t.

reason-react-native/src/types/TV.re
```diff
@@ -0,0 +1,12 @@
...
+  "shiftDistanceX": float,
+  "shiftDistanceY": float,
+  "tiltAngle": float,
+  "magnification": float,
+  "pressMagnification": float,
+  "pressDuration": float,
+  "pressDelay": float,
+};
+
+let parallaxDefault: unit => parallax = _ => true->Obj.magic;
```
cknitt: Why not

```re
let parallaxDefault: parallax = true->Obj.magic;
```

?

MoOx:
Made that too quickly while working on jsx3-compat. Fixed things here.

PR #434, opened 2019-04-22
Zero-cost bindings to DrawerLayoutAndroid

sgny:
Fix type signature for `onDrawerStateChanged` callback

### DIFF
--- a/reason-react-native/src/components/DrawerLayoutAndroid.re
+++ b/reason-react-native/src/components/DrawerLayoutAndroid.re
@@ -29,7 +29,7 @@ external make:
                        =?,
     ~onDrawerOpen: unit => unit=?,
     ~onDrawerSlide: drawerSlideEvent => unit=?,
-    ~onDrawerStateChanged: [ | `Idle | `Dragging | `Settling] => unit=?,
+    ~onDrawerStateChanged: AndroidInteractionState.t => unit=?,
     ~drawerBackgroundColor: Style.color=?,
     ~statusBarBackgroundColor: Style.color=?,
     // View props
--- a/reason-react-native/src/components/DrawerLayoutAndroid.md
+++ b/reason-react-native/src/components/DrawerLayoutAndroid.md
@@ -1,4 +1,3 @@
-
 ---
 id: components/DrawerLayoutAndroid
 title: DrawerLayoutAndroid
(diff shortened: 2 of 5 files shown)


reason-react-native/src/components/DrawerLayoutAndroid.re
```diff
@@ -29,7 +29,7 @@ external make:
                        =?,
     ~onDrawerOpen: unit => unit=?,
     ~onDrawerSlide: drawerSlideEvent => unit=?,
-    ~onDrawerStateChanged: [ | `Idle | `Dragging | `Settling] => unit=?,
```
MoOx: bs.string can't be enough here?

sgny: I couldn't get it to compile with the [@bs.string] annotation, but perhaps I'm missing how to properly annotate an inline type definition.

On the other hand, the callback function will be passed something like "Idle", etc. To act on that  we cannot make use of `[@bs.string]` as that's only good for `external` definitions and not `let` bindings.

I think this is exactly the same issue we've run into with other bindings and their callback functions.

reason-react-native/src/components/DrawerLayoutAndroid.re
```diff
@@ -29,7 +29,7 @@ external make:
...
+    ~onDrawerStateChanged: AndroidInteractionState.t => unit=?,
```
cknitt: Yes, as it is a callback param, bs.string won't work here.

But why is it called AndroidInteractionState? Flow has

```js
type DrawerStates = 'Idle' | 'Dragging' | 'Settling';
```

The plural is weird, but I would just call it DrawerLayoutAndroid.DrawerState.t.

sgny: For `ViewPagerAndroid` they are called `PageScrollState`. I'm aware it will be removed from RN core, but still it's useful to have.

Instead of two identical type definitions, I figured a common one would work better.

cknitt: Hmm, ok - what do you think @MoOx?

MoOx: Works for me the way it is, anyway this types are kind of internal types (users won't directly use it except for edge cases?) so no big deal if we want to revisit later.

Issue #429, opened 2019-04-22
Provide ref prop for all components?


cknitt:
@MoOx @sgny I saw that you added some Android components yesterday. 👍

I noticed that e.g. `ProgressBarAndroid` has a `ref` prop, but there is nothing you can do with a ref once you have obtained it (no functions using that ref are defined).

To keep consistency between components, do we want to

1. add a `ref` prop to all of them or
2. only add `ref` props to components that actually have functions defined that use the ref (like `TextInput.focus`)

?

cknitt:
Might make sense to provide `ref` everywhere to be able to use `setNativeProps`, `measure` etc.

https://facebook.github.io/react-native/docs/direct-manipulation.html

Then we could have something like e.g. for `FlatList`:

```re
type element;
type ref = React.Ref.t(Js.nullable(element));

// Convert FlatList ref to ScrollView ref to call ScrollView methods.
external scrollViewRef: ref => ScrollView.ref = "%identity";

// Convert FlatList ref to Component ref to call Component methods (setNativeProps etc.).
external componentRef: ref => Component.ref = "%identity";
```

MoOx:
IMO we should have `ref` for every component like you explained. We could even make a small stupid automated rule for that =D (just a script that check `type element;\n` is in each `src/components/*.re` files

sgny:
Many components do not have methods defined on `ref`, but I believe we should include it for those which support it. Users will probably find some use for it, especially since `ref` returns _a lot_ of information. 

Some components do not seem to support `ref`s, however. Functional components do not, as already explained in RN documentation, however, something like `DatePickerIOS` does not support it either.

I have noted that `View` is missing the `ref` prop, so I will create a PR for that. I will check which `View` derivatives do not support it, so I will need a little time.

cknitt:
@sgny Please wait, I have already started working on the ref stuff!

sgny:
> @sgny Please wait, I have already started working on the ref stuff!

I haven't started yet, I'll hold off on that then.

cknitt:
Done with #435 and #452.

PR #440, opened 2019-04-23
RFC: Converting element types

cknitt:
DO NOT MERGE!

This is for discussion/as an example on how the conversion between different element types as discussed in #429 could work.


reason-react-native/src/components/FlatList.re
```diff
@@ -178,49 +182,3 @@ external make:
   ) =>
   React.element =
   "FlatList";
-
-// VirtualizedList methods
```
cknitt: Don't need these here anymore, as they are already defined on `VirtualizedList.element` and we can convert using `asVirtualizedList`.

reason-react-native/src/components/TextInput.re
```diff
@@ -272,7 +274,3 @@ external make:
 [@bs.send] external isFocused: element => bool = "";
 
 [@bs.send] external clear: element => unit = "";
-
-[@bs.send] external focus: element => unit = "";
-
-[@bs.send] external blur: element => unit = "";
```
cknitt: These are already defined on `Component.element`.

MoOx:
I like the approach. Not sure about the as*. But to* isn't better & loose some sense.

That said, it might feel not intuitive to need to what you have to convert to to use a method...

`TextInput.asComponent` "just" to focus or blur might feel overkill.

Can't we do things like `ComponentMethods.re` & `include ComponentMethods` inside TextInput (& same for FlatList with ScrollView (ScrollViewMethods?) etc...)?

sgny:
Overall I like this, but I also agree that some edges have to be smoothed a bit.

According to the documentation you have linked

> The methods described here are available on most of the default components provided by React Native. 

As far as I know, what _most_ means is sadly not documented and would require testing.

It would be great if we could define the `ref` prop as `React.Ref.t(Js.nullable(Component.element))` for most components that support these methods, instead of defining an `element` type for each. Then `focus`, `blur`, and `measure` could simply be called without conversion on those components.

One refinement I would recommend would be to define a more specialized type for `VirtualizedList`, etc. Then type conversions would be from the more specialized to the more general (i.e. `FlatList -> Component.element`) so that it is impossible to apply  specific methods to something like `Image` for example. Then `focus`, `blur`, etc would require a conversion only for those components.

cknitt:
@MoOx This would need to be functors then, right?

```re
type element;

include ComponentMethods.Make({
  type elementType = element;
});

include ScrollViewMethods.Make({
  type elementType = element;
});

include VirtualizedListMethods.Make({
  type elementType = element;
});
```

MoOx:
@cknitt Yeah I guess so :)

I think from a user point of view, we should definitely try to avoid conversions. I prefer to have a bit more code to maintain than having to do non intuitive weird stuff

sgny:
Perhaps we are distinguishing between each component too much as `View.element` vs `Text.element` does not really yield much benefit when all allowed functions are the same, but the design would be sound so this is our best idea so far, thanks @MoOx.

I tested functors a bit last night, and I would recommend 
```
module VirtualizedListMethods = VirtualizedListMethods.Make({
  type elementType = element;
});
```
instead of `include` since this would allow calling the function as `ScrollView.VirtualizedListMethods.*` instead of `ScrollView.VirtualizedListMethods.Make.*`

cknitt:
@sgny `Text.element` has two additional allowed functions compared to `View.element`. `include` allows us to call the functions directly as `ScrollView.*`.

cknitt:
Updated this PR to show what things would look like using functors.

reason-react-native/src/components/ComponentMethods.re
```diff
@@ -0,0 +1,45 @@
...
+type measureInWindowResult = {
+  .
+  "x": float,
+  "y": float,
+  "width": float,
+  "height": float,
+};
+
+module Make = (T: {type t;}) => {
+  [@bs.module "react-native"] external findNodeHandle: T.t => nodeHandle = "";
```
cknitt: Not too happy about this one. Here I feel that the earlier approach would be nicer where we have a common `Component.ref` to which one can convert using `asComponent`, and `findNodeHandle` just takes a `Component.ref`.

cknitt:
One disadvantage is that this makes it impossible to write generic code operating on component refs, like

```re
[| someTextInputRef, someScrollViewRef, someFlatListRef |]
  ->Belt.Array.forEach(ref => ref->Component.someCommonComponentMethod);
```

But I guess that's a pretty rare use case. With the earlier approach, I would have used `asComponentRef` on each and therefore have had an array of `Component.ref`s.

MoOx:
I would prefer to have something intuitive & handy for 80% of uses cases.
We have to think about new reasonml users :)

sgny:
> include allows us to call the functions directly as ScrollView.*.

Oops, I seem to have been asleep at the wheel, you're right.

> One disadvantage is that this makes it impossible to write generic code operating on component refs, like
> 
> ```c++
> [| someTextInputRef, someScrollViewRef, someFlatListRef |]
>   ->Belt.Array.forEach(ref => ref->Component.someCommonComponentMethod);
> ```
> 
> But I guess that's a pretty rare use case. With the earlier approach, I would have used `asComponentRef` on each and therefore have had an array of `Component.ref`s.

Unless `type element` is excluded from interface files, you could always define your own type conversion functions to work around that. As we avoid defining interface files at the moment in the first place (except for inlined types), that should be no problem for now.

cknitt:
Here is a refined approach. Now in the components, we'd just do

```re
include NativeElement;
```

for "standard" components without additional methods, but e.g.

```
include VirtualizedListElement;
```

for VirtualList, FlatList etc.

cknitt:
Another big advantage is that now e.g. `Virtualized.element` and `FlatList.element` is actually the same type.

reason-react-native/src/elements/VirtualizedListElement.re
```diff
@@ -0,0 +1,14 @@
...
+  type t = element;
+});
+
+include ScrollViewMethods.Make({
+  type t = element;
+});
+
+include NativeMethods.Make({
+  type t = element;
+});
```
MoOx: Shouldn't we include in the reverse order? In case something more specific overwrite a more generic method?
Not sure if that's possible tho

cknitt: Overriding is not possible.

cknitt:
@MoOx Thanks for the approval, but I will do another update this evening or tomorrow evening as this is incomplete in the current state.

MoOx:
Yeah I know, I was only validating the idea :)

cknitt:
See #452 for the complete implementation.

PR #444, opened 2019-04-25
Expose ViewPagerAndroid callback types for reuse (jsx3-compat)

MoOx:
Requirements for #415

### DIFF
--- a/reason-react-native/src/components/ViewPagerAndroid.re
+++ b/reason-react-native/src/components/ViewPagerAndroid.re
@@ -1,6 +1,15 @@
 type element;
 type ref = React.Ref.t(Js.nullable(element));
 
+type pageScrollEvent =
+  Event.syntheticEvent({
+    .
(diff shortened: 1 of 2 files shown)


reason-react-native/src/components/ViewPagerAndroid.re
```diff
@@ -1,22 +1,25 @@
...
 type ref = React.Ref.t(Js.nullable(element));
 
+type onPageScrollArgs =
+  Event.syntheticEvent({
+    .
+    "position": int,
+    "offset": float,
+  });
+
+type onPageSelectedArgs = Event.syntheticEvent({. "position": int});
```
cknitt: For consistency with other components, shouldn't this be named `pageSelectedEvent` and the one above `pageScrollEvent`?

MoOx:
Fixed it! I also added children btw at it seems this component accept children.

Issue #413, opened 2019-04-19
bs-react-native-jsx3-compat


MoOx:
This package is going to be a copy of bs-react-native but implementation of components (apis won't change) will use reason-react-native under the hood.
This will be a nice solution for people that will migrate to jsx3 with things like https://bloodyowl.github.io/blog/2019-04-19-an-alternative-migration-path-for-reason-react/

I am pushing the base (bs-react-native copy as bs-react-native-jsx3-compat) to master branch & will make PR so we can review diff.
I will also make several pr for missing props to new bindings because I found some while starting this compat layer.

MoOx:
I am around 50%. Should open a pr this week.

MoOx:
75%

MoOx:
98% :)

MoOx:
Available as pre-alpha PR https://github.com/reasonml-community/bs-react-native/pull/472

PR #446, opened 2019-04-26
reason-react-native: TabBarIOS & TabBarIOS.Item

MoOx:
Requirement for #413

### DIFF
--- a/reason-react-native/src/components/TabBarIOS.re
+++ b/reason-react-native/src/components/TabBarIOS.re
@@ -1,0 +1,196 @@
+type element;
+type ref = React.Ref.t(Js.nullable(element));
+
+[@react.component] [@bs.module "react-native"]
+external make:
+  (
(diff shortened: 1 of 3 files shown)


reason-react-native/src/components/TabBarIOS.re
```diff
@@ -0,0 +1,196 @@
...
+                      ]
+                        =?,
+      ~removeClippedSubviews: bool=?,
+      ~renderToHardwareTextureAndroid: bool=?,
+      ~shouldRasterizeIOS: bool=?,
+      ~style: Style.t=?,
+      ~testID: string=?
+    ) =>
+    React.element =
+    "Item";
```
cknitt: I think this should be TabBarIOS.Item

cknitt: Ah no. You have the scope set.

PR #448, opened 2019-04-26
Add Color module.

cknitt:
@MoOx As discussed on Discord:

* This replaces `Style.color` with `Color.t`.
* The latter is now not an abstract type anymore, but an alias to `string`.
* This way people can just pass strings to color props like they can in JS.
* This makes the API easier to use, but less safe.
* For mitigation, we provide helper functions and constants for safe color creation/use in the `Color` module.


sgny:
Being able to pass strings has upside in ease and downside in less safety, on the other hand, now people will not make mistakes with named colors (`cornflowerblue`, etc.) if they use the constants and not simply pass their names as string.

We could always move back to an abstract type, simply by adding a type-conversion function and defining an appropriate interface for the module. So overall, I agree this is a good move.

Great work and thanks for the huge effort.

PR #449, opened 2019-04-26
Define event names for listeners using variants

sgny:


reason-react-native/src/apis/AppState.re
```diff
@@ -13,7 +13,12 @@ let inactive = "inactive";
 external currentState: t = "currentState";
 
 [@bs.scope "AppState"] [@bs.module "react-native"]
-external addEventListener: (string, unit => unit) => EventSubscription.t = "";
+external addEventListener:
+  ([@bs.string] [ | `change | `memoryWarning], unit => unit) =>
```
cknitt: According to https://facebook.github.io/react-native/docs/appstate.html, the change handler has the new app state as a parameter. But it seems the memory warning handler has none.

We should be able to model this with this trick: https://bucklescript.github.io/docs/en/function#special-case-event-listeners

sgny: Neat trick, it's curious they have pigeonholed it as a special case for event listeners.

PR #451, opened 2019-04-26
Zero-cost bindings to CameraRoll API

sgny:
Since `?` in flow does not distinguish between `null` and `undefined`, I kept optional keys to be of type `Js.Nullable.t('a)` as before.

### DIFF
--- a/reason-react-native/src/apis/CameraRoll.md
+++ b/reason-react-native/src/apis/CameraRoll.md
@@ -70,14 +70,15 @@ type photoIdentifiersPage = {
 };
 
 [@bs.module "react-native"] [@bs.scope "CameraRoll"]
-external saveToCameraRoll:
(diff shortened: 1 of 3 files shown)


reason-react-native/src/apis/CameraRoll.md
```diff
@@ -0,0 +1,83 @@
...
+  "page_info": {
+    .
+    "has_next_page": bool,
+    "start_cursor": Js.Nullable.t(string),
+    "end_cursor": Js.Nullable.t(string),
+  },
+};
+
+[@bs.module "react-native"] [@bs.scope "CameraRoll"]
+external saveToCameraRoll:
```
cknitt: Maybe we should do have both

```saveToCameraRoll(tag)```

and

```saveToCameraRollWithType(tag, type)```

Then we wouldn't need the unit.

PR #456, opened 2019-04-27
Add onAccessibilityEscape (View prop)

sgny:


sgny:
> 👍 - can we have the prop in correct alphabetical order though?

I followed the order in RN documentation, but we do have everything sorted...

PR #458, opened 2019-04-28
PushNotificationIOS

cknitt:
PushNotificationIOS is super weird. Poor documentation, 4 different notification types, some method with promises, some with callbacks, ...

Hope I got it correct.

### DIFF
--- a/reason-react-native/src/apis/PushNotificationIOS.md
+++ b/reason-react-native/src/apis/PushNotificationIOS.md
@@ -1,0 +1,203 @@
+---
+id: apis/PushNotificationIOS
+title: PushNotificationIOS
+wip: true
+---
+
+```reason
+module Notification = {
+  type t;
+
+  [@bs.send] [@bs.return nullable]
+  external getAlert: t => option(Js.Json.t) = "";
+
+  [@bs.send] [@bs.return nullable]
+  external getMessage: t => option(Js.Json.t) = "";
+
+  [@bs.send] [@bs.return nullable]
+  external getSound: t => option(string) = "";
+
+  [@bs.send] [@bs.return nullable]
+  external getCategory: t => option(string) = "";
+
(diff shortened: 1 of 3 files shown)


reason-react-native/src/apis/PushNotificationIOS.md
```diff
@@ -0,0 +1,203 @@
...
+};
+
+[@bs.module "react-native"] [@bs.scope "PushNotificationIOS"]
+external getDeliveredNotifications:
+  (array(deliveredNotification) => unit) => unit =
+  "";
+
+[@bs.module "react-native"] [@bs.scope "PushNotificationIOS"]
+external removeDeliveredNotifications: (~identifiers: array(string)) => unit =
+  "";
```
sgny: `~identifiers` isn't necessary here, did you include that to hint what argument the function requires?

cknitt: Yes, exactly. I was a bit hesitant about this, too, but otherwise it is not really clear what this `array(string)` really is.

sgny:
Any methods you've tested/not tested or that we should take a particular look at? I looked through and they seem alright to me. As you said, however, API's poorly documented and hard to figure out.

However, perhaps it's sufficient to leave this as a best effort. Since this module will be removed from core, perhaps there will be API changes or perhaps most users will simply go for an alternative.

I used `wix/react-native-notifications` (RNN) for iOS and `react-native-firebase`(FB) for Android until last fall (haven't maintained that app since) and that's only because the former appeared at a standstill since last year and GCM was being phased out on Android. Since then, they have introduced support for FCM and have been working on the iOS side as well. In light of that, I'm not sure using this module makes sense at the moment.

cknitt:
Thanks! I have not tested these methods. Let's leave this as best effort as you suggested.

PR #460, opened 2019-04-28
fix CameraRoll.saveToCameraRollOverrideType

sgny:
Prompted by #459

### DIFF
--- a/reason-react-native/src/apis/CameraRoll.md
+++ b/reason-react-native/src/apis/CameraRoll.md
@@ -75,7 +75,7 @@ external saveToCameraRoll: string => Js.Promise.t(string) = "";
 [@bs.module "react-native"] [@bs.scope "CameraRoll"]
 external saveToCameraRollOverrideType:
   (string, ~_type: [@bs.string] [ | `photo | `video]) => Js.Promise.t(string) =
-  "";
+  "saveToCameraRoll";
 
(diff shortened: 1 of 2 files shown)


MoOx:
I thing we should totally avoid this empty string.

cknitt:
I am not sure about that. Avoiding the empty string can cause other mistakes when you copy a binding and forget to change the string. And it means more work...

MoOx:
The opposite is valid: you can copy a binding & forget to update the non visible implicit string.
I thing it far more easy to forget about the empty string. 
I think Chenglou & other got a discussion about this and stopped to recommend empty string

cknitt:
In our bindings, in >99%, the `""` is correct because we use the same name on the Reason side as on the JS side.

The remaining cases are those where we provide two or more bindings for the same function, like `xxx` and `xxxWithConfig`. In those cases you need to pay close attention on implementation and review. For the remaining 99.9% it is overkill to specify the function name again IMHO.

And by using the naming convention xxx**With**Yyy, it is easy to search for these cases in the code. I used this convention in all cases in my code. I think `saveToCameraRollOverrideType` is the only case where we don't have it, maybe we should change it to `saveToCameraRollWithType`.

MoOx:
Fair enough.

`saveToCameraRollWithType` make sense then?

MoOx:
Fair enough.

saveToCameraRollWithType make sense then.

PR #465, opened 2019-04-28
Define type for onOrientationChange callback

sgny:
`onOrientationChange` takes a function defined from a synthetic event wrapping an `orientation` key. However, allowed values for the key are only `portrait` and `landscape` so the existing type is not precise enough. To work around this, I added a new type named `eventValue` and two inlined values `landscape_` and `portrait_`.

Should we add `eventValue` or allow users to define actions in the callback for values which will never occur?

### DIFF
--- a/reason-react-native/src/components/Modal.bs.js
+++ b/reason-react-native/src/components/Modal.bs.js
@@ -4,5 +4,8 @@ var NativeElement$ReactNative = require("../elements/NativeElement.bs.js");
 
 var Orientation = 0;
 
+var OrientationChange = 0;
+
 exports.Orientation = Orientation;
+exports.OrientationChange = OrientationChange;
 /* NativeElement-ReactNative Not a pure module */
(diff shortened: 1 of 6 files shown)


reason-react-native/src/types/Modal_Orientation.rei
```diff
@@ -13,4 +13,12 @@ let landscapeRight: t;
...
 
 [@bs.inline "portrait-upside-down"]
-let portraitUpsideDown: t;
\ No newline at end of file
+let portraitUpsideDown: t;
+
+type eventValue;
+
+[@bs.inline "landscape"]
+let landscape_: eventValue;
```
MoOx: I am afraid this won't be intuitive. Shouldn't we have a Change module in there or Modal_OrientationChange?

cknitt: I vote for Modal_OrientationChange.

sgny: Implemented `Modal_OrientationChange`.

PR #467, opened 2019-04-29
Update TextInput event types using Event.syntheticEvent

sgny:


cknitt:
@MoOx CI fails - could you have a look at it?

sgny:
> @MoOx CI fails - could you have a look at it?

Fixed a missing "."

PR #472, opened 2019-04-29
bs-react-native-jsx3-compat

### LINKED ISSUE #413: bs-react-native-jsx3-compat

MoOx:
To try this if you dare to

Below are basically the things you need to do to migrate your code slowly (& be able to use reason-react-native)

(Will be added as a page on the website)

---

# How to migrate to reason-react 0.7 with full jsx3

## Migrate your code to JSX3 using a script

Instead of using official upgrade path, that allow you to use jsx2 & jsx3 at the same time, which is nice but require an effort from you to make components written in one version to work with the other & vice-versa, we will follow an [alternative upgrade path](https://bloodyowl.github.io/blog/2019-04-19-an-alternative-migration-path-for-reason-react/) based on a [reason react compat hook](https://github.com/bloodyowl/reason-react-compat) coupled with our compatibility layer `bs-react-native-jsx3-compat`

So first, 

- be sure that you use `bs-platform@^5.0.1`
- be sure to use `reason-react` master branch (until 0.7.1 or 0.8 is out) (at least commit [835b90c](https://github.com/reasonml/reason-react/commit/835b90c1c73827632a9819a353a404380724b1d2))
- use `react@^16.8.0`
- add `reason-react-compat@^0.4.0`
- (if needed, upgrade some deps (eg: `reason-apollo@^0.16.0`))
- use https://github.com/bloodyowl/upgrade-reason-react
  -   **⚠ Note that if you use `->` operator a lot inside jsx, you will have a lot of bad surpise atm. We are working on improving the script.** Be warned.
  - if things breaks hard with `❗️️Errored on {somefile}`, that probably because something doesn't please the inlined version of refmt that's in the upgrade script. Things like this
  ```reason
    StyleSheet.create(
      Style.{
  ```
  will need to be replaced with
  ```reason
  Style.(StyleSheet.create({
  ```
  This also require you to adjust the other side of this StyleSheet declaration, so replace
  ```reason
      },
    );
  ```
  with
  ```reason
      },
    ));
  ```
  Same for very short declarations
  ```reason
  StyleSheet.create(Style.{
  ```
  will need to be replaced with
  ```reason
  Style.(StyleSheet.create({
  ```
  And don't forget on the other side... `});` with `}));`
  Same apply for similar patterns...
### Workaround for git usage

- `yarn add https://github.com/reasonml-community/bs-react-native.git#jsx3-migration`
- run a dirty command to get proper setup when using `bs-react-native-jsx3-compat`: add as a `prepare` script
  - `"prepare": "ln -sFf ../node_modules/bs-react-native-monorepo/reason-react-native/ ./node_modules",`
- run `yarn prepare` (this will be run each time you use yarn & will ensure that the trick is always around until you actually use released versions of this work)

Now you can try to compile your code but it's unecessarry. **You might have tons of error because `bs-react-native` is full of jsx2** & we just migrated codebase to jsx3.

So please:

- Ensure your project is full jsx3 (`bsconfig.json`: `"reason": {"react-jsx": 3},`)
- Replace `bs-react-native` by `bs-react-native-monorepo/bs-react-native-jsx3` in your `bsconfig.json` (**⚠️ check the [git monorepo trick](https://github.com/reasonml-community/bs-react-native#-usage-from-git-repo)**)

Compilation time! You should still have some errors, but probably not that much.

## Time to fix remaining errors before everything works™ 

We did our best here to avoid you having to update/fix your code but there might still be some minor stuff to adjust

- you might find some weird error related to fast-pipe swapping code, this can be related to refmt issues, you will have to fix those by hand
  - eg: `{string_of_int(value)->ReasonReact.string}` turned into `{string_of_int;value->ReasonReact.string}`
- check `children` errors and remove `let children = React.Children.toArray(children);` that might have beed added without being necessary
- you can safely remove spread `...` for `children`
- you might need to tweak place where children are optional with `{children->Belt.Option.getWithDefault(React.null)}`

### Notes about `ReasonReact.wrapJsForReason` & `ReasonReact.wrapReasonForJs`

With reason-react 0.7 & new api, code produced is just normal react component. So this things are not necessary anymore but are not handled by the script we used.

So you might need to remove those things & replace them with vanilla React code

- Remove `ReasonReact.wrapJsForReason` ([example](https://github.com/MoOx/moox.io/commit/1c38aa5bb583ff8146fc7e8fea06446d8447249a#diff-2ed4f80b953266403fd687c51270ebfa))
- Remove `ReasonReact.wrapReasonForJs` ([example](https://github.com/MoOx/moox.io/commit/1c38aa5bb583ff8146fc7e8fea06446d8447249a#diff-fe8d8d6fb5aac7a6cb2657ff9de22d89L65), [another example](https://github.com/MoOx/moox.io/commit/1c38aa5bb583ff8146fc7e8fea06446d8447249a#diff-cad3bf1049ec571062e3ad1e8c8857aaL72) )

## Compilation time

When you don't have any errors, just compile & enjoy jsx3. You can even add `reason-react-native` to your `bsconfig.json` `bs-dependencies` & use it right away with [awesome hooks](https://reasonml.github.io/reason-react/docs/en/components) 🥳

Here are some commits that might help you

- https://github.com/MoOx/moox.io/commit/1c38aa5bb583ff8146fc7e8fea06446d8447249a
- https://github.com/MoOx/dame.bio/commit/7bded4df4bf9309b7bd5ccac8f822c4fb7ff7e09

🙏 _Please do not hesitate to share your experience with the migration._ If the source is open, post your commit here, it could help someone without you even knowing 🤷‍♂️

## Need help to migrate?

If you struggle with it & have questions, [reach us on our discord server](https://discord.gg/eaU3Z6Q)

---

**If you are reading this, you can assume you are done & ready to work on other things, like feature & bugfix for your product.**

---

## But I have more time! I WANT MORE

You can do some easy stuff to cleanup your codebase now that everything is working and that you are ready to use reason-react 0.7:

- Easy replacement
  - `ReasonReact.string` => `React.string`
  - `ReasonReact.array` => `React.array`
  - `ReasonReact.null` => `React.null`
  - `ReasonReact.reactElement` => `React.element`
  - ...
- More tricky ones: you can probably remove lot of unecessary wrapping for stateless component that don't use the lifecycle
    ```reason
      ReactCompat.useRecordApi({
          ...component,
          render: _self => {
    ```
    Can be replaced with
    ```reason
    ({{
    ```
    And
    ```reason
      ReactCompat.useRecordApi({
          ...component,
          render: _self => 
    ```
    Can be replaced with
    ```reason
    ({
    ```
    At this point, the 2 replacements above have generated syntax error, so you should try to replace
    ```reason
    ,
      });
    ```
    with
    ```reason
    });
    ```
    At this point, you should be able to drop some `let component = ReasonReact.statelessComponent...` too, since some won't be used anymore thanks to the last removal step we did.

Now rely on refmt to get a cleaner code and you are good to go!

Do not hesitate to share more tips & tricks to ease the migration! Don't be shy!

---

Closes #413


cknitt:
Wow, great work @MoOx! I will test this with a project of ours when I have time later this week.

sgny:
It is a huge undertaking - to review - let alone to actually implement. Great job!

I have an old project I could test, but I'll also probably wait for the weekend.

MoOx:
Yeah just wait I waste my time on trying some migrations & I will tell if this actually work or not (at least for most parts) :D 

What I would advise is to only review *.rei tho

MoOx:
First test has been successful. Small project (simple website) but basic things as View, Text, Style, Image etc works like a charm.
There is still a manual trick to do for git usage.
I added notes in the PR description if anyone wants to try this.

MoOx:
I have edited description. Should be ready for testing now.
I will try others projets in the coming days.

Do not hesitate to share details if you think that things are missing from the migration guide.

fakenickels:
Amazing work here @MoOx! I am doing a huge migration and will try out with this branch now

bs-react-native-jsx3-compat/src/components/touchableWithoutFeedback.re
```diff
@@ -22,43 +20,34 @@ let make =
       ~onPressIn=?,
       ~onPressOut=?,
       ~pressRetentionOffset=?,
-      ~style=?,
```
fakenickels: why remove `style`?

MoOx: It doesn't exist for this component https://github.com/facebook/react-native/blob/4900077e83ec7597b2ace3c3882722e752ae64e6/Libraries/Components/Touchable/TouchableWithoutFeedback.js#L56-L81
TouchableWithoutFeedback is cloning its only child so you can pass style there https://github.com/facebook/react-native/blob/4900077e83ec7597b2ace3c3882722e752ae64e6/Libraries/Components/Touchable/TouchableWithoutFeedback.js#L254

fakenickels: aah got it! I wasn't aware of that, thanks!

MoOx: Me neither (before implementing that) haha

fakenickels: haha

fakenickels:
I ran all the steps, and just had some problems with upgrade-reason-react children and other stuff (and ofc the bound components I had that upgrade-reason-react don't covered). 

Besides that I had problem with only the hitSlop param of TouchableOpacity which changed from the Js.t obj to the bs.obj bound function and the TouchableWithoutFeedback style param that indeed doesn't exist

fakenickels:
As the project is really big (220 Reason files) I couldn't get it running just yet before I fix the other problems, but most of the compiler errors are due to these other problems and not to the bs-react-native-jsx3-compat itself

MoOx:
> I ran all the steps, and just had some problems with upgrade-reason-react children and other stuff (and ofc the bound components I had that upgrade-reason-react don't covered).

@bloodyowl is trying to improve the upgrade script, we can have some hope to improve the situation... We will see.

> Besides that I had problem with only the hitSlop param of TouchableOpacity which changed from the Js.t obj to the bs.obj bound function and the TouchableWithoutFeedback style param that indeed doesn't exist

Do you think we can improve the situation from the bindings?

> As the project is really big (220 Reason files) I couldn't get it running just yet before I fix the other problems, but most of the compiler errors are due to these other problems and not to the bs-react-native-jsx3-compat itself

Ok no rush. Feel free to give more in deep details later if needed.

Thanks for trying this :)

I am trying it on another project, and just like you said, most issue comes from some fails with the upgrade script. Beside that, things look good if we stick to evaluate the compat layer.

fakenickels:
>@bloodyowl is trying to improve the upgrade script, we can have some hope to improve the situation... We will see.
that would be something nice to see, tho compared with the rest of the code the bindings are really a few files (10 files)

>Do you think we can improve the situation from the bindings?

I think as this is a compat would be cool to keep the old type also for hitSlop

cknitt:
Great instructions! Started migrating a big project.

One thing I noticed is that the upgrade script eats labels from function calls, e.g.

```re
let getLocaleWithoutCountry = locale => locale->Js.String.substring(~from=0, ~to_=2);
```

becomes

```re
let getLocaleWithoutCountry = locale => locale->Js.String.substring(0, 2);
```

MoOx:
I am going to merge this after putting my notes above as a migration guide on the docs.
We can still fix things if necessary & make some adjustments.

fakenickels:
how are you guys planning in releasing this? will it be a npm package called bs-react-native-jsx3-compat?

MoOx:
Yes it will.

czystyl:
I migrated my project as well and everything works great! 

Good job @MoOx 👏👏👏 🚀

bs-react-native-jsx3-compat/src/components/button.re
```diff
@@ -9,14 +8,11 @@ let make =
...
-      "onPress": onPress,
-      "testID": testID,
-      "title": title,
-    },
-  );
+  <ReactNative.Button
+    ?accessibilityLabel
+    ?color
+    ?disabled
+    onPress={_ => onPress()}
```
czystyl: Shouldn't be `onPress={onPress)}`? To pass everything that comes from RN?

MoOx: Nope because current bindings doesn't tell you there is an even as an argument. That's why I used this kind of tricks https://github.com/reasonml-community/bs-react-native/pull/472/files#diff-e9317c6473601bfdc17cac2811383838R128

czystyl: Ok, make sense! Thanks for the explanation.

fakenickels:
👏 👏 👏

PR #473, opened 2019-04-29
fix(bs-react-native): fix path name

### LINKED ISSUE #469: Installing using git url
As per the documentation it says to use `bs-react-native` namespace to refer to the git installation
```
{
  // ...
  "bs-dependencies": [
    // BsReactNative, current "stable" bindings
    "bs-react-native/bs-react-native",
    // ReactNative, WIP on zero-cost bindings (require bs-platform 5.0.0 & and reason-react hooks branch)
    "bs-react-native/reason-react-native"
  },
  // When used from git with the following path, bs-platform will use path that won't be working
  // (because deps are pointing to folder in folder)`
  // Your bundler (metro or webpack) will say that it can find `bs-react-native/whatever` and/or
  // `reason-react-native/whatever`
  // which is correct because it should be `bs-react-native/bs-react-native/whatever` / `bs-react-native/reason-react-native/whatever`
  // the trick below will correct path in generated JavaScript files
  // and will just need to be removed when you won't use git directly anymore
  "js-post-build": {
    "cmd": "./node_modules/bs-react-native/git-monorepo-usage-trick"
  }
}
```

where when installing with yarn the destination is actually `node_modules/bs-react-native-monorepo` so I gotta do a `mv` to rename properly. I'm not sure if I'm doing something wrong or the docs are misleading.

Let me know if I can send a PR for this!

fakenickels:
Fixes #469

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,9 +20,9 @@ If you use this bindings from the git repo, be sure to use
   // ...
   "bs-dependencies": [
     // BsReactNative, current "stable" bindings
-    "bs-react-native/bs-react-native",
+    "bs-react-native-monorepo/bs-react-native",
     // ReactNative, WIP on zero-cost bindings (require bs-platform 5.0.0 & and reason-react hooks branch)
-    "bs-react-native/reason-react-native"
+    "bs-react-native-monorepo/reason-react-native"
   },
   // When used from git with the following path, bs-platform will use path that won't be working
   // (because deps are pointing to folder in folder)`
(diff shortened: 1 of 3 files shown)


fakenickels:
Testing here before turning clicking in `Ready for review`

MoOx:
Maybe we should instead adjust the git-monorepo-trick to make it clearer?

fakenickels:
This one worked fine, but I agree with you @MoOx. Maybe it's less confusing keeping the `-monorepo`

fakenickels:
I think this is even better because it allows people to have both the monorepo and the npm package at once

fakenickels:
I am testing it now

Issue #450, opened 2019-04-26
Directly allow strings for margin and size, too?


cknitt:
Should we use the approach for color props from #448 for margin and size props, too? I.e. have modules `Margin` and `Size` with "safe" helper functions, but `Margin.t` = `Size.t` = `string`?

Would make things easier for people coming from React Native JS.

@MoOx @sgny

MoOx:
I am more than ok with this change.

cknitt:
Hmmm, compared to `Color`, the problem with margin and size is that these accept either floats or strings.

MoOx:
This is because in JS "1" == 1...
Should we accept string only for unsafe usage & have method sending float as unitless value (pt) or string for %?

DCKT:
I think it's a good idea too, we should find the good compromise 🤔 
We have to deal with `Pt()`, `Float()`, `Float` this is pretty confusing for beginners

Actually : 
```
fontSize(Float(12.))
margin(Pt(12.))
borderRadius(4.)
```
String should be an escape hatch I think

MoOx:
Float won't be a thing in new bindings, that said, we still try to add a safety layer to avoid accepting anything that won't work anyway...

MoOx:
was wondering about a thing that bother me a little (but not too much :D ).  We recently moved some submodules from Style, like Margin, Size etc, but I was wondering if that's a good idea to have so much "generic names" taken as soon as you `open ReactNative`.

I remember in the past being unhappy because bs-rn had generic names like "Types" taken by the open... And I just saw that in some projects (some personals & some for clients), I have a Color module, as well as a Font one & sometimes a Style one too.
So I was asking myself if we shouldn't keep Margin & Friend in Style, even if that's require 2 () when you want to use those, to avoid this kind if "issues".

Thoughts?

cknitt:
@MoOx I see your point. On the other hand, that's what happens when you open modules. 😉 I am currently working on a GIS application where I have a `Map` module, so when I `open Belt` I get a warning because of `Belt.Map`...

Anyway, I am not opposed to moving `Margin` and `Size` back into `Style` because they are only referenced from `Style`. We should have `types/Style_Margin.re` and `types/Style_Size.re` then and alias these in `Style`.

`Color` should remain outside of `Style` though as it is used outside of `Style` (directly in the props of many different components).

DCKT:
+1 with @cknitt ; `Margin` & `Size` should be part of `Style` 😃

[...]

PR #478, opened 2019-05-01
Remove Margin and Size modules, add size type in Style module.

cknitt:
@MoOx Following our discussion on Discord and in #450, here is an attempt at simplifying things.

### DIFF
--- a/reason-react-native/src/apis/Style.rei
+++ b/reason-react-native/src/apis/Style.rei
@@ -1,10 +1,3 @@
----
-id: apis/Style
-title: Style
-wip: true
----
-
-```reason
 type size;
 
 external pt: float => size = "%identity";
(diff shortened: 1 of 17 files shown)


reason-react-native/src/apis/Style.rei
```diff
@@ -0,0 +1,242 @@
+type size;
+
+external pt: float => size = "%identity";
+
+let pct: float => size;
+
+type margin = size;
+
+[@bs.inline "auto"]
+let auto: margin;
```
MoOx: neat trick

website/src/components/HeaderLarge.re
```diff
@@ -123,7 +123,7 @@ let make = (~currentLocation) => {
                accessibilityLabel={item.text}>
                {item.componentFunc(
                   ~iconColor=Consts.Colors.light,
-                  ~iconSize=22.->Svg.Size.pt,
+                  ~iconSize=22.->Svg.pt,
```
MoOx: Are you sure about this? I don't see Svg module being updated.

cknitt: Aargh, search and replace error.

PR #479, opened 2019-05-01
LayoutAnimation API

sgny:


reason-react-native/src/apis/LayoutAnimation.re
```diff
@@ -0,0 +1,67 @@
...
+    ~create: animationConfig=?,
+    ~update: animationConfig=?,
+    ~delete: animationConfig=?,
+    unit
+  ) =>
+  layoutAnimationConfig =
+  "";
+
+[@bs.module "react-native"] [@bs.scope "LayoutAnimation"]
+external configureNext:
```
cknitt: Maybe we should have two bindings `configureNext` and `configureNextWithEndCallback`?

sgny: Further to your comment elsewhere regarding searching for such sister functions, perhaps we should simply add comments indicating that we have a convenience function whenever we do something like this?

sgny:
I added `Presets` so that presets like `linear` or `spring` may be passed as argument to `configureNext` and `configureNextWithCallback`. Basically,
```
LayoutAnimation.configureNext(LayoutAnimation.Presets.linear)
```
is the same as
```
LayoutAnimation.linear()
```
but they may nevertheless prove useful when making use of a callback.

Presets is actually a dictionary, so we could bind to `Presets` instead and define getters with `[@bs.get]` but this PR allows BS API to be mostly identical to the JS API.

reason-react-native/src/apis/LayoutAnimation.re
cknitt: Yes, why not, I'd be happy to approve a (separate) PR adding such comments to all the xxxWithYYY functions. 👍

Issue #239, opened 2018-09-17
Platform.os binding prevents inlining/optimization


grabbou:
Metro bundler inlines `(ReactNative.)Platform.OS` and `(ReactNative.)Platform.select`, so that when packaging for one platform, the code for the other platform is optimized away.

So if I had

```reason
[@bs.module "react-native"] [@bs.scope "Platform"] external os : string = "OS";

if (os === "ios") {
  Js.log("ios");
} else {
  Js.log("android");
};
```

this would get compiled by BuckleScript to

```js
if (ReactNative.Platform.OS === "ios") {
  console.log("ios");
} else {
  console.log("android");
}
```

and Metro Bundler would inline ReactNative.Platform.OS and subsequently optimize the Android branch away for an iOS build.

However, with the current Rebolt bindings, I have

```reason
switch (Platform.os()) {
| IOS(_) => Js.log("ios")
| Android => Js.log("android")
};
```

which gets compiled by BuckleScript to

```js
var match = Platform$Rebolt.os(/* () */0);
if (match) {
  console.log("ios");
  return /* () */0;
} else {
  console.log("android");
  return /* () */0;
}
```

This prevents the inlining/optimization from working.

grabbou:
Cross-posted from https://github.com/callstackincubator/rebolt/issues/66 via @cknitt

cknitt:
I created this issue originally in the rebolt repo.

The new bindings currently being worked on (bs-react-native-next / #344) won't have that problem. (At the moment, they only directly bind `Platform.os` as a string.)

PR #481, opened 2019-05-04
template using reason-react-native

dawee:
Here's a first bit for the template.

I removed `bs-platform` from dependencies because when it's installed through react-native init, `bsb.exe` is not generated (and I have no idea why).
I need to add an instruction to install it manually.

Also I have a warning on the platform switch. Now that `Platform.os` is not a function anymore, it's resolved at compilation, so Bucklescript seems to say that pattern matching is unused.

As I said in Discord, it's hard to have way better than that because the current template system doesn't allow much. I'll be able to push something way more "plug and play" when React Native 0.60.0 will be released.

To test the template:

1) Checkout this branch
2) Run `react-native init MyProject --template file:///absolute/path/to/bs-react-native/react-native-template-reason
3) Run `cd MyProject && yarn add -D bs-platform`
4) In a terminal: `yarn bsb -make-world -w`
6) In another terminal: `react-native run-{android|ios}`


react-native-template-reason/src/App.re
```diff
@@ -4,55 +4,50 @@
...
+    ~textAlign=`center,
+    ~margin=pt(10.)
+  );
+
+  let instructions = style(
+    ~textAlign=`center,
+    ~color="#333333",
+    ~marginBottom=pt(5.)
+  );
+};
```
MoOx: Can you revert this to an actual StyleSheet? It's better for performance so we should definitely not promote this kind of modules for styling. Imo it's a bad practise I see too often in bs-rn world.

react-native-template-reason/src/App.re
```diff
@@ -4,55 +4,50 @@
...
+
+[@react.component]
+let app = () => {
+  let instructions =
+    switch(Platform.os) {
+    | ios => "Press Cmd+R to reload,\n" ++ "Cmd+D or shake for dev menu"
+    | android =>
+      "Double tap R on your keyboard to reload,\n"
+      ++ "Shake or press menu button for dev menu"
+    };
```
MoOx: Can we keep example for other platform please (at least web)?

react-native-template-reason/bsconfig.json
```diff
@@ -15,5 +15,11 @@
...
     }
   ],
-  "bs-dependencies": ["bs-react-native", "reason-react"]
+  "bs-dependencies": [
+    "reason-react",
+    "bs-react-native/reason-react-native"
+  ],
+  "js-post-build": {
+    "cmd": "./node_modules/bs-react-native/git-monorepo-usage-trick"
+  }
```
MoOx: I think we should make some change here so it's work on CI.
And provide tweaks in the README or something.

react-native-template-reason/src/App.re
dawee: Sure I can put Stylesheet instead. When we started to work with bs-react-native we hesitated between the 2 versions. We didn't use Stylesheet because we didn't understood how it was beneficial. Can you explain to me this perfomance issue you're talking about?

react-native-template-reason/src/App.re
dawee: Yes I can add that. BTW why the choice of removing variants? I would have expect lowercase polymorphic variants as in the other new modules. Instead we have directly a string type as in JS which means we can write a pattern matching with random string and it would compile. Isn't it a regression compared to what `bs-react-native` did?

react-native-template-reason/bsconfig.json
dawee: What do you mean by "some changes"? The post build scripts won't work on CI?

react-native-template-reason/src/App.re
MoOx: When you use a module like that, you just send raw styles. StyleSheet on the other hand prepare styles & create internal identifiers (that's why StyleSheet elements are `int` in JS). Optimization can be made from there in native part (I am not 100% sure that's the case, I am guessing here).

The fact that StyleSheet.flatten have an explicit cost (see https://facebook.github.io/react-native/docs/stylesheet#flatten) make me think that an optimisation is made. Also if you have some redundant styles, a unique id will be used (and this is very likely to be the case for reusable components).

For react-native-web I am sure there is a direct benefit when using StyleSheet, so again, pretty sure (but I never checked - to be honest) that's the same for ios & android implementation.

dawee: I didn't know any of this. Thank you for the explanation.

react-native-template-reason/bsconfig.json
MoOx: node_modules/bs-react-native doesn't exist as that's the repo itself. If you are not confortable on handling the situation (which is totally understandable) we can imaging merging your work in another branch than master & continue working from here.

dawee: The git+https dependency I've added in package.json doesn't work on CI? It works on my Linux machine.

react-native-template-reason/src/App.re
MoOx: First take a look at this https://github.com/reasonml-community/bs-react-native/issues/239

Then this is simply to get a zero-cost binding. Using variant create some unecessary js. Polymorphic variant are not an option considering the fact that we cannot use bs.string in this kind of scenario (if I am correct).
So we used bs.inline (requested for this kind of purpose, see https://github.com/BuckleScript/bucklescript/issues/3472)

react-native-template-reason/bsconfig.json
MoOx: Problem is: by doing that, you refer to master branch. If we make a breaking change to something & update the template at the same time, this technique will just not work at all.
And btw, this monorepo trick will never be released so shouldn't be in the template.

Can't you just use `reason-react-native` as a dep & see if that works?

dawee: Apparently `reason-react-native` is not published on NPM yet. Shouldn't we wait for it to be published before using it?

MoOx: We are using yarn workspace here. So all packages of this repo are simlinked, so those are available as soon as this is run locally. The website package is doing that (master shows "../reason-react-native" but jsx3-migration branch does have `"reason-react-native`") & it's working :)

react-native-template-reason/src/App.re
dawee: OK, I see.

react-native-template-reason/src/App.re
```diff
@@ -4,55 +4,48 @@
...
-  | IOS(_) => "Press Cmd+R to reload,\n" ++ "Cmd+D or shake for dev menu"
-  | Android =>
+  switch (Platform.os) {
+  | ios => "Press Cmd+R to reload,\n" ++ "Cmd+D or shake for dev menu"
+  | android =>
     "Double tap R on your keyboard to reload,\n"
     ++ "Shake or press menu button for dev menu"
-  | exception (Platform.UnknownPlatform(p)) when p == "web" => "Press Cmd+R to reload."
-  | exception (Platform.UnknownPlatform(_p)) => ""
+  | web => "Press Cmd+R to reload."
```
cknitt: This won't work.

You will need to do something like

```
switch (Platform.os) {
  | os when os === Platform.ios => ...
  | os when os === Platform.android => ...
  | os when os === Platform.web => ...
  | _ => ...
}
```

(or use `if`/`else`)

Less convenient than with the old bindings, but that's because of #239.

MoOx: Indeed that's very verbose. 
See https://reasonml-community.github.io/bs-react-native/en/docs/apis/Platform/

dawee: Fixed!

react-native-template-reason/src/App.re
```diff
@@ -4,55 +4,49 @@
...
 let app = () =>
-  <View
-    style={
-      styles##container;
-    }>
-    <Text style={styles##welcome}>
-      "Welcome to (Bs) React Native!"->ReasonReact.string
+  <View style=styles##container>
+    <Text style=styles##welcome>
+      "Welcome to (Bs) React Native!"->React.string
```
cknitt: (Reason) React Native 😉

react-native-template-reason/src/App.re
```diff
@@ -4,55 +4,49 @@
...
-  | Android =>
+  switch (Platform.os) {
+  | os when os == Platform.ios =>
+    "Press Cmd+R to reload,\n" ++ "Cmd+D or shake for dev menu"
+  | os when os == Platform.android =>
     "Double tap R on your keyboard to reload,\n"
     ++ "Shake or press menu button for dev menu"
-  | exception (Platform.UnknownPlatform(p)) when p == "web" => "Press Cmd+R to reload."
-  | exception (Platform.UnknownPlatform(_p)) => ""
+  | os when os == Platform.web => "Press Cmd+R to reload."
```
cknitt: Won't this give a warning because the pattern matching might not be exhaustive?

You can add

```
_ => ""
```

at the end then.

react-native-template-reason/src/App.re
MoOx: +1

react-native-template-reason/src/App.re
MoOx: +1

MoOx:
Thanks for you work @dawee 🙏

dawee:
> Thanks for you work @dawee 🙏

No problem! It was a good opportunity for me to learn the new API. I'll be pushing another PR for this template when react-native@0.60.0 come out! Full automated, "re:*" scripts pre-written and full Reason. It's gonna be beautiful :D.

PR #484, opened 2019-05-07
Add documentation and example for LayoutAnimation

sgny:
Since `LayoutAnimation` is not widely known, I figured it would be useful to have its documentation.

### DIFF
--- a/reason-react-native/src/apis/LayoutAnimation.md
+++ b/reason-react-native/src/apis/LayoutAnimation.md
@@ -1,13 +1,62 @@
 ---
 id: apis/LayoutAnimation
 title: LayoutAnimation
-wip: true
+wip: false
 ---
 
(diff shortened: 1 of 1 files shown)


reason-react-native/src/apis/LayoutAnimation.md
```diff
@@ -1,90 +1,168 @@
 ---
 id: apis/LayoutAnimation
 title: LayoutAnimation
-wip: true
+wip: false
```
MoOx: you can just drop the flag next time :)

reason-react-native/src/apis/LayoutAnimation.md
```diff
@@ -1,90 +1,168 @@
...
+              0. -. windowWidth;
+            },
+          ),
+        ~flexDirection=`row,
+        (),
+      )}>
+      <View style={Style.style(~width=Style.pt(windowWidth), ())}>
+        <Register />
+      </View>
+      <View style={Style.style(~width=Style.pt(windowWidth), ())}>
```
MoOx: I would have used a StyleSheet but we can improve this detail later.

PR #485, opened 2019-05-07
CameraRoll documentation and example

sgny:
The example demonstrates both `saveToCameraRoll` and `getPhotos` but has perhaps become a bit too complex.

Accordingly, it might be useful to decide upon goals for examples in documentation. Should we have code that will only demonstrate the essential bits, or something more complete which will run when copied and pasted?

### DIFF
--- a/reason-react-native/src/apis/CameraRoll.md
+++ b/reason-react-native/src/apis/CameraRoll.md
@@ -1,14 +1,49 @@
 ---
 id: apis/CameraRoll
 title: CameraRoll
-wip: true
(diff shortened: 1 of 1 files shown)


MoOx:
> Accordingly, it might be useful to decide upon goals for examples in documentation. Should we have code that will only demonstrate the essential bits, or something more complete which will run when copied and pasted?

I think complete example are cool, but if we can do short examples & longer version that's even better. I liked what you did on LayoutAnimation

PR #487, opened 2019-05-08
Revise CameraRoll documentation

sgny:
Moved styles to a StyleSheet and created some helper functions to move non-essential elements out of the `make` statement.


reason-react-native/src/apis/CameraRoll.md
```diff
@@ -41,6 +40,8 @@ The function will return a `photoIdentifiersPage` object wrapped in a Promise, t
 ## Types
 
 - `getPhotosParams` can be constructed with its eponymous constructor
```
cknitt: Had to look up "eponymous" in the dictionary. 😂 Maybe use simpler wording for non-native speakers?

"with its constructor of the same name"?

reason-react-native/src/apis/CameraRoll.md
```diff
@@ -130,7 +131,95 @@ type action =
...
+    "container":
+      Style.style(
+        ~flex=1.,
+        ~flexDirection=`column,
+        ~alignItems=`center,
+        ~justifyContent=`spaceBetween,
+        (),
+      ),
+    "getPhotosExample":
+      Style.style(
```
cknitt: I'd open Style before StyleSheet.create:

```re
let styles =
	  Style.(StyleSheet.create({
	    "container":
	     style(
	     ...
```

Then you can just write `style(...)` and `pt(...)` without the `Style.`.

sgny: Good suggestion, I'm familiar with local opens, but don't use them often. Will update the example to take advantage of that.

reason-react-native/src/apis/CameraRoll.md
sgny: It was fun to use though :)

reason-react-native/src/apis/CameraRoll.md
```diff
@@ -130,7 +131,96 @@ type action =
...
+
+let styledText = s => {
+  <Text style=styles##text> s->React.string </Text>;
+};
+
+let thumbnails = photos => {
+  <View style={Style.style(~height=Style.pt(0.15 *. windowHeight), ())}>
+    <ScrollView horizontal=true>
+      {React.array(
+         Array.mapi(
```
cknitt: Small thing, but in the spirit of zero I'd rather use `Js.Array.mapi`.

sgny: Thanks!

MoOx: I would ever go mo aggresive using `Belt.Array.mapWithIndex` with fastpipe. It seems that new comers (at least the one I know) find this more natural

```reason
photos
->Belt.Array.mapWithIndex((i, s) => <Image /* .... */ />)
->React.array
```

cknitt: That's fine, too, and it's what we are doing in our projects (we are using Belt exclusively). If you are not using Belt, it brings an additional dependency into your project though.

BTW, there is now a pipe-first-compatible Js.Array2, too.

sgny: I wasn't too sure if using pipes would be confusing for newcomers (OK, it's already used here with `React.string`, but that's very simple), but it's easier to read for sure.

I think it's better to stick with Js.Array which works with `|>`) or Js.Array2 (I wasn't aware of this) if `->` is more familiar in general. Either simply maps JS native `map` method of the array. 

I like to use `Belt` when it's useful (no out of bound errors when getting items from arrays, for example), but I don't see what `mapWithIndex` has over `mapi`.

So, `->` or `|>`?

cknitt: I am leaning towards `->`, but it's not a big deal, I'll approve the PR as it is now.

sgny: I probably would not have noticed `Js.Array2` for at least a bit longer had it not been for your comment. So I'll use `Js.Array2` to draw attention to that module.

MoOx: not sure reasonml/buckelscript teams want to promote *2 modules...
I would definitely go with `->` & Belt.
Learning Belt is a better idea, having native compilation in mind :)

cknitt: Yeah, the *2 modules are hopefully only a temporary solution until everything is pipe first.

Ok, so let's go with Belt + pipe first!

Issue #486, opened 2019-05-08
Remove Types module


cknitt:
@MoOx I think we should remove the `Types` module before a release. It only contains two abstract types, `edgeInsets` and `point`, and there is a high chance it will conflict with a `Types` module in your project if you do `open ReactNative;`.

* `point` is only used for the `contentOffset` prop in `ScrollView` + subclasses and could therefore be declared in the `ScrollView` module. (Maybe it should be renamed to `contentOffset`, too?)
* `edgeInsets` is only used for some props in `View` + subclasses and could therefore be declared in the `View` module.

Alternatively, if we don't go that way, we could rename `Types` to `ViewPropTypes` instead.

MoOx:
I totally dislike generic name indeed.
Both solutions works for me, I would go to have something that require less maintenance & that's more obvious.
Moving things into proper module make sense. Not sure `ViewPropTypes` is a good idea as this will not match RN `ViewPropTypes` at all.

cknitt:
Ok, so I'll prepare a PR for `View.edgeInsets` and `ScrollView.contentOffset` later today or tomorrow. I think these are more obvious?

wokalski:
I agree @cknitt

MoOx:
Yeah looks good

sgny:
I won't erase the above comment, but upon further testing in the simulator, everything seems to work fine.

I tested the build locally, and posted when that went without errors. Then, I wanted to make sure that the later definition did not in fact replace the earlier one. So, I tried that bit out on `Try Reason` (because I did not want to start the simulator) and it didn't work at all and I commented as above.

sgny:
Upon further testing, my suggestion only seems to have issues on `Try Reason` because of an already known issue (see the note on `foo##bar`); it's perfectly fine as tested on the simulator.

sgny:
I made the mistake of testing on `Try Reason` as I forgot the issue with `##` accessors. First I posted a followup, that my code did not actually work, then posted again saying it does after I tested on the simulator.

For some reason, my later posts did not show up (apparently my first post is timestamped five hours *in the future*). Sorry about the duplicate posts (now deleted).

cknitt:
In keeping with https://github.com/reasonml-community/bs-react-native/blob/master/reason-react-native/CONTRIBUTING.md, I'd rather keep different types (abstract types for View props, Js.t types for events/callback params).

I have created PR #488 with the changes I suggested.

cknitt:
[...]

PR #488, opened 2019-05-08
Remove types module.

cknitt:
See the discussion in #486.


reason-react-native/src/apis/Event.md
```diff
@@ -112,10 +112,10 @@ type pressEventPayload = {
 
 type pressEvent = responderSyntheticEvent(pressEventPayload);
 
-type position = {
+type contentOffset = {
   .
-  "y": float,
   "x": float,
+  "y": float,
```
cknitt: Don't know why y came before x?

sgny: I followed the order in RN codebase, so there's no good reason for it.

sgny:
All seems in order to me, but I'm still not sure if we should always separate creating objects and binding to objects. I agree that some objects can only be created whereas others only consumed, but we should agree on what the norm should be when both is possible.

PR #490, opened 2019-05-09
Functorize Animated Value and ValueXY

sgny:
Since `Animated.Value` and `Animated.ValueXY` were nearly identical, I created a functor (imaginatively named `ValueMethods`!) and defined the two modules based off that.

`valueXY` was an abstract type, but given the two methods `getX` and `getY` were simply getters for the `x` and `y` keys of an object, I figured defining
```
type t = {
      .
      "x": Value.t,
      "y": Value.t,
    };
```
for the `ValueXY` module would allow us to remove those methods and use `##` instead.


cknitt:
I haven't looked into Animated too much yet, but I am usually in favor of abstract types. Maybe @MoOx can comment here, he uses Animated more intensively AFAIK.

sgny:
@cknitt Animated Value and ValueXY are essentially a number and `{x: number, y: number}`, respectively, along with associated methods.

Defining the type does not really obscure any relevant details.

PR #491, opened 2019-05-09
Add sendIntent to Linking

sgny:
I expected this to arrive in RN 0.60, but it's already in 0.59.8.

As previously discussed, I recommended and @cknitt agreed that it would be useful to add comments when we have multiple bindings to the same external function. With this PR, I'd also like to get your feedback on how the comment should read, before I propagate the change across the codebase.

I came up with `Convenience function` or `Duplicate` or `Duplicate Binding`.

@cknitt @MoOx any preferences?

### DIFF
--- a/reason-react-native/src/apis/Linking.md
+++ b/reason-react-native/src/apis/Linking.md
@@ -5,6 +5,10 @@ wip: true
 ---
 
 ```reason
+type extra;
+
+[@bs.obj] external extra: (~key: string, ~value: 'a) => extra = "";
+
 [@bs.scope "Linking"] [@bs.module "react-native"]
 external openURL: string => Js.Promise.t(unit) = "";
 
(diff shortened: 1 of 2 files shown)


cknitt:
I think convenience function is not really correct. Because without the second binding, you would not be able to call the function with a second parameter (unless you change the first binding of course).

Maybe "overloaded function" or "polymorphic function" like in https://bucklescript.github.io/docs/en/function#modeling-polymorphic-function?

sgny:
> I think convenience function is not really correct. Because without the second binding, you would not be able to call the function with a second parameter (unless you change the first binding of course).
> 
> Maybe "overloaded function" or "polymorphic function" like in https://bucklescript.github.io/docs/en/function#modeling-polymorphic-function?

I was thinking we have multiple external definitions for convenience (no need for unit, etc.), but sure it's not a convenience function. But then, this is not really function overloading or polymorphism either (at least not on the BS side). How about *multiple externals* as in the documentation? or *related externals*

MoOx:
"multiple externals" make sense to me. People will be like "what's an external??!" but they will learn & boom everyone wins.

MoOx:
We could have a faq with "what's an external" or something like that

cknitt:
"Multiple externals" is fine with me.

sgny:
Since I apparently forget to specify `sendIntent` as the name of the external... :smiley

Anyway, I added the comment
`// multiple bindings to external`
just before each binding to `sendIntent`.

Is that any clearer than `multiple external`?

MoOx:
The fact that the code is already bindings make using "bindings" a bit verbose. Small detail maybe.
I think "multiple externals" is enough.
Otherwise, new idea: we could call them `alternative externals`.

cknitt:
I'd prefer either "multiple externals" or "alternative external(s)", too.

sgny:
Since `multiple externals` comes from `BuckleScript` documentation, I went with that.

Issue #483, opened 2019-05-07
UIManager


sgny:
On Android, using LayoutAnimation requires enabling it as below: 
```
UIManager.setLayoutAnimationEnabledExperimental && UIManager.setLayoutAnimationEnabledExperimental(true);
```
This seems rather unusual, but the method is meant to be removed once `LayoutAnimation` is no longer experimental on Android. To avoid runtime errors, it is prudent to check for its presence beforehand, hence the method is called wrapped in a conditional.

While exploring this, I noticed some undocumented methods (such as popups!) for which I have started defining bindings. I will complete what I can and create a PR.

MoOx:
Closed by #496

PR #496, opened 2019-05-13
UIManager API

sgny:
Relates to #483 

This is an incomplete implementation of methods in `UIManager` but enabling `LayoutAnimation` and showing and dismissing popups on Android are included.

`setJSResponder` and `clearJSResponder` are for finer grained control of which component should be the gesture responder (such as for creating a swipable element in a `ScrollView`).

Some of the other methods have already been covered elsewhere - such as `measure` in `NativeMethods` and therefore omitted, but I will revisit this if I'm able to document other methods.

### DIFF
--- a/reason-react-native/src/apis/UIManager.bs.js
+++ b/reason-react-native/src/apis/UIManager.bs.js
@@ -1,0 +1,1 @@
+/* This output is empty. Its source's type definitions, externals and/or unused code got optimized away. */
--- a/reason-react-native/src/apis/UIManager.md
+++ b/reason-react-native/src/apis/UIManager.md
@@ -1,0 +1,48 @@
+---
+id: apis/UIManager
+title: UIManager
+wip: true
+---
+
+```reason
+type popupAction;
+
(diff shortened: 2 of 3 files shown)


reason-react-native/src/apis/UIManager.rei
```diff
@@ -0,0 +1,29 @@
+type popupAction;
```
cknitt: If we make this an abstract type, I think we should also expose the corresponding constants (UIManagerModuleConstants.ACTION_ITEM_SELECTED and UIManagerModuleConstants.ACTION_DISMISSED on the Java side). Not sure if/where they are exported to JS though.

sgny: They were exposed as `PopupMenu= { dismissed: 'dismissed', itemSelected: 'itemSelected' }`.

sgny:
> 👍

Thanks. I'll postpone merging to see if you or @MoOx have further comments.

sgny:
> Did refmt allowed this long lines?

No 😞

PR #497, opened 2019-05-13
Add comments for multiple externals and remove unnecessary names

sgny:
I removed all unnecessary names in `external` definitions and added comments for `multiple externals` where appropriate.

### DIFF
--- a/reason-react-native/src/apis/AsyncStorage.md
+++ b/reason-react-native/src/apis/AsyncStorage.md
@@ -8,27 +8,22 @@ wip: true
 type error;
 
 [@bs.scope "AsyncStorage"] [@bs.module "react-native"]
-external getItem: string => Js.Promise.t(Js.Null.t(string)) = "getItem";
-let getItem = key =>
-  getItem(key) |> Js.Promise.then_(PromiseUtils.nullableToOption);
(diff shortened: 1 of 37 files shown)


reason-react-native/src/apis/AsyncStorage.md
```diff
@@ -8,47 +8,39 @@ wip: true
...
 [@bs.scope "AsyncStorage"] [@bs.module "react-native"]
 external flushGetRequests: unit => unit = "";
 
 [@bs.scope "AsyncStorage"] [@bs.module "react-native"]
 external multiGet:
   array(string) => Js.Promise.t(Js.Null.t(array(array(string)))) =
-  "multiGet";
-let multiGet = keys =>
-  multiGet(keys) |> Js.Promise.then_(PromiseUtils.nullableToOption);
+  "";
```
cknitt: Ah, I wasn't aware we were still using the PromiseUtils module. Could you remove it, too?

sgny: `PromiseUtils` is long gone, but this doc seems to have become `async`

cknitt: Ah, ok! 😉

PR #498, opened 2019-05-14
Revise PermissionsAndroid API

sgny:
I noticed that the return types for the methods should simply be wrapped in `Js.Promise.t` and could not be `null`. I also corrected a few other issues such as `requestMultiple` being distinct from `request`.

Instead of using `[@bs.inline]` to generate strings, I defined bindings to the exposed constants.

As `Js.Dict` requires keys to be `string`, to avoid exposing that `Permission.t` is a `string`, I defined
```
type dict = Js.Dict.t(Result.t);
[@bs.get_index] external get: (dict, Permission.t) => Result.t = "";
``` 
The latter is similar to the definition of `Js.Dict.unsafeGet` in the standard `BuckleScript` library. However, as all the keys are predetermined, we should not have any issues with the function returning a `null` or `undefined`.

### DIFF
--- a/reason-react-native/src/apis/PermissionsAndroid.bs.js
+++ b/reason-react-native/src/apis/PermissionsAndroid.bs.js
@@ -3,5 +3,8 @@
 
 var Permission = /* module */[];
 
+var Result = /* module */[];
+
 exports.Permission = Permission;
+exports.Result = Result;
 /* No side effect */
--- a/reason-react-native/src/apis/PermissionsAndroid.md
+++ b/reason-react-native/src/apis/PermissionsAndroid.md
@@ -8,79 +8,114 @@ wip: true
 module Permission: {
   type t;
 
-  [@bs.inline "READ_CALENDAR"]
-  let readCalendar: t;
+  [@bs.module "react-native"]
+  [@bs.scope ("PermissionsAndroid", "PERMISSIONS")]
+  external readCalendar: t = "READ_CALENDAR";
 
-  [@bs.inline "WRITE_CALENDAR"]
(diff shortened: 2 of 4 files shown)


sgny:
Upon further reflection, it might be necessary to consider logic errors such as requesting some permissions and using some other key on the resulting `dict`, as below:
```
{Js.Promise.(
	PermissionsAndroid.requestMultiple([|PermissionsAndroid.Permission.readCalendar, PermissionsAndroid.Permission.writeCalendar|])
	|> then_(r => resolve(PermissionsAndroid.get(r, PermissionsAndroid.Permission.readContacts) |> Js.Console.warn))
	|> catch(err => resolve(err |> Js.Console.warn))
	)
};
```

Therefore, I revised to wrap the return type in `option`
```
[@bs.get_index] external get: (dict, Permission.t) => option(Result.t) = "";
```

sgny:
Would it cause any licensing issues to merge
```
[@bs.get_index] external get: (dict, Permission.t) => Result.t = "";
```
as `BuckleScript` is `GPL`/`LGPL` licensed?

sgny:
> Changes looks good to me.
> I am not sure what you mean about bs & license.
> If that because you copied & edited a line from their code somehow, I guess that's not a problem.
> If I misunderstand, please expand the problem.

Exactly, I checked the implementation of `Js.Dict.get` and worried afterwards.

MoOx:
Sadly, I know famous react native module that ripped some others, not just a line, without any credit. 
I really think a single line doesn't need credits. At least you mentioned it here :D

PR #500, opened 2019-05-14
Add comment for VirtualizedList props not supported by FlatList

sgny:
`getItem` and `getItemCount` are both illegal. They are already not enabled (commented out), this PR removes them for good. For reference, RN is supposed to throw an error whenever either prop is passed to `FlatList`, as below:
```
invariant(
      !getItem && !getItemCount,
      'FlatList does not support custom data formats.',
    );
```

Further, `data` prop cannot be any type, but must be an array.

Finally, it seems `_CellRendererComponent` can be enabled (this I did not test)

### DIFF
--- a/reason-react-native/src/components/FlatList.md
+++ b/reason-react-native/src/components/FlatList.md
@@ -5,6 +5,10 @@ wip: true
 ---
 
 ```reason
+// VirtualizedList props data, getItem, getItemCount and getItemLayout
+// are not supported on FlatList
+// FlatList has its own data prop, specified as array(item)
+
 include VirtualizedListElement;
 
 type separatorComponentProps('item) = {
--- a/reason-react-native/src/components/FlatList.re
+++ b/reason-react-native/src/components/FlatList.re
@@ -1,3 +1,7 @@
+// VirtualizedList props data, getItem, getItemCount and getItemLayout
+// are not supported on FlatList
+// FlatList has its own data prop, specified as array(item)
+
 include VirtualizedListElement;
 
 type separatorComponentProps('item) = {


reason-react-native/src/components/FlatList.md
```diff
@@ -25,19 +25,15 @@ external make:
...
     // VirtualizedList props
-    // ~_CellRendererComponent: VirtualizedList.cellRendererComponent('item)=?,
+    ~_CellRendererComponent: VirtualizedList.cellRendererComponent('item)=?,
     ~_ListEmptyComponent: unit => React.element=?,
     ~_ListFooterComponent: unit => React.element=?,
     ~_ListFooterComponentStyle: Style.t=?,
     ~_ListHeaderComponent: unit => React.element=?,
     ~_ListHeaderComponentStyle: Style.t=?,
     ~debug: bool=?,
-    // ~data: 'data, // any collection of 'item
```
cknitt: `~data` was defined twice and the compiler did not notice? I thought we had all/most warnings enabled?

reason-react-native/src/components/FlatList.re
```diff
@@ -18,19 +18,15 @@ external make:
     ~getItemLayout: (array('item), int) => VirtualizedList.itemLayout=?,
     ~numColumns: 'int=?,
     // VirtualizedList props
-    // ~_CellRendererComponent: VirtualizedList.cellRendererComponent('item)=?,
+    ~_CellRendererComponent: VirtualizedList.cellRendererComponent('item)=?,
```
MoOx: I didn't add this since it's not part of FlatList flow types https://github.com/facebook/react-native/blob/de0d7cfb79c7f4011d4b6748b1afc656d33fd5ac/Libraries/Lists/FlatList.js

reason-react-native/src/components/FlatList.md
MoOx: defined but commented. It was to keep list complete so we don't mess up when syncing `// VirtualizedList props`.
I think we should keep this things commented & maybe add another comment on top if we really want an explanation

MoOx:
I would prefer to keep all Virtualized props & comment the ones that aren't supported so we don't have to check all the time when we have to sync inherited props (my comments also works for the same pattern we use in lot of places)

sgny:
I wasn't sure if they were disabled by mistake and RN documentation was not quite helpful. I think it will be good to have comments so that it's clear what's going on with that module.

reason-react-native/src/components/FlatList.re
sgny: According to both RN documentation how `FlatList` wraps `VirtualList`, `CellRendererComponent` should be a valid prop. However, in looking into implementation of `VirtualList` in RN, there is a comment that says there is a conflict between that and `renderItem`. I don't know what to make of that, but we also have `CellRendererComponent` as a todo in the bindings for `VirtualList`.

MoOx: > we also have CellRendererComponent as a todo in the bindings for VirtualList.

Not sure what you mean.

MoOx:
Adding comment is fine, and we should have some locally (on top of each "disabled" props) for future bindings update imo.

reason-react-native/src/components/FlatList.re
sgny: In `VirtualizedList.re`, we have
```
// TODO
type cellRendererComponentProps('item) = {.};
type cellRendererComponent('item) =
  React.component(cellRendererComponentProps('item));
```
I couldn't spend much time to figure out if this is as it should be, but given the TODO comment, I suppose it's worth testing.

MoOx: oh I see. I missed that.
I think we should stick to flow types annotation & not try to make assumption or try the code.
And if flow annotation aren't good, we should maybe try to have those fixed & improve bindings to be in sync.

reason-react-native/src/components/FlatList.md
cknitt: Ah yes, it was commented out. 🤦‍♂️ Sorry!

reason-react-native/src/components/FlatList.re
sgny: `VirtualizedList.js` in RN has
```
  // TODO: Conflicts with the optional `renderItem` in
  // `VirtualizedSectionList`'s props.
  renderItem: $FlowFixMe<renderItemType>,
```
therefore we can only get so far depending on Flow types. If I can make sense of things, I'll prepare a PR for RN, but I test almost everything anyway.

I also seem to have misremembered the comment, `renderItem` props in `VirtualizedList` and `VirtualizedSectionList` are the ones that are in conflict (and not `CellRendererComponent`), just as `data` in `FlatList` conflicts that from `VirtualizedList`.

Back to `CellRendererComponent`, that one simply seems to have a misleading name. `renderItem` is used to, well render, each item, but the resultant `React.element` is wrapped in `CellRenderer `if the `CellRendererComponent` prop is provided, otherwise in `View`. I was able to successfully use the prop but I'll try a bit more. On the other hand, this seems hardly an essential prop. I could simply define merge `renderItem` and`CellRendererComponent` definitions so that the latter is not needed.

MoOx:
What do we want to do with this PR?
I would go for either just add a comment before commented stuff & maybe uncomment the thing we are not sure about?

sgny:
I've neglected this, I'll prepare a PR to at least add the comment.

PR #501, opened 2019-05-14
Remove links to missing documentation for deprecated components and APIs

sgny:
Links to `BackAndroid`, `ListView` and `ListViewDataSource` should probably be removed because
• documentation does not exist
• bindings don't exist and
• they are deprecated


cknitt:
@MoOx Do you think it would make sense to add bindings for these deprecated components?

Surely no one would use these for new projects - is it possible that people are using them in old projects they want to migrate?

MoOx:
Maybe we could leave this things so people that would look for those won't ask question?
And only remove them when they are officially removed from RN?

sgny:
> Maybe we could leave this things so people that would look for those won't ask question?
> And only remove them when they are officially removed from RN?

Should I simply add one-liners explaining that these are deprecated and `BackHandler`, `FlatList`, etc. should be used instead

MoOx:
That can be cool

sgny:
`CameraRoll.md` also needed to meet `prettier`.

MoOx:
(@sgny you can run `yarn format` at the root of the repo - but with the hook in place, we should need that only when we update the formating tools)

sgny:
> (@sgny you can run `yarn format` at the root of the repo - but with the hook in place, we should need that only when we update the formating tools)

`yarn format` does not match `.rei` files, should they not be?

MoOx:
They should be, that my mistake :)

PR #502, opened 2019-05-15
Update pre-commit hook and format script to process .rei files

sgny:
Presuming `.rei` files should also be formatted by `refmt`.


sgny:
> Can’t we do better for find pattern? `*.rei?` doesn’t work?

I wasn't too sure how to match multiple name patterns, but it's easy once I consulted the `man` page for `find`.

PR #508, opened 2019-05-20
add more precise methods to use Animated values as Style props

sgny:


sgny:
> Is `animatedStyle` required now?
> I don't think so. If you see a usage that cannot be covered by animated method above, we should probably keep it and rename as `animatedUnsafeStyle`?
> 
> Btw, `Animated.animated*` feels a bit weird. We could find better prefix or... What about those to `Style.Animated.*` directly?

`animatedStyle` is no longer needed, `animatedUnsafeStyle` could be an improvement, but ultimately most of these methods are unsafe (that is if they are used for anything other than as `Style` props to some `Animated` component). We could rename them `floatStyle`, `sizeStyle`, etc. That sounds better to me.

`Style.Animated.*` causes circular dependency as `Animated.View`, etc. already refer to `Style`

MoOx:
What about `Animated.Style.*`?  I don't like `*Style` or `style*` because these are not style. They return style *values*. Maybe `Animated.StyleValue.*`?
Sure If we only keep specific versions that returns `Style` "unit", they are not really unsafe, aren't they?

sgny:
> What about `Animated.Style.*`? I don't like `*Style` or `style*` because these are not style. They return style _values_. Maybe `Animated.StyleValue.*`?
> Sure If we only keep specific versions that returns `Style` "unit", they are not really unsafe, aren't they?

How about `Animated.StyleProp.*`?

Since the return types are `float`, `Color.t`, etc. they can be passed where they should not be, so in that sense they are a bit unsafe. In particular, passing a `style` object including `Animated` values as  props to `View` (instead of `Animated.View`) also causes errors.

MoOx:
`Animated.StyleProp` sound good to me!

sgny:
> `Animated.StyleProp` sound good to me!

couldn't stop iterating, `StyleProp` or `StylePropType`? the latter is even more accurate.

MoOx:
lol. I have to admit I am not a fan of *Type in names, and I even found PropType to be confusing (because I think of react prop-types...)

PR #510, opened 2019-05-21
Style: Add unsafeAddStyle, unsafeStyle & unsafeTransform as escape hatch

MoOx:
…for fancy platforms!

(and move array/list(Option) at the top - without huge comments)

This escape hatch are a necessity when using react-native-web (eg: position: fixed on the web - as you cannot always have 60fps sticky header with interpolation+translateY... Also RNW allows `translateZ`etc...)

### DIFF
--- a/reason-react-native/src/apis/Style.bs.js
+++ b/reason-react-native/src/apis/Style.bs.js
@@ -1,10 +1,8 @@
 'use strict';
 
 
-function unsafeAddProp(style, property, value) {
-  var o = Object.assign({ }, style);
-  o[property] = value;
-  return o;
+function unsafeAddStyle(style, styles) {
+  return Object.assign({ }, style, styles);
 }
 
 function pct(num) {
@@ -19,15 +17,8 @@ function rad(num) {
   return num.toString() + "rad";
 }
 
-function unsafeTransform(prop, value) {
-  var tf = { };
-  tf[prop] = value;
-  return tf;
-}
-
-exports.unsafeAddProp = unsafeAddProp;
+exports.unsafeAddStyle = unsafeAddStyle;
 exports.pct = pct;
 exports.deg = deg;
 exports.rad = rad;
-exports.unsafeTransform = unsafeTransform;
 /* No side effect */
--- a/reason-react-native/src/apis/Style.bs.js
+++ b/reason-react-native/src/apis/Style.bs.js
@@ -1,10 +1,6 @@
 'use strict';
 
 
-function unsafeAddStyle(style, styles) {
-  return Object.assign({ }, style, styles);
-}
-
 function pct(num) {
   return num.toString() + "%";
 }
@@ -17,7 +13,6 @@ function rad(num) {
   return num.toString() + "rad";
 }
 
-exports.unsafeAddStyle = unsafeAddStyle;
 exports.pct = pct;
 exports.deg = deg;
 exports.rad = rad;
--- a/reason-react-native/src/apis/Style.bs.js
+++ b/reason-react-native/src/apis/Style.bs.js
@@ -1,6 +1,12 @@
 'use strict';
 
 
+function unsafeAddProp(style, property, value) {
+  var o = Object.assign({ }, style);
+  o[property] = value;
+  return o;
+}
+
 function pct(num) {
   return num.toString() + "%";
 }
@@ -13,7 +19,15 @@ function rad(num) {
   return num.toString() + "rad";
 }
 
+function unsafeTransform(prop, value) {
+  var tf = { };
+  tf[prop] = value;
+  return tf;
+}
+
+exports.unsafeAddProp = unsafeAddProp;
 exports.pct = pct;
 exports.deg = deg;
 exports.rad = rad;
+exports.unsafeTransform = unsafeTransform;
 /* No side effect */
--- a/reason-react-native/src/apis/Style.re
+++ b/reason-react-native/src/apis/Style.re
@@ -1,5 +1,22 @@
 type t;
 
+external array: array(t) => t = "%identity";
+external arrayOption: array(option(t)) => t = "%identity";
+/* list works too since RN accept recursive array of styles (ocaml list are recursive arrays in JS)*/
+external list: list(t) => t = "%identity";
+external listOption: list(option(t)) => t = "%identity";
+
+// Escape hatch, in case something is added into RN but unsupported,
+// Useful if you play with fancy platforms
+// Use with caution
+let unsafeAddProp: (t, string, string) => t =
+  (style, property, value) => {
+    let o =
+      Js.Dict.empty()->Obj.magic->Js.Obj.assign(style->Obj.magic)->Obj.magic;
+    o->Js.Dict.set(property, value);
+    o->Obj.magic;
+  };
+
 type size = string;
 
 external pt: float => size = "%identity";
@@ -32,6 +49,13 @@ type transform;
 [@bs.obj] external skewX: (~skewX: angle) => transform = "";
 [@bs.obj] external skewY: (~skewY: angle) => transform = "";
 // @todo matrix
+
+let unsafeTransform: (string, string) => transform =
+  (prop, value) => {
+    let tf = Js.Dict.empty();
+    tf->Js.Dict.set(prop, value);
+    tf->Obj.magic;
+  };
 
 [@bs.obj]
 // Layout Props (https://facebook.github.io/react-native/docs/layout-props#props)
@@ -222,40 +246,3 @@ external style:
   ) =>
   t =
   "";
-
-/*
- <View style=array([|
-   styles##thing,
-   styles##whatever,
- |])>
-   */
-external array: array(t) => t = "%identity";
-
-/*
- <View style=arrayOption([|
-   Some(styles##thing),
-   Some(styles##whatever),
-   optionalStyle,
-   cond ? Some({something:"dynamic"}) : None
- |])>
- */
-external arrayOption: array(option(t)) => t = "%identity";
-
-/* list works too since RN accept recursive array of styles (list are just recursive arrays)*/
-/*
- <View style=list([
-   styles##thing,
-   styles##whatever,
- ])>
-   */
-external list: list(t) => t = "%identity";
-
-/*
- <View style=listOption([
-   Some(styles##thing),
-   Some(styles##whatever),
-   optionalStyle,
-   cond ? Some({something:"dynamic"}) : None
- ])>
- */
-external listOption: list(option(t)) => t = "%identity";
--- a/reason-react-native/src/apis/Style.re
+++ b/reason-react-native/src/apis/Style.re
@@ -9,13 +9,12 @@ external listOption: list(option(t)) => t = "%identity";
 // Escape hatch, in case something is added into RN but unsupported,
 // Useful if you play with fancy platforms
 // Use with caution
-let unsafeAddProp: (t, string, string) => t =
-  (style, property, value) => {
-    let o =
-      Js.Dict.empty()->Obj.magic->Js.Obj.assign(style->Obj.magic)->Obj.magic;
-    o->Js.Dict.set(property, value);
-    o->Obj.magic;
-  };
+[@bs.val]
+external unsafeAddStyle_: (Js.Dict.t('a), t, 'b) => t = "Object.assign";
+let unsafeAddStyle: (t, Js.t('a)) => t =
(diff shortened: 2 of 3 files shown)


reason-react-native/src/apis/Style.re
```diff
@@ -1,5 +1,28 @@
...
+    ->Js.Obj.assign(
+        {
+          let o = Js.Dict.empty();
+          o->Js.Dict.set(property, value);
+          o->Obj.magic;
+        },
+      )
+    ->Obj.magic;
+  };
+
```
MoOx: This is basically the same technique as in reason-react codebase. Not zero-cost but we can't do magic here :/

sgny:
Sorry I’ve been ridiculously pressed for tlme.

I think it would be better to avoid having to create a copy of the existing style object, especially if multiple additional keys will need to be defined - though I don’t know what those keys might be.

I think it would be better to add any such keys to these bindings, whether with a Style.t constructor that accepts only those or otherwise.

MoOx:
> I think it would be better to avoid having to create a copy of the existing style object, especially if multiple additional keys will need to be defined - though I don’t know what those keys might be.

We cannot know in advance. If user is using the bindings with an unexpected platform (first example: the web) we cannot predict all properties we don't list (we should be close as possible to RN imo & no accept everything). That's why I would prefer to offer an escape hatch (can be done in userland, but still...)

> I think it would be better to add any such keys to these bindings, whether with a Style.t constructor that accepts only those or otherwise.

Would you see another way to accept random prop/value (string/string) that this way (used by ReasonReact for dom styles too)?

I just pushed a revision of the implementation to make it shorter with a single object assign (reminder: which isn't a deep copy).

cknitt:
This way of accepting random props/values is fine with me.

Two remarks:

1. `(t, string, string)`: to distinguish between the two string params, I think we should used labelled arguments: `(t, ~prop: string, ~value: string)`.
2. For cases where the user wants to add multiple such props at once, we could provide an additional function `unsafeAddProps: (t, Js.t('a)) => t`.

sgny:
Since none of these approaches bring any real guarantees, we could simply have a function that converts a `Js.dict(string)` to `Style.t` which would be zero-cost.

A significant drawback is how clunky creating a `Js.dict(string)` object is.

I suppose simply allowing any type (possibly defined with a constructor using the `[@bs.obj]` attribute)  to be converted into `Style.t` should not really cause any issues.

cknitt:
Hmm, yes, maybe we'd be better off just having

```reason
external unsafeFromDict: Js.Dict.t(string) => t = "%identity";
external unsafeFromJs: Js.t('a) => t = "%identity";
```

One can always combine such a style with others using `Style.array`.

sgny:
> Hmm, yes, maybe we'd be better off just having
> 
> ```reason
> external unsafeFromDict: Js.Dict.t(string) => t = "%identity";
> external unsafeFromJs: Js.t('a) => t = "%identity";
> ```
> 
> One can always combine such a style with others using `Style.array`.

Exactly, passing an array of `Style.t` is overall lower cost.

MoOx:
> A significant drawback is how clunky creating a Js.dict(string) object is.

That's why I didn't start with something like that. I make me want to kill myself everytime I need to create a Js.Dict...

My idea was: since we are doing something unsafe, we should not require plenty of code to do it, as it's still unsafe anyway... 
The nice part with `unsafeAdd` is that you can do `style(....)->unsafeAddProp("prop", "value")` which is brief enough to not have to do any weird & verbose things to add something unsafe... *or unsupported* (yet?) by the bindings.

I would prefer to keep this `unsafeAddProp` for 2 reasons:
- usage simplicity (and without label... I don't think anyone will try to do `unsafeAddProp("some css value", "the key")` as it's totally not intuitive
- keep thing close to reason-react Style method.

But I am having another idea, stay tuned.

MoOx:
My idea can't be executed correctly so forget about it.

About the array thing, I prefer to have everything in my stylesheets. So yeah in this case, I should probably have an `unsafeStyle` which accept an array of js/dict. But that means I have to use 2 styles when passing them, so I have to use Style.array, & that's clearly way more verbose that a simple `unsafeAddProp`.
And as I said, verbosity isn't an issue if it's adding value (eg: types). In this case it's totally not adding value - as that's totally unsafe (right?). So I am bothered about having a verbose solution, even if zero-zero cost, that's unsafe. I would prefer a not-so-zero cost solution, way less verbose.

If we agree to got this way (no so zero unsafe cost - as proposed in the PR), we could also have `unsafeAddProps` that's basically doing the same as the singular version, with a loop (cost here!).
That said, I think since this thing is an edge case & probably most people won't use it, except weirdos like me.

I think this is acceptable

```reason
style(...)
->unsafeAddProp("one", "1")
->unsafeAddProp("two", "2")
```

It's verbose yeah, but verbose about the un-safety side, so I am ok with that :D (you have the right to say this entire point of view is weird haha).

I don't think people will use `unsafeFromDict`, as creating a dict is OMG verbose level & Js.t() will require a type  (so again, verbosity flag - people don't want that, especially for unsafe stuff).

I know the solution I am proposing in this PR isn't optimal & have a (tiny) cost, but
- this is unsafe anyway, so we should keep usage short as `unsafe` in the name is clear enough
- easy to use
- similar to reason-react
- pipeable (even if runtime cost is to consider, should be reserved to edge case).

It's all about compromise. If people want to do lot of too many unsafe things, they could totally do some Obj.magic user land anyway.

Here I just want to offer a "native" (shipped with this bindings) escape hatch that easy to use & acceptable.

cknitt:
> ... & Js.t() will require a type ...

It would just be

```reason
Style.(style->unsafeAddProps({"one": "1", "two": "2"}));
```

with the added advantage of supporting non-string props (similar to margin).

Zero-cost version:

```reason
Style.(array([style, unsafe({"one": "1", "two": "2"})]));
```

sgny:
> > ... & Js.t() will require a type ...
> 
> It would just be
> 
> ```reason
> Style.(style->unsafeAddProps({"one": "1", "two": "2"}));
> ```
> 
> with the added advantage of supporting non-string props (similar to margin).
> 
> Zero-cost version:
> 
> ```reason
> Style.(array([style, unsafe({"one": "1", "two": "2"})]));
> ```

I always forget one does not need to specify the type beforehand when constructing a `Js.t` object in this manner.

reason-react-native/src/apis/Style.bs.js
```diff
@@ -1,5 +1,13 @@
 'use strict';
 
+var Js_dict = require("bs-platform/lib/js/js_dict.js");
+
+function unsafeAddProp(style, property, value) {
```
cknitt: Did you forget to update the .bs.js file? It does not correspond to the .re file.

MoOx: No. Compilation fails & I didn't yet find why. Saying the .rei doesn't match the .re but I don't see the issue :/

sgny:
I haven't had a chance to see what might be causing the compile to fail, but we could define `unsafeAddProp` as below. I know @MoOx had that in a previous iteration.

```reason
[@bs.set_index] external unsafeAddProp: (t, string, 'a) => unit = "";
```
It doesn't return `Style.t` and is not pipeable, but could be a nice function to have. Obviously we could easily define a function 
```reason
let unsafeAddPropPipeable: (t, string, 'a) => t =
    (style, propName, propValue) => {
        unsafeAddProp(style, propName, propValue);
        style
};
```

reason-react-native/src/apis/Style.rei
```diff
@@ -1,5 +1,12 @@
 type t;
 
+external array: array(t) => t = "%identity";
+external arrayOption: array(option(t)) => t = "%identity";
+external list: list(t) => t = "%identity";
+external listOption: list(option(t)) => t = "%identity";
+let unsafeAddStyle: (t, Js.t('a)) => t;
+let unsafeStyle: Js.t('a) => t;
```
sgny: You have a `let` instead of `external`. Simply replacing with 
```reason
external unsafeStyle: Js.t('a) => t = "%identity";
```
should be enough to resolve

reason-react-native/src/apis/Style.rei
```diff
@@ -33,6 +40,8 @@ type transform;
 [@bs.obj] external skewY: (~skewY: angle) => transform = "";
 // @todo matrix
 
+let unsafeTransform: Js.t('a) => transform;
```
cknitt: Same here.

cknitt:
Just tested. Compilation will work fine if you change

```reason
    ->Js.Obj.assign(styles)
```

in the .re file to

```reason
    ->Js.Obj.assign(styles->Obj.magic)
```

Don't as me why. 😉

cknitt:
This makes the function really quite magic:

```reason
let unsafeAddStyle: (t, Js.t('a)) => t =
  (style, styles) => {
    Js.Dict.empty()
    ->Obj.magic
    ->Js.Obj.assign(style->Obj.magic)
    ->Js.Obj.assign(styles->Obj.magic)
    ->Obj.magic;
  };
```

😂

sgny:
Should we a function to merge style objects?
```reason
[@bs.val]
external merge: (t, t) => t = "Object.assign";
```
Then, we could simply have
```reason
let unsafeAddStyle: (t, Js.t('a)) => t =
  (style, styles) => {
   	merge(style, unsafeStyle(styles))
  };
```
If there are common keys, this seems to cause errors, I don't know if `magic` helps with that.

MoOx:
Object.assign mutate the first argument. That's why your proposed `merge` & `unsafeAddStyle` aren't what we want & that's why I used 2 object assign.

reason-react-native/src/apis/Style.rei
MoOx: That's not an issue. You can define `external thing` in interface as `let thing`. Btw, I changed some for the moved code.

MoOx: Actually it works with let, but create unnecessary code in the js, so I adjusted as you recommended.

reason-react-native/src/apis/Style.rei
MoOx: Fixed

sgny:
> Object.assign mutate the first argument. That's why your proposed `merge` & `unsafeAddStyle` aren't what we want & that's why I used 2 object assign.

Sure, neither are pure functions. However, they would be more efficient if the goal is to construct one style. If the goal is to construct multiple styles based off one style, sure a copy is necessary. Perhaps `impureAddStyle` would be clearer.

MoOx:
We could even add Object.assign2 binding to have a single js call. Let me do that.

MoOx:
Now it's even shorter :)

reason-react-native/src/apis/Style.re
```diff
@@ -1,5 +1,25 @@
...
+[@bs.val]
+external assign2: (Js.t('a), Js.t('b), Js.t('c)) => Js.t('d) =
+  "Object.assign";
+
+// Escape hatch, in case something is added into RN but unsupported,
+// Useful if you play with fancy platforms
+// Use with caution
+let unsafeAddStyle: (t, Js.t('a)) => t =
+  (style, styles) => {
+    Obj.(assign2(Js.Dict.empty()->magic, style->magic, styles->magic)->magic);
```
sgny: Slightly pedantic, but we could instead have
```reason
    Obj.(assign2(Js.Obj.empty(), style->magic, styles->magic)->magic);
```
which saves us one piece of `magic`, although JS output does not change even slightly.

sgny:
I think we have arrived at a good place.

reason-react-native/src/apis/Style.re
cknitt: Agreed

MoOx: I had to change assign2 method so it takes a Js.Dict for the first arg.

MoOx:
Oh just thought about a solution without any (explicit) magic:

```reason
[@bs.val]
external unsafeAddStyle_: (Js.Dict.t('a), t, 'b) => t = "Object.assign";

let unsafeAddStyle: (t, Js.t('a)) => t =
  (style, styles) => unsafeAddStyle_(Js.Dict.empty(), style, styles);
```

cknitt:
Zero cost:

```reason
[@bs.val] external unsafeAddStyle: ([@bs.as {json|{}|json}] _, t, Js.t('a)) => t = "Object.assign";
```

MoOx:
@cknitt 🤯🤯🤯

PR #511, opened 2019-05-29
Style: minor fixes in `style`, and add `viewStyle`, `textStyle` &  `imageStyle`

### LINKED ISSUE #436: More granular style for Image/Text etc?
While working on #413 I found that ImageBackground accept style & imageStyle, which aren't currently typed the same in new bindings.

But both are the same (even if RN doc on the website is confusing about it) => https://github.com/facebook/react-native/blob/master/Libraries/StyleSheet/StyleSheetTypes.js#L621-L626

Do you think we should do like RN is doing with flow more strong check on styles props depending on what element is that?

@cknitt @sgny Do you think we should (& can easily) replicate this ViewStyle, TextStyle, ImageStyle without too much pain?

MoOx:
Closes #436

### DIFF
--- a/reason-react-native/src/apis/Style.re
+++ b/reason-react-native/src/apis/Style.re
@@ -134,7 +134,7 @@ external style:
                       | `capitalize
                     ]
                       =?,
-    ~writingDirection: [@bs.string] [ | `auto | `ltr | `ltr]=?,
+    ~writingDirection: [@bs.string] [ | `auto | `ltr | `rtl]=?,
     // View styles https://facebook.github.io/react-native/docs/view-style-props
     ~backfaceVisibility: [@bs.string] [ | `visible | `hidden]=?,
     ~backgroundColor: Color.t=?,
@@ -208,7 +208,7 @@ external style:
     // ~borderTopWidth: float=?,
     // ~borderWidth: float=?,
     ~bottom: size=?,
-    ~direction: [@bs.string] [ | [@bs.as "inherit"] `inherit_ | `ltr | `ltr]=?,
+    ~direction: [@bs.string] [ | [@bs.as "inherit"] `inherit_ | `ltr | `rtl]=?,
     ~display: [@bs.string] [ | `none | `flex]=?,
     ~_end: size=?,
     ~flex: float=?,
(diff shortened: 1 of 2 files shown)


reason-react-native/src/apis/Style.re
```diff
@@ -107,11 +300,55 @@ external style:
...
+    // ~borderRightWidth: float=?,
+    // ~borderStartWidth: float=?,
+    // ~borderTopWidth: float=?,
+    // ~borderWidth: float=?,
     ~bottom: size=?,
     ~direction: [@bs.string] [ | [@bs.as "inherit"] `inherit_ | `ltr | `ltr]=?,
     ~display: [@bs.string] [ | `none | `flex]=?,
-    ~elevation: float=?,
-    ~_end: float=?,
+    ~_end: size=?,
```
MoOx: fixed it :)

sgny:
Have you been able to determine where `borderBottomWidth` may be an `Animated` value and where it cannot?

MoOx:
@sgny you can see here https://github.com/facebook/react-native/blob/099be9b35634851b178e990c47358c2129c0dd7d/Libraries/StyleSheet/StyleSheetTypes.js#L564-L569 that this props support animatednode, but in earlier declaration (in `____LayoutStyle_Internal`) they only support number https://github.com/facebook/react-native/blob/099be9b35634851b178e990c47358c2129c0dd7d/Libraries/StyleSheet/StyleSheetTypes.js#L261-L301
I tried to go deeper but not for long, so I assumed they did this for a reason :)

sgny:
> @sgny you can see here https://github.com/facebook/react-native/blob/099be9b35634851b178e990c47358c2129c0dd7d/Libraries/StyleSheet/StyleSheetTypes.js#L564-L569 that this props support animatednode, but in earlier declaration (in `____LayoutStyle_Internal`) they only support number https://github.com/facebook/react-native/blob/099be9b35634851b178e990c47358c2129c0dd7d/Libraries/StyleSheet/StyleSheetTypes.js#L261-L301
> I tried to go deeper but not for long, so I assumed they did this for a reason :)

After I read your comment, I noticed that too, but I still have no idea where I cannot simply animate `borderBottomWidth`, so I was curious if you'd run into something like that. Since there is no documentation to clarify, we cannot really go deeper at this time; this is already a great effort.

PR #512, opened 2019-05-30
fix type for Gesture Responder onResponderTerminationRequest handler

sgny:
While preparing documentation for `PanResponder`, I realized that the type for the `onResponderTerminationRequest` handler callback is incorrect according to [RN documentation](https://facebook.github.io/react-native/docs/gesture-responder-system)


MoOx:
👍

PR #513, opened 2019-05-30
Add documentation for PanResponder API

sgny:


reason-react-native/src/apis/PanResponder.md
```diff
@@ -1,9 +1,137 @@
 ---
-id: apis/Packager
-title: Packager
-wip: true
+id: apis/PanResponder
+title: PanResponder
```
MoOx: You can add `officialDoc: http://facebook.github.io/react-native/docs/panresponder` so a small link will be visible below huge title & will point to RN doc :)

reason-react-native/src/apis/PanResponder.md
```diff
@@ -1,9 +1,137 @@
...
 
+`PanResponder` allows more fine-grained control of the _Gesture Responder
+System_, as the handlers can depend also on `gestureState`, in addition to the
+`Event.pressEvent` which are passed to the _Gesture Responder System_ handlers.
+In essence, by means of a custom `PanResponder`, _Gesture Responder System_
+handlers are extended to also depend upon the `gestureState`.
+
+## Methods
+
+- `create` allows creation of a `PanResponder` given a `config`
```
MoOx: Instead of single list item (because that's how it's rendering the way it is), I think we should go with normal title ``### `method` ``. This will allows anchor links & will be better with incoming search engine :)

reason-react-native/src/apis/PanResponder.md
```diff
@@ -1,9 +1,137 @@
 ---
-id: apis/Packager
-title: Packager
-wip: true
+id: apis/PanResponder
+title: PanResponder
 ---
 
+`PanResponder` allows more fine-grained control of the _Gesture Responder
```
MoOx: A link to http://facebook.github.io/react-native/docs/gesture-responder-system would be nice

reason-react-native/src/apis/PanResponder.md
```diff
@@ -46,32 +176,109 @@ external config:
...
+
+As props spread is not possible with the ReasonML JSX 3 syntax, each _Gesture
+Responder System_ callback needs to be separately specified, with the relevant
+value obtained by means of the associated getter.
+
+While props spread may not be available, copy-paste is probably supported by
+your editor...
+
+```reason
+onMoveShouldSetResponder = panHandlers->PanResponder.onMoveShouldSetResponder
```
MoOx: space around = are probably unnecessary as they will be removed by refmt don't you think?

reason-react-native/src/apis/PanResponder.md
```diff
@@ -46,32 +176,109 @@ external config:
...
+open ReactNative;
+
+let windowHeight = Dimensions.get(`window)##height->float_of_int;
+let windowWidth = Dimensions.get(`window)##width->float_of_int;
+
+let containerStyle =
+  Style.(
+    style(
+      ~width=pt(windowWidth),
+      ~height=pt(windowHeight),
```
MoOx: Small detail but I think we should go with `->pt` in all docs. This looks like more a unit (`20px` ⋍ `20.->pt` more than `pt(20.)`) imo.

reason-react-native/src/apis/PanResponder.md
```diff
@@ -1,9 +1,137 @@
...
+title: PanResponder
 ---
 
+`PanResponder` allows more fine-grained control of the _Gesture Responder
+System_, as the handlers can depend also on `gestureState`, in addition to the
+`Event.pressEvent` which are passed to the _Gesture Responder System_ handlers.
+In essence, by means of a custom `PanResponder`, _Gesture Responder System_
+handlers are extended to also depend upon the `gestureState`.
+
+## Methods
```
MoOx: Maybe before directly listing method, we should explain who to pass those? I think the block below with the example could be better before this listing.

reason-react-native/src/apis/PanResponder.md
```diff
@@ -1,9 +1,137 @@
...
+- `onResponderEnd` is a getter method to access the `onResponderEnd` callback of
+  a `panHandlers` object
+
+```reason
+onResponderEnd(panHandlers)
+```
+
+## Types
+
+- `Event.pressEvent` is `Event.syntheticEvent(pressEventPayload)` where the
```
MoOx: I think we shouldn't doc other module like this & instead make links to ease maintenance.
You could move that to `Event.md` & make a link to it via the expected anchor link?

sgny: Perhaps replace this with a link to `Event.md` when that's ready?

MoOx: Yeah you can link to `/bs-react-native/en/docs/apis/Event/` (not ideal but I will deal with this problem later)

reason-react-native/src/apis/PanResponder.md
sgny: Should the order be
• Types
• Methods
• Example
then?

An example could be hard to make sense of without an idea of the types and methods

MoOx: Yeah depending on the thing an example can indeed be hard to digest.
This order is ok for me. A "textual transition" can even explain it if necessary :)

PR #515, opened 2019-05-30
Add documentation for Event API

sgny:
Perhaps we can get this included in the release after all.
```
officialDoc: https://github.com/facebook/react-native/blob/master/Libraries/Types/CoreEventTypes.js
```
because code is documentation 😄

### DIFF
--- a/reason-react-native/src/apis/Event.md
+++ b/reason-react-native/src/apis/Event.md
@@ -1,12 +1,17 @@
 ---
 id: apis/Event
 title: Event
-wip: true
+officialDoc: https://github.com/facebook/react-native/blob/master/Libraries/Types/CoreEventTypes.js
 ---
 
-```reason
-// see https://github.com/facebook/react-native/blob/master/Libraries/Types/CoreEventTypes.js
+These are the types of objects returned in `Event` callbacks. Types are
+instances of `syntheticEvent('a)` and `responderSyntheticEvent('a)`.
 
+## Parametrised Types
+
+### `syntheticEvent('a)`
+
+```reason
 type syntheticEvent('a) = {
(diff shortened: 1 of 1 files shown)


reason-react-native/src/apis/Event.md
```diff
@@ -1,12 +1,17 @@
...
-wip: true
+officialDoc: https://github.com/facebook/react-native/blob/master/Libraries/Types/CoreEventTypes.js
 ---
 
-```reason
-// see https://github.com/facebook/react-native/blob/master/Libraries/Types/CoreEventTypes.js
+These are the types of objects returned in `Event` callbacks. Types are
+instances of `syntheticEvent('a)` and `responderSyntheticEvent('a)`.
+
+## Prametrised Types
```
MoOx: Parameterized? (it feels weid to me & I was thinking "Generic", but "parameters" & "parameterized" seems what's used in reasonml docs https://reasonml.github.io/docs/en/more-on-type#type-argument

reason-react-native/src/apis/Event.md
```diff
@@ -1,12 +1,17 @@
 ---
 id: apis/Event
 title: Event
-wip: true
+officialDoc: https://github.com/facebook/react-native/blob/master/Libraries/Types/CoreEventTypes.js
```
MoOx: 😄

reason-react-native/src/apis/Event.md
sgny: Doesn't feel weird to me, but I had to think about what to call these, so I checked the documentation.

sgny:
fixed the typo `Prametrised`

PR #517, opened 2019-05-30
link PanResponder documentation to Event documentation

sgny:
Since `Event` documentation is in place, we can link to that instead of specifying `pressEvent` here.

I'm not sure if the anchor is right, though, perhaps backticks are required. I would appreciate comments on that.

### DIFF
--- a/reason-react-native/src/apis/PanResponder.md
+++ b/reason-react-native/src/apis/PanResponder.md
@@ -14,24 +14,7 @@ handlers are extended to also depend upon the `gestureState`.
 ## Types
 
 ### `Event.pressEvent`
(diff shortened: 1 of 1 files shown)


reason-react-native/src/apis/PanResponder.md
```diff
@@ -14,24 +14,7 @@ handlers are extended to also depend upon the `gestureState`.
...
-  "locationX": float,
-  "locationY": float,
-  "pageX": float,
-  "pageY": float,
-  "target": Js.Nullable.t(float),
-  "timestamp": float,
-  "touches": array(pressEventPayload),
-}
-```
+see [`Event`](../Event/#pressEvent)
```
MoOx: I will adjust anchor later if necessary, they are not setup yet.

PR #518, opened 2019-05-30
Add openSettings to Linking API

sgny:
I think RN 0.60 is about to arrive, perhaps we can include this in our release


MoOx:
Should we also comment the binding for version?

sgny:
I'll add proper documentation in a short while, perhaps adding the information to the bindings won't be necessary.

PR #519, opened 2019-05-30
Add documentation for Linking API

sgny:
@MoOx do you think stating version requirements in the bindings would be necessary?

### DIFF
--- a/reason-react-native/src/apis/Linking.md
+++ b/reason-react-native/src/apis/Linking.md
@@ -1,50 +1,189 @@
 ---
 id: apis/Linking
 title: Linking
-wip: true
+officialDoc: https://facebook.github.io/react-native/docs/linking.html
 ---
 
+Only for projects supporting Native Code. Requires the project to have been
+created using `react-native init` or ejected afterwards if created using
+`expo init` or `create-react-native-app`.
+
+## Methods
+
+### `openURL`
+
+To attempt opening a URL. URL should be specified as `string`, and the method
+returns `unit` wrapped in a promise. Promise is resolved if the user approves
+the request (through the _open dialog_) or the link is opened automatically. If
+the user rejects the request or there are no registered applications supporting
+the URL, the promise is rejected. It is recommended to use the `canOpenURL`
+method beforehand, to verify that the URL can indeed be opened.
+
 ```reason
-type extra;
-
-[@bs.obj] external extra: (~key: string, ~value: 'a) => extra = "";
-
-[@bs.scope "Linking"] [@bs.module "react-native"]
-external openURL: string => Js.Promise.t(unit) = "";
-
-[@bs.scope "Linking"] [@bs.module "react-native"]
-external canOpenURL: string => Js.Promise.t(bool) = "";
-
-[@bs.scope "Linking"] [@bs.module "react-native"]
-external getInitialURL: unit => Js.Promise.t(Js.Null.t(string)) = "";
+openURL: string => Js.Promise.t(unit)
 ```
 
-As of React Native 0.60
+### `canOpenURL`
+
+To determine whether a URL can be opened by registered applications. URL should
+be specified as `string`, and the method returns `bool` wrapped in a promise.
+The promise will be rejected if it is impossible to check if the URL can be
+opened on Android or for iOS 9 and later, an appropriate entry does not exist
+for the `LSApplicationQueriesSchemes` key in `Info.plist`.
+
 ```reason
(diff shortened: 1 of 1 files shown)


reason-react-native/src/apis/Linking.md
```diff
@@ -1,50 +1,112 @@
...
+returns `unit` wrapped in a promise, that is `Js.Promise.t(unit)`. Promise is
+resolved if the user approves the request (through the _open dialog_) or the
+link is opened automatically. If the user rejects the request or there are no
+registered applications supporting the URL, the promise is rejected. It is
+recommended to use the `canOpenURL` method beforehand, to verify that the URL
+can indeed be opened.
+
 ```reason
-type extra;
+openURL(string)
```
MoOx: I was wondering if we should use real world snippet by including module name.
That would mean:

```reason
open ReactNative;

Linking.openURL("http://some.url")
```

We could potentially omit `open ReactNative;` as it like a requirement (at least I guess we are going to promote this usage), but we should probably show something that people can easily copy/paste/tweak without too much thing to think about.

What do you think?

MoOx:
About version number, @cknitt think following RN version is a good idea, and since that was one of my idea, I am ok with that, so maybe we don't have to document this stuff...

reason-react-native/src/apis/Linking.md
sgny: Sure, I'll create an example. Since it wouldn't hurt, I'll keep the open statement to make it as self-contained as possible.

sgny:
> About version number, @cknitt think following RN version is a good idea, and since that was one of my idea, I am ok with that, so maybe we don't have to document this stuff...

We're all in agreement on that, then. I will keep the version notes in this documentation while we decide with which version we tag our own release. That is while we decide whether we call this 0.60 or 0.5x.

reason-react-native/src/apis/Linking.md
sgny: I revised to list the type for each method, considering that the example should be sufficient to demonstrate how the methods can be used.

Should we nevertheless have examples of how each method is called, as in the example you have given for `openURL`?

cknitt:
I meant to request changes, not to approve, sorry!

reason-react-native/src/apis/Linking.md
MoOx: I would do that for all code blocks indeed. I found this much more clear than a "raw api" declaration in code as you can directly identify how you are going to use it.
@cknitt thoughts?

MoOx: ORRR maybe we should have both like this

- code showing api, very concise & with type, not values (exactly what you did) 
- description blah blah blah
- very tiny example (like the one I shared)

And a "real world" example at the end, mixing multiple usage if possible (exactly what you did).

reason-react-native/src/apis/Linking.md
```diff
@@ -1,50 +1,192 @@
...
+removeEventListener([ `url], {. "url": string} => unit) => unit
+```
+
+### Example
+
+```reason
+open ReactNative;
+
+let windowHeight = Dimensions.get(`window)##height->float_of_int;
+let windowWidth = Dimensions.get(`window)##width->float_of_int;
```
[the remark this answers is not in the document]
sgny: Probably, most uses will require conversion to `float`.

reason-react-native/src/apis/Linking.md
```diff
@@ -1,50 +1,192 @@
...
+```
 
-[@bs.obj] external extra: (~key: string, ~value: 'a) => extra = "";
+### `canOpenURL`
 
-[@bs.scope "Linking"] [@bs.module "react-native"]
-external openURL: string => Js.Promise.t(unit) = "";
+To determine whether a URL can be opened by registered applications. URL should
+be specified as `string`, and the method returns `bool` wrapped in a promise,
+that is `Js.Promise.t(unit)`. The promise will be rejected if it is impossible
```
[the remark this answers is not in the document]
sgny: I was too hasty with the copy&paste; fixed.

At first I did not actually have explicit type information (as in the code blocks now), so wanted to include that in the description. With the types explicitly stated now, I don't think that explanation is necessary anywhere.

sgny:
> Feel free to merge, we can improve the global layout of the doc later.

I fixed the error @cknitt noticed and cleaned up the description, but individual examples might have to wait.

MoOx:
Thanks for all this work!

PR #520, opened 2019-05-30
Add documentation for StyleSheet API

sgny:
Also added the relevant link to `Style.md`

### DIFF
--- a/reason-react-native/src/apis/StyleSheet.md
+++ b/reason-react-native/src/apis/StyleSheet.md
@@ -1,20 +1,109 @@
 ---
 id: apis/StyleSheet
 title: StyleSheet
-wip: true
+officialDoc: https://facebook.github.io/react-native/docs/stylesheet
 ---
 
+`StyleSheet` mainly has a method to [create](#create) an object containing many
+styles, similar to a CSS StyleSheet. This is helpful to keep style definitions
+at fewer locations, away from the render function. Separating concerns in your
+code in this manner should help readability. Named styles might also help
+differentiate between your components. Finally, allowing components to refer to
+styles by ID, it also reduces the amount of data transferred over the JS bridge.
+
+There is also a [`flatten`](#flatten) method to merge multiple styles into a single style
+and several [constants](#constants) for commonly used style definitions.
+
+## Methods
+
+### `create`
+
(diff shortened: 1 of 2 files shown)


reason-react-native/src/apis/StyleSheet.md
```diff
@@ -1,20 +1,104 @@
...
+the key.
+
+```reason
+create: Js.t('a) => Js.t('a)
+```
+
+### `flatten`
+
+Takes an array of styles (of type `Style.t`) and returns a single style (also of
+type `Style.t`).
```
cknitt: Maybe add a comment that using styles created with `Style.array` in `StyleSheet.create` will lead to a runtime error, so `StyleSheet.flatten` *must* be used in that case. (`Style.array` is fine for passing styles to component props though.)

reason-react-native/src/apis/StyleSheet.md
```diff
@@ -1,20 +1,104 @@
...
+operator when fields are not explicitly set, you may use the `flatten` method
+instead.
+
+```reason
+absoluteFill: Style.t
+```
+
+### `absoluteFillObject`
+
+This is identical to `absoluteFill` when used with pure functions.
```
MoOx: I won't say that as this thing "identical" as it's not. See #523

reason-react-native/src/apis/StyleSheet.md
```diff
@@ -1,20 +1,104 @@
...
+
+### `create`
+
+Takes and returns `Js.t('a)` objects, with `string` keys and `Style.t` values.
+Individual styles can be accessed using `##`, followed by the name specified as
+the key.
+
+```reason
+create: Js.t('a) => Js.t('a)
+```
```
MoOx: Same question as https://github.com/reasonml-community/bs-react-native/pull/519#discussion_r289291691

sgny: Perhaps we revisit this issue later, but I don't think there is a way to exclude `array(Style.t)` without a lot of pain elsewhere.

reason-react-native/src/apis/StyleSheet.md
sgny: Added such comments

MoOx:
👍

PR #521, opened 2019-05-31
Add documentation for ImageEditor API

sgny:


reason-react-native/src/apis/ImageEditor.md
```diff
@@ -1,45 +1,231 @@
...
+                ~width=displayWidth,
+                ~height=
+                  (
+                    if (state.imageLoaded) {
+                      1.;
+                    } else {
+                      0.;
+                    }
+                  )
+                  *. displayHeight,
```
cknitt: Why not just put `displayHeight` instead of `1.` above?

sgny: I was trying different multiples to align the text in the center of the box, but ultimately I had to pick `0.`, so `displayHeight` and `0.` would be cleaner.

sgny: Removed that artifact

MoOx:
Don't you think instead of those if/else we could use ternary thing `cond ? ifValue : elseValue` to make things more concise?

sgny:
> Don't you think instead of those if/else we could use ternary thing `cond ? ifValue : elseValue` to make things more concise?

I was surprised to see `refmt` make that change, as I don't really like that syntax, I had only `if` statements.

Since the example isn't really for us, I did change the others as well.

PR #528, opened 2019-05-31
Revise documentation for LayoutAnimation API

sgny:
Revised documentation to
• list function signatures for each method
• take into account the change from int to float for width and height in Dimensions


MoOx:
👌

PR #532, opened 2019-05-31
Alternative git-monorepo-usage-trick

### LINKED ISSUE #358: git-monorepo-usage-trick dies in CI only
Hi guys,
We're using the latest of this inside of our own monorepo which has our mobile app, a webpage, an api server, and other stuff. To get the mobile app to play nice we did have to update from the latest published version to the repo version, and so we tacked the monorepo trick into our bsconfig. Locally, this works great. In CI however, it dies, because none of the files are present at the time it runs, but they are after `bsb -make-world` is complete.

I tried sleeping first inside the trick (I pulled the script out locally to be able to run a modified version)

```bash
if [ ! -f $1 ]; then
    echo "File not found: $1"
    sleep 1
fi

if [ ! -f $1 ]; then
    echo "File still not found: $1"
    sleep 5
fi

if [ ! -f $1 ]; then
    echo "File still not found, but avoiding errors: $1"
fi

sed -i '' -e 's/require("bs-react-native\//require("bs-react-native\/bs-react-native\//g' $1 || echo ""
sed -i '' -e 's/from "bs-react-native\//from "bs-react-native\/bs-react-native\//g' $1 || echo ""
```

However, it appears that the file is never present until the hook runs. 

We've resorted to a fork of the trick that doesn't run in CI, and instead when we're in CI we run it on all .bs.js files when `-make-world` completes. 

We are running with a custom build agent inside gitlab, based on the node10.15 image. I've tried loading that image into docker locally, and cloned the repo and ran the build, and it was successful. Not entirely sure if this is actually a bs-react-native issue vs a bs-platform issue, so I can file this over there if that's more appropriate.

MoOx:
Closes #358

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,6 +38,8 @@ If you use this bindings from the git repo, be sure to use
   "js-post-build": {
     "cmd": "./node_modules/bs-react-native-monorepo/git-monorepo-usage-trick"
   }
+  // if you this tricky script doesn't work for you correctly, try
+  // "cmd": "./node_modules/bs-react-native-monorepo/git-monorepo-usage-trick-node"
 }
 ```
 
--- a/README.md
+++ b/README.md
@@ -37,9 +37,9 @@ If you use this bindings from the git repo, be sure to use
   // and will just need to be removed when you won't use git directly anymore
   "js-post-build": {
     "cmd": "./node_modules/bs-react-native-monorepo/git-monorepo-usage-trick"
+    // if this tricky script doesn't work for you correctly, try
+    // "cmd": "./node_modules/bs-react-native-monorepo/git-monorepo-usage-trick-node"
   }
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -38,6 +38,8 @@ If you use this bindings from the git repo, be sure to use
   "js-post-build": {
     "cmd": "./node_modules/bs-react-native-monorepo/git-monorepo-usage-trick"
   }
+  // if you this tricky script doesn't work for you correctly, try
```
cknitt: if you this tricky script
-> if this tricky script

(Why wouldn't it work BTW?)

MoOx: Windows or any system that doesn't have vim/ex (I think circle-ci doesn't)

MoOx: I fixed the thing. Thanks :)

PR #535, opened 2019-06-01
Replace Style.pt with Style.dp

### LINKED ISSUE #530: Replace Style.pt with Style.px
RN doc never mention point, but instead "density-independent pixels", also know as "logical pixels".
I guess we used `pt` because previous bindings where using `Pt`? 
Also I am proposing this because `pt` vs `pct` is harder to differentiate with `px` vs `pct`.

Thoughts @cknitt @sgny ?

MoOx:
To avoid brutal breaking change, I used `[@deprecated]` for pt. We can remove it  a bit later, because we know some people rely on this bindings already :)

Closes #530


MoOx:
Just so you know, I don't like `dp` & would have used `px` but I think it's better to use the proper unit name & I am just very new to the "dp" naming. I think RN docs should be more explicit about it and will try to make this unit name more visible there.

cknitt:
Exact same feelings here. Probably needs some time to get used to it. 😉 But it is the correct unit according to my understanding.

cknitt:
BTW I think for beginners

```reason
~margin=dp(10.)
```

might feel clearer/less "weird" than

```reason
~margin=10.->dp
```

MoOx:
I know it can feel weird a little, but to be honest, I think that, in comparison with web/normal css, it's a little bit more familiar as it "looks like" a unit:

```reason
~margin="10px" // reason-react css/web

~margin=10.->dp // intuition => "10dp"
~margin=dp(10.) // what, a function for a unit?
````

PR #542, opened 2019-06-09
Remove legacy bindings

sgny:
No longer needed, superseded by `bs-react-native-jsx3-compat`.


sgny:
As the example also depends on the legacy bindings, they are outdated.

MoOx:
I will try to update the example & convert it to reason-react-native.
I would prefer to wait a little to delete bs-react-native (as we already got minor fixes that aren't a big deal to handle)

PR #546, opened 2019-06-10
Add documentation for ImagePickerIOS API

sgny:
As promised...


reason-react-native/src/apis/ImagePickerIOS.md
```diff
@@ -1,47 +1,117 @@
 ---
 id: apis/ImagePickerIOS
 title: ImagePickerIOS
-wip: true
+officialDoc: https://facebook.github.io/react-native/docs/imagepickerios
```
MoOx: Link is necessary if it cannot be guessed from the component name :)

PR #547, opened 2019-06-12
Transformer Deg should be encoded to float instead of percentage.

bjornj12:
There a subtle bug because we are encoding Deg(45.) to percentage, making it crash in android.

The IOS native seems to be resilient enough to handle it and transform it to 45 degrees, but android JS crashes at runtime when you open a screen with rotate(Deg(..))

### DIFF
--- a/bs-react-native-jsx3-compat/src/Internals.re
+++ b/bs-react-native-jsx3-compat/src/Internals.re
@@ -16,5 +16,5 @@ module Encoder = {
 
   let pct = f => string(Printf.sprintf("%.2f%%", f));
 
-  let deg = f => string(Printf.sprintf("%.2f%%", f) ++ "deg");
+  let deg = f => string(Printf.sprintf("%.2f", f) ++ "deg");
 };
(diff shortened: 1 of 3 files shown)


sgny:
Your PR is indeed correct - it corresponds to the definition in the current binding - so well spotted. However, as the `bs-react-native`  "branch" in this repo is not going to continue for long, perhaps it's better to merge that to the `bs-react-native` repo.

@MoOx What do you think?

MoOx:
@sgny I am not sure what's the best scenario here. Because ideally we will have to ensure bs-react-native-jsx3-compat works too. But until reason-react-native is release, bs-react-native-jsx3-compat cannot really live elsewhere.
I guess it's probably simpler to keep bs-react-native here for a while & accept this PR & release this asap as a patch.

@bjornj12 is it too much to ask you to also update bs-react-native-jsx3-compat (or just check if that's not necessary)?

bjornj12:
Hey @sgny and @MoOx, 
Thanks for reviewing so fast.

I'm not entirely sure how i confirm the error in "bs-react-native-jsx3-compat", however the internals there had the same error in the encoding, and since it was a native android that failed, i assume it is the same there.

I verified this using: 
```
react-native: 0.59.9, 
bs-platform: 5.0.4, 
reason-React: 0.7.0,
```
I however hace not run it with "bs-react-native-jsx3-compat", just basic bs-react-native

PR #556, opened 2019-06-30
Add more navigation bindings with Twitter example

cem2ran:
This PR provides almost complete binding of the latest version of react-navigation (3.11.0) together with an example (expo) project that makes use of it to replicate the Twitter mobile application.

Twitter, for me seemed to be a good way to stress test the bindings due to its use of many of the Navigation styles found in React Navigation. Demo below demonstrates most of these:

ezgif com-resize

The initial goal of this set of bindings is to provide a low/zero-cost feature-complete bindings to react-navigation onto which one can build a more idiomatic Reason/reasonable set of bindings that makes use of language features.

The bindings includes one such example with a `Route` functor which can be included within a certain route screen module such that you're able to call `navigate` with the required params:

https://github.com/reasonml-community/reason-react-native/blob/ceea77b58fc45d54d4e2745206a74ead41598a69/reason-react-navigation-example/src/TwitterNavigation.re#L167-L179

https://github.com/reasonml-community/reason-react-native/blob/ceea77b58fc45d54d4e2745206a74ead41598a69/reason-react-navigation-example/src/TwitterNavigation.re#L289

I've been hesitant releasing this as I wanted to get it running in our production app first, which is using a much older version of these bindings, but we haven't had time to upgrade to JSX 3 yet, so here they are so everyone can consume them, test and iterate.

I have marked several places with **//TODO:** that are places that could use some attention or that I haven't spent enough time trying to solve/represent in Reason.

ReactNavigation.re should probably also be split up into individual files, but has been kept as one for now so you can simply consume the bindings by copy-pasting that file into your project.


reason-react-navigation-example/rn-cli.config.js
```diff
@@ -0,0 +1,13 @@
+// const path = require("path");
+
+// const config = {
+//   getProjectRoots() {
+//     return [path.resolve(__dirname, ".."), __dirname];
+//   },
+// };
+
+// module.exports = config;
```
MoOx: is it necessary to keep this comments?

reason-react-navigation-example/src/LoadingScreen.re
```diff
@@ -0,0 +1,50 @@
...
+    },
+    [|count|],
+  );
+
+  <>
+    <View
+      style=Style.(
+        style(~flex=1., ~justifyContent=`center, ~alignItems=`center, ())
+      )>
+      <Text> {React.string(count |> string_of_int)} </Text>
```
MoOx: Small detail but I found nice to have a more intuitive flow in jsx that would be `count->string_of_int->React.string`.
Maybe we could use more `->` when possible in this example?

reason-react-navigation-example/src/TwitterNavigation.re
```diff
@@ -0,0 +1,628 @@
...
+        source=Icons.compose
+      />
+    </TouchableOpacity>;
+};
+
+let placeholder = (~text, ~navigation: navigation) =>
+  <>
+    <Centered> <Text> {text |> React.string} </Text> </Centered>
+    <ComposeButton
+      onPress={_ => navigation->Navigation.navigate("Compose")}
```
MoOx: This seems a bit unsafe isn't it? What happen if you made a typo in the string? red screen?

reason-react-navigation-example/src/TwitterNavigation.re
```diff
@@ -0,0 +1,628 @@
...
+    [@react.component]
+    let make = (~navigation: navigation) =>
+      <>
+        <ReactNavigation.FlatList
+          data=dummyTweets
+          keyExtractor={({id}, _) => id}
+          renderItem={props =>
+            <TimelineItem
+              onPress={_ =>
+                navigation->Tweet.navigateWithParams({"tweet": props##item})
```
MoOx: Sound safe this way!

reason-react-navigation-example/src/TwitterNavigation.re
```diff
@@ -0,0 +1,628 @@
...
+              }
+              text={props##item.text}
+            />
+          }
+        />
+        <ComposeButton onPress={_ => navigation->ComposeRoute.navigate} />
```
MoOx: Here this seems better too.

reason-react-navigation-example/src/TwitterNavigation.re
```diff
@@ -0,0 +1,628 @@
...
+            )
+          ),
+      },
+      config(~contentComponent, ()),
+    );
+
+  navigator->setDynamicNavigationOptions(params => {
+    let navigation = params->NavigationParams.navigation;
+    let state = navigation->Navigation.state;
+    let routeName = getActiveRoute(state)->NavigationState.routeNameGet;
```
MoOx: We could simplify this patterns to (where `navigation` is retrieved & used once) via

```reason
let routeName = params
  ->NavigationParams.navigation
  ->Navigation.state
  ->getActiveRoute
  ->NavigationState.routeNameGet
```

reason-react-navigation-example/src/PrivacyPolicy.re
```diff
@@ -0,0 +1,15 @@
...
+open ReactNavigation;
+
+[@react.component]
+let make = (~navigation: navigation) =>
+  <Text> {React.string("Privacy Policy")} </Text>;
+
+make->setDynamicNavigationOptions(params => {
+  let navigation = params->NavigationParams.navigation;
+  let navigationOptions = params->NavigationParams.navigationOptions;
+  let screenProps = params->NavigationParams.screenProps;
```
MoOx: Could be commented to avoid unused warning?

reason-react-navigation-example/src/TwitterNavigation.re
```diff
@@ -0,0 +1,628 @@
...
+    SwitchNavigator.make({
+      "App":
+        StackNavigator.(
+          makeWithConfig(
+            {
+              "View": ContentView.navigator,
+              "Compose": Compose.make,
+              "Tweet": Tweet.make,
+              "Profile": Profile.make,
+            },
```
MoOx: Could we have in mind a safer way than strings to reference screens? I always found that totally unsafe when using rrnav in js...

cknitt:
@cem2ran Thanks a lot! :-) This is awesome! I like the nice example and your Route functor.

AFAIR your bindings are based on a very early version of mine. Unfortunately, mine have diverged in the meantime, too. I am not sure about the best course of action to re-unify the bindings again. The questions is also where we want to go (cc @MoOx):

1. bs.obj vs. Js.t vs. bs.abstract: In the bs-react-native bindings, we mostly used bs.obj for "write-only" stuff (things we pass into the API calls,  like config objects) and Js.t for "read-only" stuff (events/callback params). I think we should keep it this way as far as possible.

2. Using "write-only" objects with bs.obj allows us to flatten the module structure somewhat because we do not have to fear name clashes for accessors anymore if we don't have accessors. I think that's a good thing.

3. screenProps: I basically made the bindings a functor in the last version to be able to have the same screenProps type everywhere. Not sure if that was a good idea, but I do not like having `Js.t({..})` for the screenProps everywhere.

4. In my latest version, I have a nice AppContainer functor (or actually function returning a first-class module) that I would like to keep.

5. Are you sure your usage of `bs.string` is actually working? E.g.

```reason
type tabBarPosition = [@bs.string] [ | `top | `bottom];
```

gives me `Warning 101: Unused BuckleScript attribute: bs.string`. Also `type navigationAction` etc.

If we can agree where we want to go, maybe the best course of action would be for me to take your latest version of the bindings, reapply my changes and whatever we agree on in this discussion, test in our current project and then submit a PR with the combined version?

cem2ran:
@MoOx Thanks for the feedback! I'll make the changes you proposed. I think I'll split up into separate files per module once we have a more clear direction / I've addressed @cknitt comments.

@cknitt Yeah I did not realize you had updated the bindings. 🙀

1. Agree.

2. 👍

3. I have been considering this as well. I was just not quite sure about `screenProps` and if they're static or can be different in subparts of your application. As far as I can tell you can add them in the AppContainer e.g. https://reactnavigation.org/docs/en/localization.html#wiring-up-your-localization-library-to-navigation, but also in StackNavigator, where I'm not sure if you replace or merge with parent navigators... https://reactnavigation.org/docs/en/stack-navigator.html#navigator-props

4. I'll re-add this as mentioned didn't realize you added this. While I do think this is maybe a bit too advanced OCaml I think it might be fine as it is only written once in your application. Would be afraid to scare off people or increase the barrier of entry, but it's fine if you just need to copy paste this snippet of code once in your applications lifetime.

5. I'll double check. There's actually quite a few `unused` warnings that I'm a bit confused about, including bs.optional. Not sure if you have run into similar in the reason-react-native codebase?

sgny:
Great contribution, I found `react-navigation` rather underwhelming (and buggy) somewhat over a year ago, but with kmagiera's `gesture handler` and `screens` perhaps it's time to reevaluate.

> 5. I'll double check. There's actually quite a few `unused` warnings that I'm a bit confused about, including bs.optional. Not sure if you have run into similar in the reason-react-native codebase?

No warnings on the `reason-react-native` codebase as any warning is turned into an error in `bsconfig.json` with `"bsc-flags": ["-warn-error @a"]`. I have a PR #557 to make an exception, though.

I do have one suggestion, now that we have several bindings being added. If we move other bindings to their own repos under `reasonml-community`, we could simply instruct people to replace `react-native-community` with `reasonml-community` when installing packages with `npm` or `yarn`.

MoOx:
> I do have one suggestion, now that we have several bindings being added. If we move other bindings to their own repos under reasonml-community, we could simply instruct people to replace react-native-community with reasonml-community when installing GitHub packages with npm or yarn and for packages use the @reasonml-community scope or something like that.

I created a npm org reason-react-native so we could publish bindings as `@reason-react-native/something`. Exception made with react-navigation that doesn't have -native in its name.

sgny:
> > I do have one suggestion, now that we have several bindings being added. If we move other bindings to their own repos under reasonml-community, we could simply instruct people to replace react-native-community with reasonml-community when installing GitHub packages with npm or yarn and for packages use the @reasonml-community scope or something like that.
> 
> I created a npm org reason-react-native so we could publish bindings as `@reason-react-native/something`. Exception made with react-navigation that doesn't have -native in its name.

Then the bindings would be would be named identically to the JS package, is that right?

MoOx:
> Then the bindings would be would be named identically to the JS package, is that right?

We should continue discussion there https://github.com/reasonml-community/reason-react-native/issues/474#issuecomment-507967698

cknitt:
Today I tried to start with my original idea to take @cem2ran's version of the code and reapply my changes.

However, I found that these are too many changes at once for me. Also, with some changes/additions, I am not confident that they actually work. Especially with all those unused attributes in there - the corresponding functions/types can never have worked/have been never tested.

I found these four types of unused attribute errors:

1. `[@bs.abstract light = true]`: As far as I can see, `bs.abstract` does not exist and so this has no effect. `[@bs.deriving {abstract: light}]` would be correct.

2. Things like

```reason
type backBehavior = [@bs.string] [ | `none | `initialRoute | `history | `order];
```

`bs.string` only works directly on a function parameter in the `external` declaration. It won't work if you set in on a type beforehand and use that type in the `external` declaration. Same with

```reason
type navigationTabAction = [ | `NavigationInitAction | `NavigationNavigateAction | `NavigationBackAction];
```

etc.

3. `[@bs.optional]` outside of bs.deriving abstract.

4. `[@bs.unwrap]` outside of external function declarations.

So I think this needs some more work, and we should better apply smaller, incremental improvements to the current code. Do you agree, @MoOx?

@cem2ran would it be possible for you to submit smaller PRs that just add missing fields (or new modules/types/functions) to the existing code?

cem2ran:
@cknitt I'm reworking the bindings based on your feedback. Will have some time to finish this up during the weekend. I'll make sure to correct the bindings that are incorrect/untested.

I can submit separate PRs for the individual modules/navigators that exists in ReactNavigation.re. 

Do you think this would be better and make it easier to review?

cknitt:
@cem2ran Thanks a lot! :-) Yes, I think separate PRs for the individual modules would make review much easier and help us to get your contributions merged faster.

reason-react-navigation/src/ReactNavigation.re
```diff
@@ -319,6 +1140,596 @@ module Make = (S: {type screenProps;}) => {
...
+  let useNavigation = () => React.useContext(navigationContext);
+};
+
+/* TODO: Maybe we can do this smarter without declaring everything from scratch.
+   We just need to state that the bs.module is `react-navigation` instead of `react-native`
+   */
+module FlatList = {
+  include ReactNative.FlatList;
+
+  [@react.component] [@bs.module "react-navigation"]
```
cem2ran: @MoOx Is there a way to achieve this without having to copy-paste the bindings from react-native?

reason-react-navigation-example/src/TwitterNavigation.re
cem2ran: Absolutely. Currently still in the "making it work" phase :) Will look at this next.

cem2ran:
@cknitt I've re-introduced `makeAppContainer` and resolved most if not all the points you raised:

Now im mostly using `Js.t` for props/read-only and `bs.obj` for "write-only" objects.

I believe the usage of bs.string, bs.optional and bs.unwrap are now correct (used inside external function declarations).

I've extracted and implemented bindings for `NavigationActions` and also added `MaterialTopTabBar` & `BottomTabBar`.

Next-up I'll be looking into extracting the modules out and tackle: https://github.com/reasonml-community/reason-react-native/pull/556#discussion_r298933038

reason-react-navigation/src/ReactNavigation.re
MoOx: It seems you changed to include, which is exactly what you where looking for :)

MoOx:
> So I think this needs some more work, and we should better apply smaller, incremental improvements to the current code. Do you agree, @MoOx?

Yeah I guess so.

Maybe in the first place we should split the ReactNavigation.re as multiple files & use a bs custom namespace (so api is still `ReactNavigation.*`, like for reason-react-native being available as `ReactNative.*`).

I can split this in master branch so you can rebase/adjust your pr (or split into multiple prs) to review is more easily?

reason-react-navigation/src/ReactNavigation.re
cem2ran: I think I did this previously as well. I essentially need the exact same `makeProps` as the React Native binding, which I guess I can include from ReactNative.FlatList, but I need to override `make` such that it's importing from [@bs.module "react-navigation"] instead of [@bs.module "react-native"]. 

Is there a way I can reference `FlatList.props` that go into `FlatList.make`, such that I could do:

```reason
[@bs.module "react-navigation"]
external make: (FlatList.props) => FlatList.element = "FlatList";
```

or is there a way I could override bs.module?

MoOx: Not sure you can easily do that. We already have to do some copy pasting to achieve the zero-cost in lots of place.
Copy pasting is not a big deal as soon it's documented somewhere. You can probably copy paste & add this kind of comment https://github.com/reasonml-community/reason-react-native/blob/9327dd58a5c56659259f43a78e57f2069b259793/reason-react-native/src/components/FlatList.re#L17

cknitt:
@cem2ran Great! 👍

@MoOx That sounds great! @cem2ran Fine with you, too?

reason-react-navigation/src/ReactNavigation.re
cem2ran: Copy paste is ok. To my understanding it seems we're only missing a type for the props we can reference, which react-ppx doesn't appear to give us. I might look at adding this if it's missing.

cem2ran:
@cknitt @MoOx Yup all good! 👍 Thanks.

MoOx:
I started this locally, I except to publish the react navigation split next week.

MoOx:
@cem2ran you can have a look to the split I started to work on https://github.com/reasonml-community/reason-react-native/pull/563

MoOx:
@cem2ran I merged the split. Do you want to rebase your work on this now?

idkjs:
Now that the namespace as been set, do we still install from git using instructions on the readme? Also, if we are installing from git do we have to install both reason-react-native and reason-react-navigation from git? Or do we npm install -native then install -navigation from git?

Thank you, gentlemen.

MoOx:
You can use one from npm & another from git. Should work. I just updated the trick to support reason-react-navigation as the string replacement wasn't enough before.

MoOx:
TIL https://twitter.com/satya164/status/1159390789774036992

So maybe we should try to focus on low level bindings for packages right away instead of "wasting" time on this bindings which are (quickly?) going to be deprecated

cknitt:
@MoOx Ok, that's interesting. I will look into it when I have some time. I have also started trying to get Cem's Twitter example to run with what we currently have in master. I hope to make some progress until the end of next week if my workload allows it.

sgny:
I doubt there will be a change in API. It's probably only going to require adding various individual packages as dependencies (`react-navigation` might be pulling those already) and referring to those with the `@bs.module` annotation.

MoOx:
Yeah exactly. I am currently working on a JS/Flow app (sad moox) and I just basically changed some ref from multiple packages 

- `@react-navigation/core`
- `@react-navigation/native`
- `react-navigation-tabs`
- `react-navigation-stack`
- `react-navigation-drawer`
- etc

What is definitely more interesting is THIS (said to be type safe friendly) https://github.com/react-navigation/navigation-ex
Alpha will be published [soon](https://twitter.com/satya164/status/1159395496743624704) 😋

sgny:
> Yeah exactly. I am currently working on a JS/Flow app (sad moox) and I just basically changed some ref from multiple packages

Bon courage, JS is mostly read-only for me :)
 
> What is definitely more interesting is THIS (said to be type safe friendly) https://github.com/react-navigation/navigation-ex
> Alpha will be published [soon](https://twitter.com/satya164/status/1159395496743624704) 😋

That does sound promising, hopefully it will reach production status.

cknitt:
@cem2ran I adapted your example to the latest master, see #573.

PR #557, opened 2019-07-02
Disable warning 105 (inferred JS names in external definitions)

sgny:
In recent BuckleSscript releases, warning 105 has been defined for inferred JS names in external definitions. For more information, you may refer to the relevant [post](https://bucklescript.github.io/blog/#a-new-warning-number-105) on their blog.

Thus, a definition such as
```
[@bs.module "react-native"] [@bs.scope "Easing"] external linear: t = "";
```
leads to a warning that reads

> the external name is inferred from val name is unsafe from refactoring when changing value name

We could go through the codebase to specify the JS name for all such definitions (which would be almost the entire codebase) or we could revise `bsconfig.json` to ignore warning 105.

I did not take the high road...

### DIFF
--- a/reason-react-native/bsconfig.json
+++ b/reason-react-native/bsconfig.json
@@ -16,6 +16,6 @@
       "subdirs": true
     }
   ],
-  "bsc-flags": ["-bs-no-version-header", "-warn-error @a"],
+  "bsc-flags": ["-bs-no-version-header", "-warn-error @a", "-w @a-105"],
   "bs-dependencies": ["reason-react"]
 }


sgny:
> What do you think about using a space like the item before so we more easily understand that they go together?
> 
> `"-w @a-105"`
> 
> Btw, does -w == -warn-error?

That's equivalent, but has the benefit of making it difficult to add a flag in-between `-w` and `@a-105`, so I'll revise the PR.

`-warn-error` is to convert warnings into errors. We could do "-warn-error @a-105" but that would simply mean all warnings but 105 would be turned into errors and warning 105 would still be thrown.

`-w` is used to define the list of warnings and seems to take precedence. `a` is for all warnings and as above `@a-105` simply removes 105 from the list of warnings.

MoOx:
Thanks for the explanation!

PR #559, opened 2019-07-19
Update README.md for Reason Conf

dylanirlbeck:
@peterpme


MoOx:
Could you give an explanation about the motivation to put this at this specific location?
I am not against it but you could at least say something about it...

dylanirlbeck:
Hey!

I'd like to do my part of spreading the word about Reason Conf and thought that this would be a great way to show it! I'm a huge fan of your reason-react-native and was hoping you'd be open to it!

Thank you :relaxed:

MoOx:
Not a fan of the idea of a message at the very top of the ready. Maybe instead we could add a message on the homepage of the website?

@cknitt @sgny any thoughts?

cknitt:
Fine with me to add it on the website!

MoOx:
I have updated the thing just a little, will go with that.

PR #560, opened 2019-07-22
Add AppRegistry.registerHeadlessTask

cknitt:
This PR adds the `AppRegistry.registerHeadlessTask` function.

There are more AppRegistry functions missing though, see https://facebook.github.io/react-native/docs/appregistry. These are probably rarely used, but I will file an issue.


cknitt:
Personally, I do not really need a patch release immediately for this, and I am sure it's quite a bit of work for you, so unless the community asks for it, I would wait some more before doing another release.

PR #563, opened 2019-07-26
ReactNavigation split WIP

MoOx:
(Sorry I didn't have much time this days)

This is just a start that going to help us to see where we wanna go.

(Just to let you know, I only used react-navigation once with js, and it was v1 - but I think it didn't changed that much from what I saw on docs)

One question I have: how are we supposed to define screenProps per screen?

I were using js+flow when using react-navigation 1 & here are things that I disliked

- no way to validate that your are going to navigate to a screen that exist (string to navigate)
- no way to know if the screenProps you are going to pass are valid

Not sure how we can address that with reason-react-navigation as I only see on screenProps passed to main functor.

Did I miss something?


MoOx:
I just saw that react-navigation is split into multiple repos https://github.com/react-navigation. 
Do you think that's a good idea for us to first make binding for stack, tabs etc or directly continue to the main interface described here https://github.com/react-navigation/react-navigation/blob/master/src/react-navigation.js ?

cknitt:
screenProps: I have not found a good way to do this on a per-screen basis. For our purposes, defining the type globally has been sufficient so far.

Multiple repos: I think binding to the main interface/public API as documented under https://reactnavigation.org/docs/en/api-reference.html is sufficient.

cknitt:
I guess we need to handle the screenProps differently anyway.

Now that the modules are split into multiple files, we have multiple functors for the screenProps instead of a single outer one, so the API is less convenient and it is not ensured anymore that the screenProps type is the same everywhere.

I could not think of a good solution yet though. Any ideas are welcome!

MoOx:
Actually I only see a single functor for the drawer (in addition to the main one).

One thing that bother me just a little: I activated the custom namespace, so the modules are available under `ReactNavigation.*` but we still have `ReactNavigation.ReactNavigation.Make`. Do you see an idea to avoid this ultra verbose init call?

cknitt:
Ok, yes, "multiple" = two at the moment. (But maybe the screenProps are still missing in other places?)

Still, the functors are inconvenient now. Maybe we should get rid of them? Not sure how complicated it will get if `'screenProps` is a type parameter everywhere. I think I had that before and didn't like it.

In any case, I'd say let's merge and then experiment further regarding the functors/screenProps.

MoOx:
Yeah I was thinking about `'screenProps` too.

I have to admit that without actually using this bindings, I am not the best person to advice anything (currently I had to go back with a client with js+flow 😥 because I don't find freelance missions this days with reasonml)

I guess the twitter example made by @cem2ran will be a good place to play with.

PR #564, opened 2019-07-26
Docs for AppState.re

idkjs:
Please take a look at this. As noted at https://react-native.org/doc/appstate.html, this wont ever do anything. Also, the `Js.log` statement in `handleAppStateChange` never seems to fire. I don't believe that is normal or we should probably not put it out there that way.

This is the output I am getting 

```reason Remote debugger is in a background tab which may cause apps to perform slowly. Fix this by foregrounding the tab (or opening it in a separate window).
- node_modules/expo/build/logs/LogSerialization.js:146:14 in _captureConsoleStackTrace
- node_modules/expo/build/logs/LogSerialization.js:41:24 in serializeLogDataAsync$
- ... 9 more stack frames from framework internals

Finished building JavaScript bundle in 1114ms.
Finished building JavaScript bundle in 102ms.
Finished building JavaScript bundle in 39ms.
Finished building JavaScript bundle in 492ms.
Running application on iPhone X.
active
```
Here is the demo repo if you want to run it. https://github.com/idkjs/reason-appstate-demo.

### DIFF
--- a/reason-react-native/src/apis/AppState.md
+++ b/reason-react-native/src/apis/AppState.md
@@ -42,28 +42,23 @@ open ReactNative;
 
 [@react.component]
 let make = () => {
-  let (appState, setAppState) = React.useState(() => AppState.currentState);
-  Js.log(appState);
+  let (appState, setAppState) = React.useState(_ => AppState.currentState);
 
-  let handleAppStateChange = (nextAppState: AppState.t) => {
-    Js.log2("nextAppState: ", nextAppState);
-    Js.log2("appState: ", appState);
-    switch (appState,nextAppState) {
-    | (_background, _active) =>
-      Js.log(
-        "App has come to the foreground!",
-      );
-    | (_inactive, _active) =>
-      Js.log(
-        "App has come to the foreground!",
-      );
+  let handleAppStateChange = nextAppState => {
+    switch (appState, nextAppState) {
+    | (_, state) when state === AppState.background =>
+      Js.log("App has come to the background!")
+    | (_, state) when state === AppState.active =>
+      Js.log("App has come to the foreground!")
     | _ => ()
     };
     setAppState(_ => nextAppState);
   };
 
   React.useEffect(() => {
-    AppState.addEventListener(`change(state => handleAppStateChange(state)))
+    AppState.addEventListener(
+      `change(state => handleAppStateChange(state)),
+    )
     |> ignore;
     Some(
       () =>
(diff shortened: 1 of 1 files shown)


MoOx:
The example you have provided currently doesn't compile for several mistakes
Here is one that should work

```re
[@react.component]
let make = () => {
  let (appState, setAppState) = React.useState(_ => AppState.currentState);

  let handleAppStateChange = nextAppState => {
    switch (appState, nextAppState) {
    | (_, state) when state === AppState.background =>
      Js.log("App has come to the background!")
    | (_, state) when state === AppState.active =>
      Js.log("App has come to the foreground!")
    | _ => ()
    };
    setAppState(_ => nextAppState);
  };

  React.useEffect(() => {
    AppState.addEventListener(`change(state => handleAppStateChange(state)))
    |> ignore;
    Some(
      () =>
        AppState.removeEventListener(
          `change(state => handleAppStateChange(state)),
        ),
    );
  });

  let renderAppState =
    switch (appState) {
    | appState when appState === AppState.active => "active"
    | appState when appState === AppState.background => "background"
    | appState when appState === AppState.inactive => "inactive"
    | _ => "unknown"
    };
  <Text> {"Current state is: " ++ renderAppState |> React.string} </Text>;
};
```

@cknitt @sgny what do you think about having X.example.re in the repo, next to implementation & .md to be sure that example provided are up to date, formatted & valid (compiled).

sgny:
> @cknitt @sgny what do you think about having X.example.re in the repo, next to implementation & .md to be sure that example provided are up to date, formatted & valid (compiled).

Having standalone examples would be helpful and more convenient, however, it might be better to have them in a separate directory as organisation by file type would not be sufficient and they would get compiled when clean building a project.

idkjs:
@MoOx Merci, brother. By not compile i suppose you meant the warning messages. Is that correct? Just trying to understand the language since the project ran when before i posted it. So warnings = doesnt compile.  

Thanks for the feedback. Very helpful.

I updated the the AppState.md pull and the example repo with your code if it matters.

cknitt:
@

> > @cknitt @sgny what do you think about having X.example.re in the repo, next to implementation & .md to be sure that example provided are up to date, formatted & valid (compiled).
> 
> Having standalone examples would be helpful and more convenient, however, it might be better to have them in a separate directory as organisation by file type would not be sufficient and they would get compiled when clean building a project.

I agree. Such examples would be convenient, but should be put into a separate directory.

cknitt:
I just discovered that there is a bug in the `AppState` binding - addEventListener actually returns `unit`, not `EventSubscription.t`. I am preparing a PR for that, this one will need to be rebased afterwards.

cknitt:
See #568.

idkjs:
@cknitt does @MoOx example above need to be updated to reflect that?

cknitt:
Yes, the “ignore” won’t be needed anymore.

idkjs:
I will wait on that update to go through before making this doc change.

MoOx:
@idkjs I actually got warnings indeed but those warnings would have revealed errors on runtime (not fatal error, but more behavior errors)

cknitt:
@idkjs Can you make the requested change (remove the `ignore`) so that we can get the PR merged?

MoOx:
Closing in favor of #589.

Issue #561, opened 2019-07-22
AppRegistry bindings incomplete


cknitt:
AppRegistry bindings are incomplete.

We currently have `registerComponent`, and #560 adds `registerHeadlessTask`, but there are more functions missing, see https://facebook.github.io/react-native/docs/appregistry.

sgny:
Thanks for adding registerHeadlessTask.

I’ve had incomplete bindings to AppRegistry on hand for a very long time now, I’ll see if I can dig them out.

This API is rather poorly documented, I gave up because at least a few of them didn’t seem to work properly, at least with iOS.

sgny:
OK, I still have them. I could not check before as I was on my phone.

I'll prepare a PR and we'll proceed from there.

cknitt:
@sgny Great! Thanks a lot! 👍

MoOx:
Closed by #567

PR #567, opened 2019-07-28
add missing AppRegistry functions

sgny:
Related to #561, these are almost complete, if not quite. What are still missing:
- `setComponentProviderInstrumentationHook`
- `createPerformanceLogger` (which is needed to create the hook)

(I still need to figure out how to bind to `createPerformanceLogger`, perhaps in a later PR?)

Some of the functions take an argument of type `Function`; that argument is passed `appParameters` (which is of type `any`), therefore I decided to define `Function` to be of type `appParameters => unit` and added a function `asAppParameters` to cast any type as `appParameters`. Hopefully that should be flexible enough. I don't believe the return type of `Function` matters at all, hence I chose to specify `unit` instead of some abstract type.

### DIFF
--- a/reason-react-native/src/apis/AppRegistry.md
+++ b/reason-react-native/src/apis/AppRegistry.md
@@ -5,6 +5,11 @@ wip: true
 ---
 
 ```reason
+type appKey = string;
+type section = string;
+type taskId = float;
+type taskKey = string;
+
 type task('data) = 'data => Js.Promise.t(unit);
 type taskProvider('data) = unit => task('data);
 type taskCanceller = unit => unit;
@@ -39,18 +44,18 @@ type runnable('a) = {
 
 type registry('a) = {
   .
-  "sections": array(string),
+  "sections": array(section),
   "runnables": Js.Dict.t(runnable('a)),
 };
 
 [@bs.module "react-native"] [@bs.scope "AppRegistry"]
-external getAppKeys: unit => array(string) = "";
+external getAppKeys: unit => array(appKey) = "";
 
 [@bs.module "react-native"] [@bs.scope "AppRegistry"]
 external getRegistry: unit => registry('a) = "";
 
 [@bs.module "react-native"] [@bs.scope "AppRegistry"]
-external getRunnable: string => Js.Nullable.t(runnable('a)) = "";
+external getRunnable: appKey => option(runnable('a)) = "";
 
 [@bs.module "react-native"] [@bs.scope "AppRegistry"]
 external getSectionKeys: unit => array(string) = "";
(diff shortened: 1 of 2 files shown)


sgny:
I caught a mistake with the return type of the `registerComponentWithSection` function, it should have been `unit` as with `registerComponent`. I also changed the second argument to be `componentProvider('a)` in line with the React Native code and for better consistency with the sister function.

```reason
// multiple externals
[@bs.module "react-native"] [@bs.scope "AppRegistry"]
external registerComponent: (string, componentProvider('a)) => unit = "";

// multiple externals
[@bs.module "react-native"] [@bs.scope "AppRegistry"]
external registerComponentWithSection:
  (string, componentProvider('a), string) => unit =
  "registerComponent";
```

reason-react-native/src/apis/AppRegistry.md
```diff
@@ -5,14 +5,103 @@ wip: true
...
+};
+
+[@bs.module "react-native"] [@bs.scope "AppRegistry"]
+external getAppKeys: unit => array(string) = "";
+
+[@bs.module "react-native"] [@bs.scope "AppRegistry"]
+external getRegistry: unit => registry('a) = "";
+
+[@bs.module "react-native"] [@bs.scope "AppRegistry"]
+external getRunnable: string => Js.Nullable.t(runnable('a)) = "";
```
cknitt: I think this should only return `undefined`, not `null`, so we could use `option` instead of `Js.Nullable.t`.

reason-react-native/src/apis/AppRegistry.md
```diff
@@ -5,14 +5,103 @@ wip: true
...
 [@bs.module "react-native"] [@bs.scope "AppRegistry"]
-external registerComponent: (string, unit => React.component('a)) => unit =
+external registerSection: (string, componentProvider('a)) => unit = "";
+
+[@bs.module "react-native"] [@bs.scope "AppRegistry"]
+external runApplication: (string, 'a) => unit = "";
+
+[@bs.module "react-native"] [@bs.scope "AppRegistry"]
+external setWrapperComponentProvider:
+  Js.Nullable.t(wrapperComponentProvider('a, 'b)) => unit =
```
cknitt: Why is this nullable?

reason-react-native/src/apis/AppRegistry.md
```diff
@@ -5,14 +5,103 @@ wip: true
...
   "";
 
-type task('data) = 'data => Js.Promise.t(unit);
+[@bs.module "react-native"] [@bs.scope "AppRegistry"]
+external unmountApplicationComponentAtRootTag: string => unit = "";
 
+// Android only
 [@bs.module "react-native"] [@bs.scope "AppRegistry"]
-external registerHeadlessTask: (string, unit => task('data)) => unit = "";
+external cancelHeadlessTask: (float, string) => unit = "";
```
cknitt: As it is unclear what the float and string params are here, we could consider using labelled args (`~taskId`, `~taskKey`).

(Same for the other task functions below.)

sgny: Would type aliases like `type taskId = float` and `type taskKey = string` help?

reason-react-native/src/apis/AppRegistry.md
sgny: `wrapperComponentProvider` is initiated with
```
let wrapperComponentProvider: ?WrapperComponentProvider;
```
I wasn't sure if setting that `null` might be a valid use pattern, so defined that as nullable.

On second though, it probably not needed at all.

reason-react-native/src/apis/AppRegistry.md
sgny: You're right. I was probably too conservative, mapping `?` to `undefined | null` by default, but since this is looking up a value in a hash map, `option` should be enough.

sgny:
@cknitt Thanks for the comments. I'll revise the PR accordingly but that will probably be later tonight.

sgny:
I defined type aliases instead of labelled arguments, as I believe this approach offers clarity without verbosity. I am certainly open to feedback on this.

I kept `taskId` as a float, but I suppose it should be an `int`. Redefine it as an `int` and switch to a `float` if anyone runs into problems, or just leave it as it is?

PR #573, opened 2019-08-16
Cem's Twitter example adapted to latest master.

cknitt:
I adapted @cem2ran's Twitter example app to the latest master.

After the split up of ReactNavigation.re into individual modules by @MoOx there were a few things I had to adapt in the bindings. I got rid of the functors except for `AppContainer`.

Functionality-wise, I added to the bindings only what was required for Cem's example to run, improving a few things like `NavigationOptions.HeaderTitle` etc. I did not use Cem's Route functors. While interesting/useful, I think these should go into a separate module in a separate PR.

The rest of Cem's improvements to the bindings can be taken over step by step as well.


cknitt:
BTW, this is an Expo project as created by Cem. I am not particularly fond of Expo and would prefer a plain vanilla React Native project, but we can still change that in a future PR.

cknitt:
Also, I think `TwitterNavigation.re` is huge and should probably be split up into separate modules later.

PR #577, opened 2019-08-20
reason-react-native: add example for ScrollView scrollTo method

MoOx:
totally wip but better than nothing?

### DIFF
--- a/reason-react-native/src/components/ScrollView.md
+++ b/reason-react-native/src/components/ScrollView.md
@@ -8,32 +8,58 @@ wip: true
 
 ### `scrollTo`
 
+Scrolls to a given `x`, `y` offset, either immediately, with a smooth animation, or,
+for Android only, a custom animation duration time.
+
 ```reason
+let scrollParams = (~x: float, ~y: float, ~animated: bool=?, ~duration: float=?, unit) => unit;
+
+let scrollTo = (scrollView, scrollParams) => unit;
+```
+
+#### `scrollTo` Example
+
+```reason
+open ReactNative;
+
 [@react.component]
-let make = () => {
+let make2 = () => {
   let scrollViewRef = React.useRef(Js.Nullable.null);
   <ScrollView ref=scrollViewRef>
     <TouchableOpacity
-      onPress={_ => {
-        scrollViewRef
-        ->React.Ref.current
-        ->Js.Nullable.toOption
-        ->Option.map(scrollView =>
+      onPress={_ =>
+        switch (scrollViewRef->React.Ref.current->Js.Nullable.toOption) {
+        | Some(scrollView) =>
           scrollView->ScrollView.scrollTo(
-            ScrollView.scrollToParams(
-            ~x=0,
-            ~y=0.,
-            ~animated=true,
-            (),
-          ),
-        )
-      )
-      ->ignore;
-    }}>
-    <Text> "ScrollTo 0, 0"->React.string </Text>
+            ScrollView.scrollToParams(~x=0., ~y=0., ~animated=true, ()),
+          )
+        | _ => ignore()
+        }
+      }>
+      <Text> "ScrollTo 0, 0"->React.string </Text>
     </TouchableOpacity>
-  </ScrollView>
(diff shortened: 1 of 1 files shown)


reason-react-native/src/components/ScrollView.md
```diff
@@ -4,6 +4,41 @@ title: ScrollView
...
+[@react.component]
+let make = () => {
+  let scrollViewRef = React.useRef(Js.Nullable.null);
+  <ScrollView ref=scrollViewRef>
+    <TouchableOpacity
+      onPress={_ => {
+        scrollViewRef
+        ->React.Ref.current
+        ->Js.Nullable.toOption
+        ->Option.map(scrollView =>
```
cknitt: I am not sure I like the usage of `Option.map` here. Unfortunately Belt does not have `Option.apply`. So maybe just use a `switch`? (Speaking of Belt - you should actually it open it.)

MoOx:
I used a switch for the example.

Btw, I noticed that scrollTo have `scrollToParams` & that `scrollToEndWithOptions` have `scrollToEndOptions`. What do you think about normalising naming with Params or Options (& same for similar places?).
We could normalize & deprecate with warnings & remove deprecating warning in next major release?

MoOx:
I used a switch for the example.

Btw, I noticed that scrollTo have scrollToParams & that scrollToEndWithOptions have scrollToEndOptions. What do you think about normalising naming with Params or Options (& same for similar places?).

@cknitt @sgny We could normalize & deprecate with warnings & remove deprecating warning in next major release?

cknitt:
I think I used the different names on purpose. In the case of `scrollTo`, we have a *required* `params` object (similar to the props object of a React component). In the case of `scrollToEnd`, we have an *optional* `options` object that may be supplied as an additional parameter to override some default values.

Not sure if this is consistent across the code though...

reason-react-native/src/components/ScrollView.md
```diff
@@ -4,6 +4,67 @@ title: ScrollView
...
+  let scrollViewRef = React.useRef(Js.Nullable.null);
+  <ScrollView ref=scrollViewRef>
+    <TouchableOpacity
+      onPress={_ =>
+        switch (scrollViewRef->React.Ref.current->Js.Nullable.toOption) {
+        | Some(scrollView) =>
+          scrollView->ScrollView.scrollTo(
+            ScrollView.scrollToParams(~x=0., ~y=0., ~animated=true, ()),
+          )
+        | _ => ignore()
```
cknitt: `| _ => ()` should be sufficient.

MoOx:
@cknitt I just find confusing the 2 namings. I get the idea.
But maybe we could stick to Option, since we might have xxx & xxxWithOptions ? (and xxxOptions would be the method to generate options).
Not sure if naming things option/param/argument are really different & clear for users. We should imo keep one & stick to it.

@sgny any comment?

cknitt:
I am open to changing these method args from "params" to "options".

Callback params like `VirtualizedList.onEndReachedParams` would remain as they currently are though, right?

sgny:
I think the function name should remain `scrollToEndWithOptions` as that is consistent across the various modules. However, the optional argument could be renamed `scrollToEndParams` which would be more consistent with `scrollParams` and still be a reasonable name for what it is.

cknitt:
I find `scrollToEndWithOptions` with `scrollToEndParams` even more confusing.

I'd expect `xxxWithYyy` to take a `yyy` param...

MoOx:
I would go for just ditching one & keep only the other for simplicity.
Ditch "Params" & use "Options" everywhere. Or the opposite.
Having this kind of naming is anyway "confusing" somehow, so we should just have one.
Migration won't be a problem: we can deprecate call easily, people will have warnings, & next major release we remove them.

cknitt:
If we want to unify this, I am for "Options" everywhere.

sgny:
How about instead of `withOptions` we use `withOptional` (or `withOptionals`)? Then we could have optional parameters or configs without causing much confusion.

PR #580, opened 2019-08-20
Use itemSeparatorComponent in flatList

tatchi:
Thanks for this amazing work, so cool to be able to use ReasonML for building mobile apps :heart_eyes:

It's probably a matter or preferences, but I moved the horizontal separator used in the `Timeline` component from the `TimelineItem` to the `TimelineList` itself (using the `_ItemSeparatorComponent` prop).

Since this prop is specifically used for this purpose, I thought it would probably be a good idea to use it. One "nice" little improvement is that by using this prop, the separator is not displayed below the last item of the list (it's handled automatically by the `FlatList`).

### DIFF
--- a/reason-react-navigation-example/src/components/Timeline.re
+++ b/reason-react-navigation-example/src/components/Timeline.re
@@ -4,20 +4,13 @@ open Helpers;
 module TimelineItem = {
   [@react.component]
   let make = (~text, ~onPress) =>
-    <>
-      <TouchableOpacity
-        onPress
-        style=Style.(
-          style(~paddingHorizontal=16.->dp, ~paddingVertical=26.->dp, ())
-        )>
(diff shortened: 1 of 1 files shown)


cknitt:
👍  for using `_ItemSeparatorComponent`.

However, I think that final separator below the last item should actually be there. At least on iOS, not sure about Android. `_ListFooterComponent` could be used for that though.

tatchi:
> 👍 for using `_ItemSeparatorComponent`.
> 
> However, I think that final separator below the last item should actually be there. At least on iOS, not sure about Android. `_ListFooterComponent` could be used for that though.

I added a `_ListFooterComponent ` separator. It seems to add a separator below the last item on iOS but not on Android.

cknitt:
Great, thanks a lot!

PR #582, opened 2019-08-20
Add bindings to react-native-netinfo

sgny:
I have moved the bindings over from my repo, renamed the module and updated documentation accordingly.

I believe they are ready to be published but the version shields should probably be removed for that.

### DIFF
--- a/@reason-react-native/netinfo/README.md
+++ b/@reason-react-native/netinfo/README.md
@@ -1,0 +1,277 @@
+# BuckleScript bindings to React Native NetInfo
+
+[![Version](https://img.shields.io/npm/v/@reason-react-native/netinfo.svg)](https://www.npmjs.com/@reason-react-native/netinfo)
+
+These are BuckleScript bindings to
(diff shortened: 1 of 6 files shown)


cknitt:
Great work! 👍 

General question (/cc @MoOx): If we add `reason-react-native-xxx` to the monorepo, containing bindings for `react-native-xxx`, do we want the namespace to be `ReactNativeXxx` or just the shorter form `RNXxx`?

MoOx:
I would go for this rule

- folder & npm name: `@reason-react-native/xxx`
- bsconfig custom namespace: `ReactNativeXxx` or `ReactNative_Xxx`

RN is easy to think about but I prefer verbosity, which helps discoverability (even if RN might seems obvious)

(Bonus on `@reason-react-native/*` npm ns: we can add people to the org on npm & we directly have access admin rights to all packages under this scope)

MoOx:
Merging, I will propose a few changes to the README & we could release in a few

PR #583, opened 2019-08-22
revise documentation for YellowBox API

sgny:
@cknitt already included an example and the documentation was quite complete, but I felt it might be helpful to flesh out the documentation a bit further.

### DIFF
--- a/reason-react-native/src/apis/YellowBox.md
+++ b/reason-react-native/src/apis/YellowBox.md
@@ -1,13 +1,23 @@
 ---
 id: apis/YellowBox
 title: YellowBox
-wip: true
+officialDoc: https://facebook.github.io/react-native/docs/debugging#warnings
 ---
 
-```reason
-[@bs.module "react-native"] [@bs.scope "YellowBox"]
-external ignoreWarnings: array(string) => unit = "";
+Warning messages are displayed in white on a yellow background; which is known
+as a YellowBox. Warnings may be trigged with the `warn` method already included
+the BuckleScript module `Js.Console`. This module relates to the `YellowBox`
+module in React Native.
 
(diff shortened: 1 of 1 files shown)


reason-react-native/src/apis/YellowBox.md
```diff
@@ -1,13 +1,23 @@
...
 title: YellowBox
-wip: true
+officialDoc: https://facebook.github.io/react-native/docs/debugging#warnings
 ---
 
-```reason
-[@bs.module "react-native"] [@bs.scope "YellowBox"]
-external ignoreWarnings: array(string) => unit = "";
+Warning messages are displayed in white on a yellow background; which is known
+as a YellowBox. Warnings may be trigged with the `warn` method already included
```
MoOx: I would go for ``Warnings may be trigged with the `Js.Console.warn` method included BuckleScript`` (so you can directly copy/paste/visualise the call to make

cknitt: I would write ```Warnings may be trigged with the `Js.Console.warn` function included in BuckleScript.```

reason-react-native/src/apis/YellowBox.md
```diff
@@ -1,13 +1,23 @@
...
+the BuckleScript module `Js.Console`. This module relates to the `YellowBox`
+module in React Native.
```
cknitt: I'd remove the last sentence.

MoOx:
@sgny one nice little thing is that we should prefix commit with module name when possible to help for changelog (here adding `reason-react-native: ` would have been nice ^^).
We can do it when using `Squash & Merge` has it allow you to adjust the commit message.

PR #584, opened 2019-08-23
Fix Animated.event type so it can be used in `Animated.ScrollView` `onScroll` props (& similar)

MoOx:
According to RN docs (and experience) it's handy to be able to use `Animated.event` to map `onScroll` event directly to an `Animated.Value`.
Currently the bindings make this impossible as if you try to do this

```reason
open ReactNative;

let styles = Style.(StyleSheet.create({"container": style(~flex=1., ())}));

type renderArgs = {
  scrollYAnimatedValue: Animated.Value.t,
  scrollViewRef:
    React.Ref.t(Js.Nullable.t(ReactNative.Animated.ScrollView.element)),
};

[@react.component]
let make =
    (
      ~style as s=?,
      ~contentContainerStyle=?,
      ~children: renderArgs => React.element,
    ) => {
  let scrollViewRef = React.useRef(Js.Nullable.null);
  let (scrollYAnimatedValue, _setScrollYAnimatedValue) =
    React.useState(() => Animated.Value.create(0.));

  <Animated.ScrollView
    ref=scrollViewRef
    style=Style.(arrayOption([|Some(styles##container), s|]))
    ?contentContainerStyle
    scrollEventThrottle=1
    onScroll={Animated.event(
      [|{
          "nativeEvent": {
            "contentOffset": {
              "y": scrollYAnimatedValue,
            },
          },
        }|],
      {"useNativeDriver": true},
    )}
    showsVerticalScrollIndicator=false
    showsHorizontalScrollIndicator=false>
    {children({scrollYAnimatedValue, scrollViewRef})}
  </Animated.ScrollView>;
};
```

You will get an error like this

```
  We've found a bug for you!
......./AnimatedScrollView.re 21:14-30:6

  19 ┆ ?contentContainerStyle
  20 ┆ scrollEventThrottle=1
  21 ┆ onScroll={Animated.event(
  22 ┆   [|{
   . ┆ ...
  29 ┆   {"useNativeDriver": true},
  30 ┆ )}
  31 ┆ showsVerticalScrollIndicator=false
  32 ┆ showsHorizontalScrollIndicator=false>

  This has type:
    ReactNative.Animated.animatedEvent (defined as
      ReactNative.Animated.animatedEvent)
  But somewhere wanted:
    ReactNative.Event.scrollEvent => unit
```

Which is valid!

This change make this valid (tested in the app I currently work on).

### DIFF
--- a/src/apis/Animated.md
+++ b/src/apis/Animated.md
@@ -262,7 +262,7 @@ external eventOptions2:
 type event('event) = 'event => unit;
 
 [@bs.module "react-native"] [@bs.scope "Animated"]
-external event: ('mapping, eventOptions) => event('event) = "";
+external event: (array('mapping), eventOptions) => event('event) = "";
 
 [@bs.module "react-native"] [@bs.scope "Animated"]
 external event2: (('mapping, 'mapping2), eventOptions2) => event('event) = "";
--- a/src/apis/Animated.md
+++ b/src/apis/Animated.md
@@ -246,10 +246,10 @@ external loop: Animation.t => Animation.t = "";
 [@bs.module "react-native"] [@bs.scope "Animated"]
 external loopWithConfig: (Animation.t, loopConfig) => Animation.t = "loop";
 
-type animatedEvent;
+type animatedEvent('a) = 'a => unit;
 
 [@bs.module "react-native"] [@bs.scope "Animated"]
-external event: (array('a), 'b) => animatedEvent = "";
+external event: (array('a), 'b) => animatedEvent('c) = "";
 
 [@bs.module "react-native"] [@bs.scope "Animated"]
 external createAnimatedComponent:
--- a/src/apis/Animated.md
+++ b/src/apis/Animated.md
@@ -246,10 +246,26 @@ external loop: Animation.t => Animation.t = "";
 [@bs.module "react-native"] [@bs.scope "Animated"]
 external loopWithConfig: (Animation.t, loopConfig) => Animation.t = "loop";
 
-type animatedEvent('a) = 'a => unit;
+type eventOptions;
+[@bs.obj]
+external eventOptions:
+  (~listener: 'event => unit=?, ~useNativeDriver: bool=?, unit) => eventOptions =
+  "";
+
+type eventOptions2;
+[@bs.obj]
+external eventOptions2:
+  (~listener: ('event, 'option2) => unit=?, ~useNativeDriver: bool=?, unit) =>
+  eventOptions2 =
+  "";
+
+type event('event) = 'event => unit;
 
 [@bs.module "react-native"] [@bs.scope "Animated"]
-external event: (array('a), 'b) => animatedEvent('c) = "";
+external event: ('mapping, eventOptions) => event('event) = "";
+
+[@bs.module "react-native"] [@bs.scope "Animated"]
+external event2: (('mapping, 'mapping2), eventOptions2) => event('event) = "";
 
 [@bs.module "react-native"] [@bs.scope "Animated"]
 external createAnimatedComponent:
--- a/src/apis/Animated.md
+++ b/src/apis/Animated.md
@@ -262,10 +262,10 @@ external eventOptions2:
 type event('event) = 'event => unit;
 
 [@bs.module "react-native"] [@bs.scope "Animated"]
-external event: (array('mapping), eventOptions) => event('event) = "";
+external event1: (array('mapping), eventOptions) => event('event) = "event";
 
 [@bs.module "react-native"] [@bs.scope "Animated"]
-external event2: (('mapping, 'mapping2), eventOptions2) => event('event) = "";
+external event2: (('mapping, 'mapping2), eventOptions2) => event('event) = "event";
 
 [@bs.module "react-native"] [@bs.scope "Animated"]
 external createAnimatedComponent:
--- a/src/apis/Animated.md
+++ b/src/apis/Animated.md
@@ -246,26 +246,19 @@ external loop: Animation.t => Animation.t = "";
 [@bs.module "react-native"] [@bs.scope "Animated"]
 external loopWithConfig: (Animation.t, loopConfig) => Animation.t = "loop";
 
-type eventOptions;
+type eventOptions('a);
 [@bs.obj]
 external eventOptions:
-  (~listener: 'event => unit=?, ~useNativeDriver: bool=?, unit) => eventOptions =
+  (~listener: 'a=?, ~useNativeDriver: bool=?, unit) => eventOptions('a) =
   "";
 
-type eventOptions2;
-[@bs.obj]
-external eventOptions2:
-  (~listener: ('event, 'option2) => unit=?, ~useNativeDriver: bool=?, unit) =>
-  eventOptions2 =
-  "";
-
-type event('event) = 'event => unit;
-
+// multiple externals
 [@bs.module "react-native"] [@bs.scope "Animated"]
-external event1: (array('mapping), eventOptions) => event('event) = "event";
+external event1: (array('mapping), eventOptions('a)) => 'a = "event";
 
+// multiple externals
 [@bs.module "react-native"] [@bs.scope "Animated"]
-external event2: (('mapping, 'mapping2), eventOptions2) => event('event) = "event";
+external event2: (('mapping1, 'mapping2), eventOptions('a)) => 'a = "event";
 
 [@bs.module "react-native"] [@bs.scope "Animated"]
 external createAnimatedComponent:
--- a/src/apis/Animated.re
+++ b/src/apis/Animated.re
@@ -242,10 +242,19 @@ external loop: Animation.t => Animation.t = "";
 [@bs.module "react-native"] [@bs.scope "Animated"]
 external loopWithConfig: (Animation.t, loopConfig) => Animation.t = "loop";
 
-type animatedEvent;
+type eventOptions('a);
+[@bs.obj]
+external eventOptions:
+  (~listener: 'a=?, ~useNativeDriver: bool=?, unit) => eventOptions('a) =
+  "";
 
+// multiple externals
 [@bs.module "react-native"] [@bs.scope "Animated"]
-external event: (array('a), 'b) => animatedEvent = "";
+external event1: (array('mapping), eventOptions('a)) => 'a = "event";
+
+// multiple externals
+[@bs.module "react-native"] [@bs.scope "Animated"]
+external event2: (('mapping1, 'mapping2), eventOptions('a)) => 'a = "event";
 
 [@bs.module "react-native"] [@bs.scope "Animated"]
 external createAnimatedComponent:


reason-react-native/src/apis/Animated.re
```diff
@@ -239,10 +239,10 @@ external loop: Animation.t => Animation.t = "";
 [@bs.module "react-native"] [@bs.scope "Animated"]
 external loopWithConfig: (Animation.t, loopConfig) => Animation.t = "loop";
 
-type animatedEvent;
+type animatedEvent('a) = 'a => unit;
 
 [@bs.module "react-native"] [@bs.scope "Animated"]
-external event: (array('a), 'b) => animatedEvent = "";
+external event: (array('a), 'b) => animatedEvent('c) = "";
```
cknitt: Some documentation / explanation for this type signature would be nice.

sgny:
@MoOx I haven't used this either, but I realize that it is used either with `ScrollView` or `PanResponder`. Your PR would cover use with `ScrollView, however, according to these [examples](https://animationbook.codedaily.io/animated-event/), it may not work for `PanResponder`.

Before merging, we should better understand what's required for use with `PanResponder` (perhaps then it only takes the array argument). For my part, I should be able to take a proper look later tonight, but let's discuss this a bit further.

sgny:
@MoOx We can be more specific about 'b, basically it's an optional config which has two optional keys `listener` (which may then be a function `'b => unit` ) and `useNativeDriver` which is a `bool`.

MoOx:
So currently the problem was that animatedEvent type was "undefined" (not in the JS sens), like "unspecified".
So when passed to `onScroll`, compiler isn't happy since it need "something that look like what onScroll expect".
Passing a unspecified arg to the type allow to "workaround" the issue.

It should be enough for PanResponder:
It expect for the listener `(event: PressEvent, gestureState: GestureState) => mixed` (per flow annotations). In the precise case of `onPanResponderMove`, it doesn't return anything https://github.com/facebook/react-native/blob/8e04a1485f377e2d7d744fe74fc26e1dd0c056fc/Libraries/Interaction/PanResponder.js#L521
The first arg is ok as it's an array to map the params passed to the underlying function (that we can have in the optional listener arg.

Here we can see that signature are the same

```js
 onScroll={Animated.event(
   [{nativeEvent: {contentOffset: {x: this._scrollX}}}],
   {listener: (event) => console.log(event)}, // Optional async listener
 )}
///  ...
 onPanResponderMove: Animated.event([
   null,                // raw event arg ignored
   {dx: this._panX}],    // gestureState arg
{listener: (event, gestureState) => console.log(event, gestureState)}, // Optional async listener
 ),
```

But I agree that we could make this more safe.

Take a look at my recent commit & tell me what you think. It should be ok to be able to put things like this

```reason
    onScroll={Animated.event(
      [|
        {
          "nativeEvent": {
            "contentOffset": {
              "x": scrollXAnimatedValue,
              "y": scrollYAnimatedValue,
            },
          },
        },
      |],
      Animated.eventOptions(
        ~useNativeDriver=true,
        ~listener=
          event => {
            Js.log(event);
            ();
          },
        (),
      ),
    )}
```

```reason
    onPanResponderMove={Animated.event2(
      (Js.null, {"dx": "whatever"}),
      Animated.eventOptions2(
        ~listener=
          (event, ges) => {
            Js.log2(event, ges);
            ();
          },
        (),
      ),
    )}
```

(I used event & event2 as event could take an array but array doesn't allow to mix type (I just did like useEffectX in reason-react)

reason-react-native/src/apis/Animated.re
```diff
@@ -239,10 +239,26 @@ external loop: Animation.t => Animation.t = "";
...
+  "";
+
+type event('event) = 'event => unit;
+
+[@bs.module "react-native"] [@bs.scope "Animated"]
+external event: (array('mapping), eventOptions) => event('event) = "";
 
 [@bs.module "react-native"] [@bs.scope "Animated"]
-external event: (array('a), 'b) => animatedEvent = "";
+external event2: (('mapping, 'mapping2), eventOptions2) => event('event) = "";
```
cknitt: Should be `= "event"`.

reason-react-native/src/apis/Animated.re
```diff
@@ -239,10 +239,26 @@ external loop: Animation.t => Animation.t = "";
...
+[@bs.obj]
+external eventOptions2:
+  (~listener: ('event, 'option2) => unit=?, ~useNativeDriver: bool=?, unit) =>
+  eventOptions2 =
+  "";
+
+type event('event) = 'event => unit;
+
+[@bs.module "react-native"] [@bs.scope "Animated"]
+external event: (array('mapping), eventOptions) => event('event) = "";
```
cknitt: To match `useCallbackX`, `useEffectX`etc., maybe this should be called `event1`?

reason-react-native/src/apis/Animated.re
```diff
@@ -239,10 +239,26 @@ external loop: Animation.t => Animation.t = "";
...
-type animatedEvent;
+type eventOptions;
+[@bs.obj]
+external eventOptions:
+  (~listener: 'event => unit=?, ~useNativeDriver: bool=?, unit) => eventOptions =
+  "";
+
+type eventOptions2;
+[@bs.obj]
+external eventOptions2:
```
cknitt: Do we really need two different eventOptions types? In the worst case, the second listener param would be `undefined` for some event types, right?

MoOx: Yeah it will be undefined as it won't be explicitly defined.

Maybe we can go with an `=?` (not sure if that ok in practice tho)

```reason
type eventOptions;
	[@bs.obj]
	external eventOptions:
	  (~listener: ('event, 'option2=?) => unit=?, ~useNativeDriver: bool=?, unit) => eventOptions =
	  "";
```

reason-react-native/src/apis/Animated.re
MoOx: Yeah I was thinking about that. Will make the change.

reason-react-native/src/apis/Animated.re
sgny: If the `listener` has to be either `event => unit` or `(event, gestureState) => unit`, why don't we make use of that?

Sorry, I haven't had time to try out `Animated.event` yet, so I can't be more useful.

sgny:
Shouldn't `event` return `Event.scrollEvent` for use with `ScrollView` and `(Event.pressEvent, gestureState) => unit` for use with PanResponder`? Then we only need to define in `Animated.re`

```
type eventOptions('a);
[@bs.obj]
external eventOptions:
  (~listener: 'a=?, ~useNativeDriver: bool=?, unit) => eventOptions('a) =
  "";

[@bs.module "react-native"] [@bs.scope "Animated"]
external event: (array(mapping), eventOptions('a)) => 'a = "";
```

If we need to handle multiple mappings, we could define a function to cast anything as a `mapping`.

sgny:
What I'm still not sure on is how to actually use this. Basically, if `e` is the `scrollEvent`, the mapping
```
{"nativeEvent": {"contentOffset": {"y": scrollYAnimatedValue}}}
```
is supposed to set `scrollYAnimatedValue = e##nativeEvent##contentOffset##y`, and then any animated value which references `scrollYAnimatedValue` tracks that. I don't know how that can be expressed in Reason.

MoOx:
@sgny Not sure what you mean with your last comment but that's the point of animated value, to transparently affect all usage (interpolation etc). I don't think we need to do something more than what's done here without making bindings with a js cost.

My idea was to keep the thing "like RN". We could go with `Animated.scrollEvent` & `Animated.panResponderEvent`, without generic types, more specific & very strongly typed, but that may be less accessible for new comers. I would go to keep things simple, as anyway this kind of usage need to be tested IRL, compilation doesn't ensure that the code does what you will think it describe.

sgny:
@MoOx Exactly, we need to check whether this works. Some sample code makes it seem like the mapping is in fact the left hand side of a destructuring statement where the (implied) right hand side is the event or gestureState.

I think this will work, but if I’m right, what looks like the creation of a hashmap for which some key equals the variable, will in fact be a reassignment of that variable.  That is an unexpected call to an impure function.

MoOx:
Actually I am currently using this patch & it's working great (at least for ScrollView onScroll).

Not sure to follow you on the left/right/impure thing.

I used event1 & event2 to match the mapping required. The mapping for event function actually not directly match the argument from the listener. The same original event are applied that's it. The mapping made via things like `{"nativeEvent": {"contentOffset": {"y": scrollYAnimatedValue}}}` is just to adjust the `scrollYAnimatedValue`. The listener just receive the original event, untouched (that's how it make sense to me, as the listener is in practice unnecessary most of the time, & even probably a bad idea as using JS to listen to animated value can be really slow...)

sgny:
> Actually I am currently using this patch & it's working great (at least for ScrollView onScroll).
 
I wasn't sure you were already using it. Since you've clarified that, the discussion becomes a bit pedantic.

> Not sure to follow you on the left/right/impure thing.
> 
> I used event1 & event2 to match the mapping required. The mapping for event function actually not directly match the argument from the listener. The same original event are applied that's it. The mapping made via things like `{"nativeEvent": {"contentOffset": {"y": scrollYAnimatedValue}}}` is just to adjust the `scrollYAnimatedValue`. 

What I mean is that when we supply the mapping, it is used in a destructuring statement as below
```
var {nativeEvent: {contentOffset: {y: scrollYAnimatedValue}}} = e
```
where `e` is the event. That's why I was referring to the mapping as the left-hand side of the statement. Then `scrollYAnimatedValue` gets assigned the value `e##nativeEvent##contentOffset##y` and there is an impure function which might need to be documented as such. Obviously the compiler simply renders `{"nativeEvent": {"contentOffset": {"y": scrollYAnimatedValue}}}` as the creation of a new object and unaware of the reassignment, but the JS output then is as expected by React Native.

> The listener just receive the original event, untouched (that's how it make sense to me, as the listener is in practice unnecessary most of the time, & even probably a bad idea as using JS to listen to animated value can be really slow...)

Without `Animated.event`, you'd be passing a handler to the prop to handle the input; I believe `listener` is in fact for such handlers.

I hope I've made myself clearer.

sgny:
@MoOx I have had some opportunity to test over the weekend. If you take a look at this example below, it seems that we could indeed have only `eventOptions`, as the return type required for use with either prop will enforce type safety.

On the other hand, we could have multiple bindings to account for the array being either a singleton or having two elements.

```
open ReactNative;

type mapping;
external makeMapping: 'a => mapping = "%identity";

type eventOptions('a);
[@bs.obj]
external eventOptions:
  (~listener: 'a=?, ~useNativeDriver: bool=?, unit) => eventOptions('a) =
  "";

[@bs.module "react-native"] [@bs.scope "Animated"]
external event: (array(mapping), eventOptions('a)) => 'a = "";


let widthAnimation = Animated.Value.create(0.);
let heightAnimation = Animated.Value.create(0.);

let panConfig = PanResponder.config(
    ~onMoveShouldSetPanResponder={(_,_) => true},
    ~onPanResponderMove=event([|
        makeMapping(Js.null),
        makeMapping({
          "moveX": widthAnimation,
          "moveY": heightAnimation
        }),
      |], 
  		eventOptions(~listener={(_, _) => Js.Console.warn(g##moveX)}, ())
  	),
    ()
  );

let panHandlers = PanResponder.create(panConfig) -> PanResponder.panHandlers;

[@react.component]
let make = () => {
	<View
		style=MLStyles.Layout.innerbody
		onMoveShouldSetResponder={
		  panHandlers->PanResponder.onMoveShouldSetResponder
		}
		onMoveShouldSetResponderCapture={
		  panHandlers->PanResponder.onMoveShouldSetResponderCapture
		}
		onStartShouldSetResponder={
		  panHandlers->PanResponder.onStartShouldSetResponder
		}
		onStartShouldSetResponderCapture={
		  panHandlers->PanResponder.onStartShouldSetResponderCapture
		}
		onResponderReject={panHandlers->PanResponder.onResponderReject}
		onResponderGrant={panHandlers->PanResponder.onResponderGrant}
		onResponderRelease={panHandlers->PanResponder.onResponderRelease}
		onResponderMove={panHandlers->PanResponder.onResponderMove}
		onResponderTerminate={panHandlers->PanResponder.onResponderTerminate}
		onResponderStart={panHandlers->PanResponder.onResponderStart}
		onResponderTerminationRequest={
		  panHandlers->PanResponder.onResponderTerminationRequest
		}
		onResponderEnd={panHandlers->PanResponder.onResponderEnd}
	>
    <Animated.View
      style=Style.style(
        ~backgroundColor=Color.red,
        ~width=Animated.StyleProp.size(
          Animated.Interpolation.interpolate(
            widthAnimation,
            Animated.Interpolation.config(
              ~inputRange=[|0., 2.|],
              ~outputRange=Animated.Interpolation.fromFloatArray([|0., 1.|]),
              ()
            )
          )
        ),
        ~height=Animated.StyleProp.size(
          Animated.Interpolation.interpolate(
            heightAnimation,
            Animated.Interpolation.config(
              ~inputRange=[|0., 2.|],
              ~outputRange=Animated.Interpolation.fromFloatArray([|0., 1.|]),
              ()
            )
          )
        ),
        ()
      )
    />

  </View>
};
```

I also tried `null` as the second mapping and that seems to work fine, even when not working with `PanResponder`. Presumably, we could define a tuple of mappings and avoid having arrays of incorrect sizes.

What do you think?

MoOx:
This discussion is starting to confuse me, sorry about that haha.
I didn't followed the message you are trying to send totally (maybe I am tired).

With the current implementation made by this PR, we have a minimum of safety about the number of argument, for animated (one arg) & panresponder (2 args) with *1 & *2).
But we have to "handle this" by hand.

On the other hand, with what you propose 

```reason
type eventOptions('a);
[@bs.obj]
external eventOptions:
  (~listener: 'a=?, ~useNativeDriver: bool=?, unit) => eventOptions('a) =
  "";

[@bs.module "react-native"] [@bs.scope "Animated"]
external event: (('b, 'c), eventOptions('a)) => 'a = "";
```

I understood, we keep a single implementation, that will autovalidate itself by the fact that `'a` will handle the proper inference. What I don't follow is how we validate the fact that `'c` doesn't make sense in the case of scrollview onScroll event.

For this, we could try to mix having a single eventOptions, but 2 `event1` & `event2` like it's the case in my PR ?

I really don't know what will be the safest in practise and more importantly for adoption, the simplest to use without compromising quality of the binding.

sgny:
Sorry I could not reply earlier.

As far as I can tell (but I only tried on the iOS simulator so far), the second mapping is ignored when it's not relevant and should not create issues. On the other hand, it might be slightly painful to require an additional argument when it's absolutely unnecessary.

We could avoid the `gestureState` mapping where it is unnecessary and still have type safety with only one binding. To do that, however, I suppose we'd have to define additional types and functions, which might be rather clunky.

Having two bindings as you have proposed might be better from an ease-of-use perspective.

sgny:
Upon further thought, should we have constructors for the various event mappings? Similar to what I have added for `gestureState`?

MoOx:
> Having two bindings as you have proposed might be better from an ease-of-use perspective.

Like I already said, simple & "like reason-react" codebase. So might help people to find similar pattern.

> should we have constructors for the various event mappings? Similar to what I have added for gestureState?

You mean add the gestureState mentionned here https://github.com/reason-react-native/reason-react-native/pull/584#issuecomment-527232691?

I don't find this necessary *ATM* since I just used this for scrollView & never used something more.
Maybe we could merge my current solution & iterate later if necessary?

sgny:
How about a mix and match of both proposal, keeping the paramterised `eventOptions` type but having two bindings as below?

```reason
type eventOptions('a);
[@bs.obj]
external eventOptions:
  (~listener: 'a=?, ~useNativeDriver: bool=?, unit) => eventOptions('a) =
  "";

[@bs.module "react-native"] [@bs.scope "Animated"]
external event1: (array('mapping), eventOptions('a)) => 'a = "event";

[@bs.module "react-native"] [@bs.scope "Animated"]
external event1: (('mapping, 'mapping2), eventOptions('a)) => 'a = "event";
```

We could actually enforce that `array('mapping)` is a singleton (by means of a function that takes the mapping and returns that mapping in an array), that would enforce correctness, but would that be too much noise?

MoOx:
Yeah ok will do that. I don't want to add a method to enforce single item array as this won't be zero cost anymore.

sgny:
Given your agreement to merge the proposals, I added another commit to handle that. I also added the `multiple externals` comments as we have `event1` and `event2`.

PR #586, opened 2019-08-26
Add documentation for Clipboard

sgny:
Quite barebones, but then there isn't all that much to explain...


reason-react-native/src/apis/Clipboard.md
```diff
@@ -1,14 +1,24 @@
 ---
 id: apis/Clipboard
 title: Clipboard
-wip: true
+officialDoc: https://facebook.github.io/react-native/docs/clipboard
```
MoOx: @sgny do you know that officialDoc is optional & otherwise generated from the module name?

PR #587, opened 2019-08-27
reason-react-native: add AccessibilityInfo API

sgny:
I noticed that these were missing.

### DIFF
--- a/reason-react-native/src/apis/AccessibilityInfo.md
+++ b/reason-react-native/src/apis/AccessibilityInfo.md
@@ -1,0 +1,72 @@
+---
+id: apis/AccessibilityInfo
+title: AccessibilityInfo
+wip: true
+---
+
+```reason
+type announcementResult = {
+  .
+  "announcement": string,
+  "success": bool,
+};
+
+[@bs.scope "AccessibilityInfo"] [@bs.module "react-native"]
+external isBoldTextEnabled: unit => Js.Promise.t(bool) = "";
+
+[@bs.scope "AccessibilityInfo"] [@bs.module "react-native"]
+external isGrayscaleEnabled: unit => Js.Promise.t(bool) = "";
+
+[@bs.scope "AccessibilityInfo"] [@bs.module "react-native"]
+external isInvertColorsEnabled: unit => Js.Promise.t(bool) = "";
+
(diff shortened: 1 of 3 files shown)


reason-react-native/src/apis/AccessibilityInfo.md
```diff
@@ -0,0 +1,72 @@
+---
+id: apis/AccesibilityInfo
+title: AccesibilityInfo
+wip: true
+---
+
+```reason
+type payload = {
```
cknitt: Shouldn't this be called `announcementFinishedPayload` (or maybe `announcementFinishedParams` to match other event types)?

sgny: How about announcementResult or announcementOutcome?

reason-react-native/src/apis/AccessibilityInfo.md
```diff
@@ -0,0 +1,72 @@
...
+  .
+  "announcement": string,
+  "success": bool,
+};
+
+[@bs.scope "AccesibilityInfo"] [@bs.module "react-native"]
```
MoOx: AccesibilityInfo => AccessibilityInfo

Should be adjusted everywhere :)

reason-react-native/src/apis/AccessibilityInfo.md
MoOx: `announcementFinishedResult` seems a good compromise.
Not a fan of "params" that could be interpreted as an argument from the user.

cknitt: 👍

MoOx:
@sgny You missed a place, and I cannot edit since you did that on your fork (reminder: you have write access on this repo, no need for a fork ^^).

Btw, maybe `announcementFinishedResult` would be more explicit that `*Paypload` ? (I am saying that since a payload look more that something that I would "give" than something I "receive" (but I am not English fluent so I can be wrong)

sgny:
@MoOx I’ll fix that tomorrow, sorry about that.

Regarding payload you are more accurate, but I couldn’t find a more appropriate name at the time.

Would announcementResult be clear enough?

MoOx:
I guess announcementResult is good enough

Issue #594, opened 2019-08-29
Reimplement core deprecated module as new packages (@reason-react-native/*)


MoOx:
Ordered by number of download on npm

- [x] https://www.npmjs.com/package/@react-native-community/async-storage (100+k dl/w) [WIP](https://github.com/reasonml-community/reason-react-native/pull/597)
- [x] https://www.npmjs.com/package/@react-native-community/geolocation (15k dl/w)
- [x] https://www.npmjs.com/package/@react-native-community/slider (12k dl/w)
- [x] https://www.npmjs.com/package/react-native-webview (10+k dl/w)
- [ ] https://www.npmjs.com/package/@react-native-community/viewpager (4+k dl/w)
- [ ] https://www.npmjs.com/package/@react-native-community/masked-view (4k dl/w)
- [x] https://www.npmjs.com/package/@react-native-community/checkbox (-1k dl/w)

Another step after that will be famous community packages.

(Note: we could do that now, & integrate the list to priorize on the number of npm download, extracted from core or from the community)

Should we consider at this point to make a dedicated GitHub org `@reason-react-native` and host all packages there?

idkjs:
You know, I think the github org thing might sound more legitimate to some people. If the idea to get more people involved.

cknitt:
Github org sounds good to me! 👍

tatchi:
I would be happy to help with the `async-storage` package :)

MoOx:
@tatchi feel free to open a PR based on netinfo folder, even if it's a WIP :)

MoOx:
I created an https://github.com/reason-react-native, I can create repo when necessary, just ask here for now.
I will split this monorepo asap into multiple packages as (imo) it doesn't make sense now to keep a monorepo as different modules won't need to be in sync with the reason-react-native core repo now that we have something stable.

cknitt:
@MoOx I could contribute:

* https://www.npmjs.com/package/@cca-io/bs-react-native-device-info
* https://www.npmjs.com/package/@cca-io/bs-rn-fetch-blob
* Some stuff that we have not publicly released yet, e.g. react-native-keychain bindings.

Shall we move the core `reason-react-native` repo to the `reason-react-native` org, too?

MoOx:
Nice!

I will extract/move all packages in this monorepo indeed, including the "core" one.

[...]

PR #597, opened 2019-08-30
async-storage bindings

tatchi:
Bindings for the [async-storage](https://github.com/react-native-community/async-storage) package. (#594)

I have implemented them based on the (current ones)[https://github.com/reasonml-community/reason-react-native/blob/master/reason-react-native/src/apis/AsyncStorage.re] but have slightly modified two of them:

- multiMerge: use `array((string, string))` instead of `array(array(string))`. Since we always pass an array of arrays containing 2 elements (key and value), I think tuple would be a better fit (like it's already done for the `multiSet)`.
- multiGet: also replaced the double arrays to an array of tuples (for the same reason). Furthermore and according to my tests, we never get back a `null` value. If a `key` does not exist, this key will still exist in the returned array but only the corresponding `value` will be `null`. That still seems strange since that's not what the doc mention: https://github.com/react-native-community/async-storage/blob/LEGACY/docs/API.md#multiget

Except from that, the rest should be the same.

I still need to create the binding for the [useAsyncStorage](https://github.com/react-native-community/async-storage/blob/LEGACY/docs/API.md#useasyncstorage) hooks. I will address that later on.

### DIFF
--- a/@reason-react-native/async-storage/.gitignore
+++ b/@reason-react-native/async-storage/.gitignore
@@ -1,0 +1,5 @@
+# Ocaml / Reason / BuckleScript artifacts
+.bsb.lock
+**/lib/bs
+**/lib/ocaml
+**/.merlin
--- a/@reason-react-native/async-storage/bsconfig.json
+++ b/@reason-react-native/async-storage/bsconfig.json
@@ -1,0 +1,20 @@
+{
+  "name": "@reason-react-native/async-storage",
+  "refmt": 3,
+  "reason": {
+    "react-jsx": 3
+  },
+  "package-specs": {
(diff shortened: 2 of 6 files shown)


MoOx:
@tatchi Thanks a lot for starting this.
Following https://github.com/reasonml-community/reason-react-native/issues/594 we are going to stop using a huge monorepo, to make contributions easier (and release process for things not directly related doesn't make a lot of sense imo in a monorepo).
So can you start your PR into https://github.com/reason-react-native/async-storage please?
Would be awesome!

tatchi:
Done.

Closing this in favor of https://github.com/reason-react-native/async-storage/pull/1

PR #600, opened 2019-09-03
Add Animated.SectionList and Animated.FlatList

### LINKED ISSUE #599: Add missing Animated.FlatList & SectionList
https://github.com/reasonml-community/reason-react-native/blob/1bbb716dff888ad67b06641c67a31aa3e01f7c49/reason-react-native/src/apis/Animated.re#L284-L306

vs

https://github.com/facebook/react-native/blob/adac04601ec36d35eca42532ec824c5790dc71ed/Libraries/Animated/src/Animated.js#L23-L40

sgny:
This may close #599, but it's not perfect. 

I have tested that any prop not valid for `FlatList` will cause a compile error with `Animated.FlatList` so it works quite alright, at least with respect to type safety without additional costs of endless instances of `undefined` in the JS output.

The slight (IMO) drawback is that the JS output has one additional layer of `React.createElement` and there are two ways to avoid it (which would be alternatives to this implementation).

• duplicate FlatList.re, but add the `[@bs.scope "Animated"]` decorator.
• instead of using type arguments like `'item`, etc. in `FlatList` or `SectionList`, use types like `item` and add functions to cast `'a` into `item` (basically using the `%identity` `external` definition).

All in all, I have wanted to solve this, ever since the initial commit that introduced `Animated.*`, but apparently I needed an impetus for that.

### DIFF
--- a/src/apis/Animated.bs.js
+++ b/src/apis/Animated.bs.js
@@ -1,11 +1,14 @@
 'use strict';
 
+var React = require("react");
 var Caml_option = require("bs-platform/lib/js/caml_option.js");
 var ReactNative = require("react-native");
 var Text$ReactNative = require("../components/Text.bs.js");
 var View$ReactNative = require("../components/View.bs.js");
 var Image$ReactNative = require("../components/Image.bs.js");
+var FlatList$ReactNative = require("../components/FlatList.bs.js");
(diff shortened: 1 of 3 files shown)


MoOx:
@sgny You might have to reapply your change since I moved code at the root of the repo. Try rebasing, maybe that will work painlessly

PR #603, opened 2019-09-04
Fix Animated.Value.addListener parameter type

mtnptrsn:
The callback for `Animated.Value.addListener` is currently typed like this: `float => unit`, but is supposed to be `{. "value": float} => unit`.

Link to RN docs: https://facebook.github.io/react-native/docs/animatedvalue#addlistener

### DIFF
--- a/src/apis/Animated.re
+++ b/src/apis/Animated.re
@@ -12,7 +12,7 @@ module Animation = {
 module type Value = {
   type t;
   type jsValue;
-  type jsValueAddListener;
+  type addListenerCallback;
 };
 
 type calculated;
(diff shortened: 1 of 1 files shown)


mtnptrsn:
> Good catch for `Animated.Value`, but this would break `Animated.ValueXY`.

That's right. Good catch. Will resolve this :)

mtnptrsn:
@sgny Done some changes now. Please let me know what you think.

sgny:
Yes, that's along how I was thinking.

However, `Animated` is quite complicated. At least that's my excuse for why I missed the type of the callback for `addListener`. There were three such types before and I removed `{. "value": jsValue}` when I realized that was incorrect for `stopAnimation`. I did not realise that `addListener` did not mirror the others. Anyway, `stopAnimation` needs to take `float => unit` for `Value`, so we need to roll that back.

Also, perhaps it's simpler to include a type such as `addListenerCallback` in the module type signature, rather than use a type parameter with `callback`

```reason
module type Value = {
  type t;
  type jsValue;
  type addListenerCallback;
};
```

for the other callbacks, you could simply define
```
type callback = jsValue => unit;
```
as before.

mtnptrsn:
> Yes, that's along how I was thinking.
> 
> However, `Animated` is quite complicated. At least that's my excuse for why I missed the type of the callback for `addListener`. There were three such types before and I removed `{. "value": jsValue}` when I realized that was incorrect for `stopAnimation`. I did not realise that `addListener` did not mirror the others. Anyway, `stopAnimation` needs to take `float => unit` for `Value`, so we need to roll that back.
> 
> Also, perhaps it's simpler to include a type such as `addListenerCallback` in the module type signature, rather than use a type parameter with `callback`
> 
> ```reason
> module type Value = {
>   type t;
>   type jsValue;
>   type addListenerCallback;
> };
> ```
> 
> for the other callbacks, you could simply define
> 
> ```
> type callback = jsValue => unit;
> ```
> 
> as before. Or if you'd keep `callback('a)`, `callback(jsValue)`.

Great suggestions. Way cleaner. What do you think now?

sgny:
Looks great. Welcome to the team of contributors.

sgny:
@MoOx Can you tell why the test is not completing?

MoOx:
@sgny apparently I didn't grant circleci full permissions so it wasn't setup correctly to report status back.

This PR is good.

image

PR #607, opened 2019-09-13
fix(Webview): typo in `contentInset` prop

cem2ran:
According to RN Repo it should be `contentInset` and not `contentInsets`:
```
contentInset?: ?EdgeInsetsProp,
```
https://github.com/facebook/react-native/blob/f8e509382315f84e3c29b07cc174093739ed980f/Libraries/Components/ScrollView/ScrollView.js#L106

This change should also be made to [bs-react-native-jsx3-compat](https://github.com/reason-react-native/bs-react-native-jsx3-compat) and bs-react-native if it's still supported. 

Let me know if you want me to do this or if it's something you'll do when you cut the next release.

PR #608, opened 2019-09-13
fix(WebView): event types

cem2ran:
This PR adds additional fields to event types in the WebView component and does this by inheriting fields from a base object type `webViewBaseEvent` as is the case in the [Typescript bindings](https://github.com/react-native-community/react-native-webview/blob/7d0c4a798e559a7d64fbe3172ee93f58401d005c/src/WebViewTypes.ts#L81-L103).

Once we upgrade to bucklescript 6 and the [reason parser has been upgraded](https://github.com/facebook/reason/pull/2036) we will be able to express this far more succinctly through Inherited fields in object type (available since 4.06):

```reason
type webViewNavigation = {. "navigationType": string, ...webViewBaseEvent};
```

This PR also fixes an incorrect typing of the `onNavigationStateChange` props callback param, which should not be wrapped in `Event.syntheticEvent`.

### DIFF
--- a/src/components/WebView.md
+++ b/src/components/WebView.md
@@ -29,17 +29,52 @@ module Source = {
 module DataDetectorTypes = WebView_DataDetectorTypes;
 module DecelerationRate = WebView_DecelerationRate;
 
-type messageEvent = Event.syntheticEvent({. "data": string});
+class type webViewBaseEvent = {
+  pub url: string;
+  pub title: string;
+  pub loading: bool;
+  pub canGoBack: bool;
+  pub canGoForward: bool;
+  pub lockIdentifier: int;
+};
 
-type webViewEvent =
-  Event.syntheticEvent({
-    .
-    "url": string,
-    "title": string,
-    "loading": bool,
-    "canGoBack": bool,
-    "canGoForward": bool,
-  });
+class type virtual webViewNavigation = {
+  as 'self;
+  constraint 'self = #webViewBaseEvent;
+  pub navigationType: string;
+  pub mainDocumentURL: option(string);
+};
+
+class type virtual webViewError = {
+  as 'self;
+  constraint 'self = #webViewBaseEvent;
+  pub domain: option(string);
+  pub code: int;
+  pub description: string;
+};
+
+class type virtual webViewErrorOrNavigation = {
+  as 'self;
+  constraint 'self = #webViewBaseEvent;
+  pub domain: option(string);
+  pub code: option(int);
+  pub description: string;
+  pub navigationType: string;
+  pub mainDocumentURL: option(string);
+};
+
+class type virtual webViewMessage = {
+  as 'self;
+  constraint 'self = #webViewBaseEvent;
+  pub data: string;
+};
+
+type webViewErrorEvent = Event.syntheticEvent(Js.t(webViewError));
+type webViewNavigationEvent = Event.syntheticEvent(Js.t(webViewNavigation));
+type webViewErrorOrNavigationEvent =
+  Event.syntheticEvent(Js.t(webViewErrorOrNavigation));
+type webViewMessageEvent = Event.syntheticEvent(Js.t(webViewMessage));
+type webViewEvent = Js.t(webViewBaseEvent);
 
 type request = {
   .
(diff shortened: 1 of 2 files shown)


cknitt:
@cem2ran Thanks for the PR! Hadn't heard of the inherited fields in object type, that's interesting!

I have mixed feelings about using class types though - we have not used them anywhere else in our react-native bindings yet.

Also, I think we should remove WebView from the core bindings and put the WebView bindings into the separate repo https://github.com/reason-react-native/webview.

cknitt:
Hmm, of course my last remark regarding removing WebView from core would apply to #607, too... 
what do you think @MoOx?

cem2ran:
@cknitt I was also slightly hesitant about using class types even though to my understanding we use them implicitly through JS object syntax, which desugars to the class types wrapped in Js.t? 
The alternative is to duplicate the inherited fields in every "sub-class".

I'm happy to do either or third option if you know of any.

Hypothetically, had we had the newer object type spread syntax would you have accepted it? 😜

sgny:
We have the same issue with `syntheticEvent` and `responderSyntheticEvent` in `Event.re` and I've been holding off for inherited fields as [JS output](https://reasonml.github.io/en/try.html?rrjsx=true&reason=MYGwhgzhAEEC5mAawPoHsBmKCWA7OMAvNAN4BQ00ksApiBgNwXQBuYI0AtgK4IBGIGtDgALGihDZ4ALmiT4ACjxwAlNGIBtALoNoAegBU0PPDC5gQtgCdsYAUIN7mAB259oriCPXQAHuoA+ZkpDD24vLhpRNAATaEdg4TEJKTgfDV8AGmgAOjzRcXk4HTIAXyYgA) with class types would have added overhead, particularly imports.

However, I wasn't aware of the `virtual` keyword which prevents all that (JS output is indeed unchanged and not merely omitted from this PR). @cem2ran This is greatly appreciated, I hadn't looked into OO with OCaml.

Perhaps one strike against adopting class types might be the lack of documentation. I still can't find anything further on `private` in Reason or BuckleScript documentation - despite knowing now to look for it. One has to refer to OCaml documentation and use the parser to figure out Reason syntax.

On the other hand, perhaps inheriting fields is simply sugar for class types? In that case, we could use it directly.

I've been working on the spin-off WebView bindings (PR is almost ready) and was looking at the TS bindings as you have (documentation is sometimes incorrect), but I decided to simply duplicate fields for now. I will mirror our decision in the main repo.

cem2ran:
@sgny it's completely possible that I'm wrong about the desugaring. Haven't looked into it in detail.

@cknitt I'm happy to close this PR or change this PR in favor of a smaller one that only fixes `onNavigationStateChange`. The larger changes can go into webview repo? 

Just trying to get rid of all my `patch-package` patches and contribute my changes upstream :)

cknitt:
@cem2ran I played a bit with the class types, and this looks good to me, as it remains zero cost, and from the user's perspective, usage remains the same as if defining "standard" `Js.t` types with duplicate fields.

Still, we would need to document that in `WebView.md` somehow. Most people won't be familiar with class types and https://bucklescript.github.io/docs/en/class#bind-to-js-classes.

sgny:
I also intend to integrate these into the bindings for the standalone `WebView`. As I mentioned, they are almost complete, but even the TypeScript spec is not perfectly accurate. I have tested on iOS and identified a few discrepancies, but I will prepare at least one PR before I go through with Android testing.

sgny:
Alright, I committed [bindings for the standalone `WebView` component](https://github.com/reason-react-native/webview). I incorporated @cem2ran's contribution, but I also added a way to specify a callback function for an untagged union type.

I'm not sure that the bindings are indeed correct in all details (e.g. I simply included all HTTP methods, but most likely only some are supported), but I tested as much as I could. I noticed that quite a few details differ from the API documentation, as well as the TypeScript specification.

MoOx:
So how we should proceed with this PR?
I am ok to merge a fix & release is under 0.60 patches, until we remove them in 0.61 in favor of new dedicated package/bindings

cknitt:
@MoOx Fine with me. Docs update is still missing though.

PR #609, opened 2019-09-14
Add Global module, expose __DEV__

### LINKED ISSUE #588: Expose __DEV__
```reason
external __DEV__: bool = "__DEV__";
```

sgny:
Closes #588. 

This has been lingering for a while. @MoOx mentioned there might be other methods that could be added, but I'm drawing a blank.

Any suggestions?

### DIFF
--- a/src/apis/Global.bs.js
+++ b/src/apis/Global.bs.js
@@ -1,0 +1,1 @@
+/* This output is empty. Its source's type definitions, externals and/or unused code got optimized away. */
--- a/src/apis/Global.md
+++ b/src/apis/Global.md
@@ -1,0 +1,10 @@
+---
+id: apis/Global
+title: Global
+wip: true
+---
+
+```reason
+[@bs.val] external __DEV__: bool = "__DEV__";
+
+```
--- a/src/apis/Global.re
+++ b/src/apis/Global.re
@@ -1,0 +1,1 @@
+[@bs.val] external __DEV__: bool = "__DEV__";


cknitt:
Thanks! Will approve, as this is what we agreed on in #588.

It occurred to me though that users will usually do `open ReactNative;`, bringing all submodules of ReactNative into scope, and therefore a generic name like `Types`, `Common` or `Global` as used here is maybe not such a good idea after all as it might already be in use in the user's project. If we decide to rename the module, we can still do this in a later PR though.

sgny:
@cknitt You make a good point. When `RNTypes` was renamed into `Types` in the past, that conflicted with my `Types` module. I don't know how common `Global` would be (definitely less common than `Types`), but we could perhaps go with `GlobalVariables` instead.

cknitt:
Or maybe even `ReactNativeGlobals` (even though that's redundant). @MoOx What do you think?

MoOx:
About other stuff, I had this in mind https://github.com/reason-react-native/reason-react-native/blob/master/src/apis/AnimationFrame.re which has its own module...
But maybe that's not intuitive when coming from JS... 
If we decide to move this stuff here, we cannot decide to use GlobalVar... This would be functions :)

Global is good for me as I don't see myself having this name as a module, but that's just me.
That said, I got the "Types" issue with BSRN in the past.

`ReactNativeGlobals` would feel a bit verbose indeed but that's an option.

Btw, `Globals` or `Global` ? Not sure what is best.

Ok my comment is totally not helping...

---

If we take a step back, what "globals" things do we have except raf functions & dev flag? 
I don't see anything...
And raf & timers are all now in `Timers/JSTimers.js`... We could have a timer module (but that would also be generic 😂 But not that much.

I am wondering if we could have directly `__DEV__` exposed by ReactNative... But since we have a namespace generated by BS, not sure we can do this...
Otherwise we could do `Metro.__DEV__` ?? Since Metro handles the flag... But maybe that's not intuitive... ?
Otherwise... I don't have anymore idea. Or we could stop using bs namespace & do the namespace by hand and expose `__DEV__` in main module... 😢

cknitt:
Speaking of Metro - we already have a `Packager` module. Why not put it there?

sgny:
I think we could also have `global.HermesInternal`, but that's not quite straightforward so I opened issue #610 to discuss it.

MoOx:
@cknitt maybe instead if we go with a global (since we might have some stuff like Hermes flag) we could move `require` there?

MoOx:
Let's make thing simple & let's go with Global. Shouldn't be something tons of people use (!= the `Types.re` 😅)

PR #614, opened 2019-09-22
Add constructor for section type in VirtualizedSectionList

### LINKED ISSUE #605: SectionList classic RN example is impossible with current typings
The following example is a bit hard to pull off with ReasonML, making SectionList quite awkward to use:

https://facebook.github.io/react-native/docs/sectionlist.html
```
// Example 1 (Homogeneous Rendering)
<SectionList
  renderItem={({item, index, section}) => <Text key={index}>{item}</Text>}
  renderSectionHeader={({section: {title}}) => (
    <Text style={{fontWeight: 'bold'}}>{title}</Text>
  )}
  sections={[
    {title: 'Title1', data: ['item1', 'item2']},
    {title: 'Title2', data: ['item3', 'item4']},
    {title: 'Title3', data: ['item5', 'item6']},
  ]}
  keyExtractor={(item, index) => item + index}
/>
```

The problem is that it's not possible to put "title" or other custom fields into `sections`. However, practical usecases of SectionList do require that you're able to access custom datapoints in `renderSectionHeader` that are part of the sections (imagine if each section has a larger Js.t object for each SectionHeader, not just title string...).

sgny:
Should close #605 

Notes on implementation:

- `section` needs to be consumed as well as constructed, therefore I added a constructor after the type definition. This works without any issues, as tested for this case but also from prior experience.
- To avoid confusion with `extraData`, I chose the name `moreFields`.
- If the additional key `moreFields` was another parameter (`'moreFields`), then `section` would need to be typed as `section('item, 'moreFields)` which necessitates specifying `'moreFields` when creating `section` - if the key is not used. To avoid such unnecessary complexity, I chose to define a `moreFields` type instead and added a function to cast anything as that type.

### DIFF
--- a/src/components/VirtualizedSectionList.md
+++ b/src/components/VirtualizedSectionList.md
@@ -25,6 +25,7 @@ and section('item) = {
   "renderItem": option(renderItemCallback('item)),
   "ItemSeparatorComponent": option(unit => React.element),
   "keyExtractor": option(('item, int) => string),
+  "sectionData": option(sectionData),
 }
 and renderItemCallback('item) = renderItemProps('item) => React.element
 and renderSectionHeaderProps('item) = {. "section": section('item)}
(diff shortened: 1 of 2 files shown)


sgny:
Since additional fields will be relevant for data relevant to the section (such as title as in the example given in #605), I thought `sectionData` might be a better name. What do you think @MoOx ?

I also remembered to update documentation

MoOx:
I am ok with `sectionData`

PR #615, opened 2019-10-01
Add binding to useWindowDimensions hook

sgny:
This hook has been added in RN 0.61.0 and is (according to the release notes) intended to replace most uses of various `Dimensions` methods.

`Dimensions.t` type is called `DisplayMetrics` in RN, and that type has accordingly been renamed `displayMetrics`.

### DIFF
--- a/src/apis/Dimensions.md
+++ b/src/apis/Dimensions.md
@@ -5,7 +5,7 @@ wip: true
 ---
 
 ```reason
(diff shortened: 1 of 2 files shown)


MoOx:
Sorry for the delay to review all of this new awesome stuff. I will merge that as soon as I released a patch release for 0.60

PR #618, opened 2019-10-12
SectionList/VirtualizedSectionList: add section data as parameterized type

jfrolich:
The previous implementation created an opaque type for this. I think it's better to have the section data type be parameterized just like the item type. 

In the main use case you'd want to store some data for sections (for instance a title), so it is something that you'd generally use, so I think that warrants it to be typed properly.


jfrolich:
By the way, such a great effort on reason-react-native! Loving it!

sgny:
That was what I did at first (mentioned in the [PR](https://github.com/reason-react-native/react-native/pull/618#614)), but I abandoned that when testing raised issues.

Specifically, while defining sections, I was getting compiler errors (that some type variables could not be generalized) unless I specified a type for `sectionData`. Figuring that to be user hostile, I switched to the current implementation.

So I tested this again, that doesn't seem to be the case and I can't figure out why.

There are a few minor corrections to be made, I've marked those.

src/components/VirtualizedSectionList.re
```diff
@@ -2,67 +2,73 @@ include VirtualizedListElement;
 
 type sectionData;
```
sgny: This is no longer needed

src/components/VirtualizedSectionList.re
```diff
@@ -235,4 +241,4 @@ external make:
     ~testID: string=?
   ) =>
   React.element =
-  "VirtualizedSectionList";
+  "VirtualizedSectionList";
```
sgny: I suppose you haven't run `refmt` on these

src/components/VirtualizedSectionList.re
jfrolich: Removed.

src/components/VirtualizedSectionList.re
jfrolich: Yes. My `refmt` stopped working. I ran it on the new commit.

MoOx: (refmt should be run automatically via a transparent git hook)

src/components/SectionList.re
```diff
@@ -195,4 +204,4 @@ external make:
     ~testID: string=?
   ) =>
   React.element =
-  "SectionList";
+  "SectionList";
```
MoOx:
```suggestion
  "SectionList";

```

jfrolich: On the last version there was no sectionData (that is just in master I guess). But it might be breaking in other ways?

MoOx: Yeah maybe we introduced a breaking change anyway. Not sure how breaking it was.

sgny: Sorry, I didn't specifically point out the missing newline here.

@jfrolich is right, the previous PR was not in 0.60. Even a breaking change will be easy to handle as one simply has to remove the typecasting function.

MoOx: @jfrolich tiny thing but can you keep the new line please?

jfrolich: I think refmt removes it. Is it a special refmt setting?

MoOx: Not sure, I will investigate...

PR #621, opened 2019-10-19
Add missing [@bs.meth] in Event.re and others.

cknitt:
The following code

```reason
<TouchableOpacity onPress={evt => evt##stopPropagation()} />;
```

currently gives the compile error

```
  This has type:
    unit => unit
  But somewhere wanted:
    (unit => 'a) [@bs.meth]
```

This is because the `[@bs.meth]` attribute is missing on the `stopPropagation` field.

I added the missing `[@bs.meth]` attributes for the different event types and for various callback handler `remove` types.


sgny:
@MoOx @cknitt I think we can merge this without waiting for 0.61, since it's a bug fix. I can think of no valid use cases for `Event` that this PR might break, as any `Event` has to be .

However, it is breaking for specific patterns, namely, if the field is not annotated and you bind to the field first and apply the function later, the compiler warning is not triggered. For example, the below code works without issues.

```reason
  React.useEffect0(() => {
    let remove = NetInfo.IsConnected.addEventListener(`connectionChange, someFunction)##remove;
    Some(() => remove());
  });
```

Once the annotation is added, then, `remove` can be bound but cannot be called as above. Actually, I don't know how such a function can be called. All I could find is this open BuckleScript [issue](https://github.com/BuckleScript/bucklescript/issues/2953). As far as I can tell, it has not been addressed, yet. @cknitt any ideas?

In contrast, the below code triggers the error and this PR addresses that.

```reason
  React.useEffect0(() => {
    let remove = NetInfo.IsConnected.addEventListener(`connectionChange, someFunction);
    Some(() => remove##remove());
  });
```

While this PR is necessary for `Event`, I think the first block is a valid pattern. Decision to annotate also dictates one of the patterns. I think `remove` should be annotated for consistency and correctness, but I wanted to raise the issue and solicit your opinion.

cknitt:
@sgny As far as I understand, the point of the `bs.meth` attribute is to ensure that the function is invoked as a method on an object so that `this` is bound correctly.

In your first example, the function will be invoked without the `this` binding and therefore most likely not work correctly.

sgny:
Yes, according to the docs, the `@bs.meth` decorator is to take into account the possibility that the underlying `this` reference may change when the function is applied.

However, at least for the `NetInfo` event listeners, references to `this` are not used and therefore the concern doesn't arise. With or without annotation, calling `remove` works and change in connection no longer triggers the handler.

Now that `NetInfo` is going to be removed, this affects only a few components anyway. I believe we should break the pattern, even if it would have worked otherwise, as it is easy to apply the necessary changes.

sgny:
Since it's required for `Event` I'm definitely in favor.

Issue #619, opened 2019-10-14
No easy way to read from a Style.t value (using StyleSheet.flatten)


MoOx:
## The problem

We currently don't have an easy way to read style values.

## Considered solution

Adding an external to pass from `Style.t` to a raw JS object or json.

## Alternatives solutions

Code the object with all possible keys, but that can be a lot.

## Additional context

Standard RN [StyleSheet.flatten](https://facebook.github.io/react-native/docs/stylesheet#flatten) returns a raw js style object that you can read from.
Our abstract type `Style.t` doesn't allow this atm & our StyleSheet flatten method takes an array of Style.t & returns a Style.t.

sgny:
I think a neat solution is quite established by now. Whenever we have a type that needs to be created as well as consumed, we add a constructor using `@bs.obj` as well as a type definition as a `Js.t` object. This way `##` accessors can be used as usual.

If that seems reasonable, I'll fashion a PR.

MoOx:
Problem is that StyleSheet.create returns a map of ids (int) that are resolved by the style props of all components. We have to keep in mind that RN "styles" can be int | {string: 'value}... So we can't just do that that easily. Maybe we could make a method that can raise an error in case reading the object value fails? Because StyleSheet.flatten actually returns an object so maybe we should change this? And if people want to use this, they can use it as unsafe styles?!

sgny:
@MoOx I'm surprised this did not actually work, because I have been using similar implementations elsewhere with no problem.

I have already prepared a WIP PR (#622) with an alternative approach. That does indeed work 😁

MoOx:
This issue is an edge case and it seems very few people are looking for this.
Can be handled in user land.

In case you aren't sure, we can still continue discussion here.

PR #622, opened 2019-10-22
add getters for Style props

sgny:
This PR is to solicit feedback on another approach to address #619.

Basically, the idea is to define a getter for each type prop value, taking advantage of `[@bs.string]` to restrict the keys to valid props.  We would have to list all relevant polymorphic variants for each getter but that's a manageable problem. We could add an unsafe getter to handle any values set by the unsafe methods already included in `Style`.

The main drawback is that we probably have to define functions to match getter output to values such as `` `alignLeft`` but perhaps users can handle that themselves.

### DIFF
--- a/src/apis/Style.bs.js
+++ b/src/apis/Style.bs.js
@@ -1,6 +1,8 @@
 'use strict';
 
 
+var GetProp = { };
+
 function pct(num) {
   return num.toString() + "%";
 }
@@ -13,6 +15,7 @@ function rad(num) {
   return num.toString() + "rad";
 }
 
+exports.GetProp = GetProp;
 exports.pct = pct;
 exports.deg = deg;
 exports.rad = rad;
--- a/src/apis/Style.re
+++ b/src/apis/Style.re
@@ -15,11 +15,6 @@ external unsafeAddStyle: ([@bs.as {json|{}|json}] _, t, Js.t('a)) => t =
 
 external unsafeStyle: Js.t('a) => t = "%identity";
 
-module GetProp = {
-  [@bs.get_index] external float: (t, [@bs.string] [ | `flex | `fontSize]) => option(float) = "";
-  [@bs.get_index] external string: (t, [@bs.string] [ | `fontFamily | `borderStyle]) => option(string) = "";
-};
-
 type size = string;
 
 external pt: float => size = "%identity";
@@ -55,6 +50,19 @@ type transform;
 // @todo matrix
 
 external unsafeTransform: Js.t('a) => transform = "%identity";
+
+module GetProp = {
+  type key = string;
+
+  external readKey: key => string = "%identity";
+
+  [@bs.val] external keys: t => array(key) = "Object.keys";
+
+  [@bs.get_index] external color: (t, key) => Color.t = "";
+  [@bs.get_index] external float: (t, key) => float = "";
+  [@bs.get_index] external size: (t, key) => size = "";
+  [@bs.get_index] external string: (t, key) => string = "";
+};
 
 // Styles are documented here
 // https://github.com/facebook/react-native/blob/master/Libraries/StyleSheet/StyleSheetTypes.js
--- a/src/apis/Style.re
+++ b/src/apis/Style.re
@@ -15,6 +15,11 @@ external unsafeAddStyle: ([@bs.as {json|{}|json}] _, t, Js.t('a)) => t =
 
 external unsafeStyle: Js.t('a) => t = "%identity";
 
+module GetProp = {
+  [@bs.get_index] external float: (t, [@bs.string] [ | `flex | `fontSize]) => option(float) = "";
+  [@bs.get_index] external string: (t, [@bs.string] [ | `fontFamily | `borderStyle]) => option(string) = "";
+};
+
 type size = string;
 
 external pt: float => size = "%identity";
--- a/src/apis/Style.rei
+++ b/src/apis/Style.rei
@@ -9,11 +9,6 @@ external unsafeAddStyle: ([@bs.as {json|{}|json}] _, t, Js.t('a)) => t =
   "Object.assign";
 external unsafeStyle: Js.t('a) => t = "%identity";
 
-module GetProp: {
-  [@bs.get_index] external float: (t, [@bs.string] [ | `flex | `fontSize]) => option(float) = "";
-  [@bs.get_index] external string: (t, [@bs.string] [ | `fontFamily | `borderStyle]) => option(string) = "";
-};
-
 type size;
 
 [@deprecated
@@ -53,6 +48,19 @@ type transform;
 // @todo matrix
 
 external unsafeTransform: Js.t('a) => transform = "%identity";
+
+module GetProp: {
+  type key;
+
+  external readKey: key => string = "%identity";
+
+  [@bs.val] external keys: t => array(key) = "Object.keys";
+
+  [@bs.get_index] external color: (t, key) => Color.t = "";
+  [@bs.get_index] external float: (t, key) => float = "";
+  [@bs.get_index] external size: (t, key) => size = "";
+  [@bs.get_index] external string: (t, key) => string = "";
+};
 
 // Styles are documented here
 // https://github.com/facebook/react-native/blob/master/Libraries/StyleSheet/StyleSheetTypes.js
--- a/src/apis/Style.rei
+++ b/src/apis/Style.rei
@@ -9,6 +9,11 @@ external unsafeAddStyle: ([@bs.as {json|{}|json}] _, t, Js.t('a)) => t =
   "Object.assign";
 external unsafeStyle: Js.t('a) => t = "%identity";
 
+module GetProp: {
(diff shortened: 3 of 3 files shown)


MoOx:
Since this kind of usage should be an edge case, I am not sure if going down this road worth it. I would go for something more simple, even if maybe less safe. I am thinking about maintenance & code involved for this "rare use case"

sgny:
Then we can define one function `unsafeGet` or similar, which takes `string` for the prop name and returns `string` (and perhaps another which returns `float`) with which the user can deal themselves.

MoOx:
Yeah I was more thinking about something simple like that.

sgny:
This change must have taken place somewhat recently, because it definitely was an `int` when I was adding examples for `CameraRoll`, etc. In fact it was you who had drawn my attention to that.

However, it was an `int` because it was more efficient to limit data exchanged over the bridge. Perhaps with the recent rearchitecturing, that is no longer as necessary.

cknitt:
Interesting approach, I wonder if it's worth the effort though.

I have written a lot of code based on these bindings and there was not a single time when I wanted to access props of a previously created style.

For a "dirty hack", something like

```
(style->Obj.magic)##fontSize
```

should suffice.

cknitt:
If we really want an "official" way to access the props, I think I would prefer having a `getProps` function (defined as an external using `%identity%`) that returns a `Js.t` object with all the fields. (With enum fields as strings, not polymorphic variants, but still better than nothing...)

sgny:
@cknitt I wouldn't have though the compiler would consider the output of `Obj.magic` to be a `Js.t` object with a `fontSize` key. Nice trick!

@cknitt @MoOx

I have added a `keys` method to list keys included in a `Style.t` object. Making `key` an abstract type allows some safety so we can define the getters to return a value without having to wrap them in an `option`.

(`float` might need to be marked as unsafe as we cannot guarantee a `float` for just any `key`).

I have added a function to _read_ a `key` as `string` so that users can determine which keys are included, I suppose we could also add a function to simply convert `array(key)` to a `array(string)` which might be more straightforward to use.

Users can always pass the getter functions anything as a `key` using a suitable function, but something a bit rigid might help users avoid hurting themselves.

We could also dial this back to simply adding
```reason
[@bs.val] external keys: t => array(string) = "Object.keys";
```
which should still be a useful addition.

wokalski:
Sorry for chiming in but; I don't think all of this needs to be unsafe. Did you try adding some phantom types to it?

wokalski:
The pattern that’s been used for animated

sgny:
@wokalski If you're thinking of defining `key(float)`, etc which might have values like `"fontSize"`, I don't see how that would be lower maintenance than listing all allowed polymorphic variants. Could you explain what you have in mind?

wokalski:
Sorry, I thought that part of the problem was that Style can be anything - obj, array, StyleSheet key etc. so Style.t(’a) would allow to only get value for Style.t(obj) for instance

sgny:
@wokalski You are right, I wasn't thinking of that. Surely, these methods would not work with for example `array(Style.t)`. Perhaps we can reasonably expect users would simply use these methods on `Style.t` objects they have created, I don't suppose the `style` of an element can be queried.

Still defining `Style.t(simple)` (which would the output of `Style.style`) and `Style.t(array)` or something similar to make it more rigorous would not actually cause inconvenience.

cknitt:
This approach feels too complicated to me.

Why not something simple like

```reason
type styleProps = {
  .
  "color": option(Color.t),
  "fontFamily": option(string),
  "fontSize": option(float),
  "fontStyle": option(string),
  // and many many more...
};

external getStyleProps: t => styleProps = "%identity";
```

wokalski:
@cknitt I agree, with the keys it’s a bit complcated but my thing is that the function that you suggested could be restricted to using on Style.t values which are objects and not arrays or smth else.

cknitt:
@wokalski Right. As far as I can see that problem exists with either approach though (my proposed function as well as the individual accessor functions proposed by @sgny).

Of course we could check at runtime whether the style is an object or an array, but that won't be zero-cost anymore.

sgny:
@cknitt No conversion is needed. We can define
```reason
type t = {
  .
  "fontSize": option(float),
  "fontFamily": option(string),
  "color": option(Color.t)
  ...
};
```
and `Style.style` will happily output `Style.t`.

That was my first proposal, but it didn't seem to work. Now I tried it again, and it seems that I must have missed the interface file (`Style.rei`) then.

Anyway, it works as expected.

cknitt:
@sgny I would not do that, I would keep `Style.t` as an abstract type because, as @wokalski said, it might also be an array. So the type definition as a `Js.t` object would be incorrect.

With a `getStyleProps` function, this is not that bad because we can either call it `unsafeGetStyleProps` :wink: or check at runtime whether the style is an array or not.

sgny:
@cknitt I've been trying to work around that, but as we can't use `bs.unwrap` with `bs.obj` for the `style` prop in components, I haven't arrived at a solution yet. I know that has been raised before for BuckleScript, but I don't think it's being planned.

I don't  suppose there is a way to define `t(simple)` as a `Js.t` and `t(array)` as everything else. Perhaps conversion is the only option.

sgny:
Since `bs.unwrap`is unavailable with `bs.obj`, four approaches seem available to me. I think [the last one](#approach4) is the most appealing.

## Approach 1

Define `t(simple)` and `t(complex)` and define getters which only take `t(simple)` as argument.

## Approach 2 

While we cannot use `@bs.unwrap`, `@bs.ignore` works. We would need to define a new type as below:

```reason
type style(_) =
  | Single: style(Style.t)
  | Array: style(array(Style.t));
  | List: style(list(Style.t));
```

Then, the `style` prop can be specified as
```reason
    ~style: ([@bs.ignore] style('a), 'a)=?,
```
The compiler complains with the `The type of this module contains type variables that cannot be generalized:` error, for the `Animated.*` components, but we can simply wrap the declaration with `Obj.magic` thanks to @jfrolich. 

Style can be specified as
```reason
   style=(Single,someStyle)
```
or
```reason
   style=(Array,[|someStyle1, someStyle2|])
```

Admittedly, this is a bit ugly.

## Approach 3

In `Style.re`, define 
```reason
type simple = {
  .
  "fontSize": option(float),
  "fontFamily": option(string),
  "color": option(Color.t)
  ...
};
```
such that `Style.style` creates `Style.simple` and add functions to cast `Style.simple` or `array(Style.simple)`, etc. as `Style.t`. Then the `##` accessors will work only when they should.

## Approach 4

In `Style.re`, define 
```reason
type t = {
  .
  "fontSize": option(float),
  "fontFamily": option(string),
  "color": option(Color.t)
  ...
};

`style__array` and `style__list` as props to components, taking `array(Style.t)` and `list(Style.t)`, respectively. These will get rendered as `style`. Users can make mistakes and specify more than one among `style`, `style__array` and `style__list`, but I wouldn't be too worried about that. In that case, the one specified later prevails, so no runtime errors even if the style isn't quite what the user intended. Then, with multiple specifications, what would be the user's intention anyway.

We'd also remove functions such as `Style.array`.

MoOx:
I am not sure I like this ideas. Not that they are not good but I am thinking about DX & making things future proof.
None of this approach seems intuitive to me, and I am afraid this will be worst for new comers familiar with react native.
Also, in case RN StyleSheet decide to change the result provided by `create` (again) we will be in a weird situation where we will have to introduce others breaking changes (as all the solutions above will introduce breaking changes if I followed correctly).
Also, we are not sure that RN will returns us something defined in our type. JS is by nature unsafe, so if we want something safe, we can't really trust it (and for example that's why decoding json is such a pain lol). If for any reason RN decide for some reason to change a value provided to stylesheet.create, our layer won't be safe.
Since in the first place, the usage that would justify this changes is such a rare case (nobody asked for this in the first place, month after this bindings being out) it's probably doesn't worth such a big change in the DX we want to offer.

To make this simple, I would really go for a simple method that returns unsafe js that user should validate themselves. It's not a big deal & user will have to handle the mess if the value returned isn't what they expect.

sgny:
Actually what `StyleSheet.create` returns doesn't matter because we have typed that as `Js.t('a)` and we do not actually have type safety there. We could have it return `Js.Dict.t(Style.t)` but then we cannot use the `##` accessors.

I suppose anyone who has such a need, can handle it in an ad hoc manner.

MoOx:
I prefer to keep this simple as we can & "just" provide an escape hatch that will return some unsafe json a, as you said a dict(string).

MoOx:
Looks like this is an edge case nobody really seems to required (I am saying that because nobody asked for this).
Do we still want this?

jfrolich:
I have a fairly large RN app and do not (yet) see the need for this. Better use a intermediate data structure for styles if you need to inspect it.

MoOx:
@sgny are you ok if we forget about this? This can be done in user land anyway right?

sgny:
@MoOx Sure, it's been a useful exercise to investigate as it may translate to something else, but I haven't really needed anything like this either.

PR #626, opened 2019-11-11
update type for DrawerLayoutAndroid drawerPosition prop

sgny:
Constants `Left` and `Right` are no longer defined, the prop accepts either "left" or "right" as `string`.


sgny:
Quite recent, only as of 0.61.0.

MoOx:
Ok I missed that in RN 0.61 release notes. Merging & will cut a release tomorrow with all the latests stuff

PR #627, opened 2019-11-18
Add touchSoundDisabled property binding to Button and Touchable components

stuartkeith:
Hi! I noticed the obscure touchSoundDisabled property isn't accessible for the Button and Touchable components, so I went ahead and added it. Thanks!


MoOx:
Thanks @stuartkeith ! Will be in upcoming release

PR #631, opened 2019-12-13
[WIP] Cleanup codebase from Js.t thanks to bs-platform@7 object as records

### LINKED ISSUE #629: Support bs-platform@7
## The problem

Is `bs-platform@7.0.1` on the roadmap?

## Considered solution

I started the conversion here: https://github.com/idkjs/reason-react-native/tree/bs_platform_701

Tested on the template here: https://github.com/idkjs/reason-react-native-bs701

Of note, I have not completed [VirtualizedList.re](https://github.com/idkjs/reason-react-native/blob/bs_platform_701/src/components/VirtualizedSectionList.re) yet since there is some trickery with `section` in that file. 

I also have not converted the `.md`'s yet. I wanted to get some feedback before continuing. 

All of the changes are to implement the new syntax. A few examples are:

registry
photoIdentifier
layoutEvent

How do you test this repo? Is there a way to find out if I have broken anything?

Should I make a pull request?

MoOx:
Will closes #629  when finished.


idkjs:
I just made some changes converting `Js.Nullable.t` to `option`. The code compiles but I am not sure about it. How do we go about this. Do I ask for  a review? Also, how have you been testing your code? Thanks.

cknitt:
@idkjs Please be careful regarding `Js.Nullable.t`. The `option` value `None` only covers `undefined`, not `null`. If an API really returns a `null` value, replacing `Js.Nullable.t` with `option` is incorrect and will lead to crashes/unpredictable behavior.

So you would need to check each case individually to be sure it is safe and the API never returns `null` before converting it to `option`.

src/apis/ActionSheetIOS.re
```diff
@@ -1,36 +1,48 @@
...
-  options =
-  "";
+type options = {
+  options: array(string),
+  cancelButtonIndex: option(int),
+  destructiveButtonIndex: option(int),
+  title: option(string),
+  message: option(string),
+  tintColor: option(Color.t),
+};
```
cknitt: Only values returned from an API (e.g. as callback params) should be converted to records. Values passed to an API (options) should not be converted to records, because records cannot have optional fields.

With the current bindings, you can do

```reason
let options = ActionSheetIOS.options(~options=[|"A", "B", "C"|], ());
```

(all the other props being optional). If converting to records you **won't** be able to do

```reason
let options = { ActionSheetIOS.options: [|"A", "B", "C"|] };
```

but would have to supply all the remaining props with value `None` which is very inconvenient.

cknitt: In other words, we should be using records where we are currently using `Js.t` objects, but not where we are currently using abstract types.

src/apis/CameraRoll.re
```diff
@@ -23,43 +23,39 @@ external getPhotosParams:
...
+  width: float,
+  isStored: option(bool),
+  playableDuration: float,
+}
+and location = {
+  latitude: option(float),
+  longitude: option(float),
+  altitude: option(float),
+  heading: option(float),
+  speed: option(float),
```
cknitt: I think we should just reorder the type definitions instead of using `and`.

src/apis/ActionSheetIOS.re
idkjs: > So you would need to check each case individually to be sure it is safe and the API never returns null before converting it to option.

I am happy to do this. How would you go about doing that?

jfrolich: @cknitt: You can create the same constructor function (that currently  creates objects) to  create records with default value None.

MoOx: @jfrolich Can you show an example to be sure about what you mean?

MoOx:
@idkjs are you still motivated to continue the effort here or do you want help with this?

idkjs:
@moox I am motivated. Saw a message where you said you had forked and merged. Thought you were done. Confused about what happened here. 

I am in. I need to review the comments and get back on it. Sorry for the misunderstanding

MoOx:
No problem. I just created this PR. I think I merged some commits into this repo but it's not "in master" yet (hence, this PR).

You can push commits on this branch, you should have the proper rights :)

idkjs:
Ok. Will get back on this. Thanks for the heads up.

src/apis/ActionSheetIOS.re
jfrolich: Basically have a function with optional labels that construct the record.

cknitt: @jfrolich @MoOx 

There is a subtle difference between optional params in a `bs.obj` creation function and `option` fields in a record. With `bs.obj`, the optional fields will be omitted from the JS object if not specified.

However,

```re
let options = {
  options: [| "A", "B" |],
  cancelButtonIndex: None,
  destructiveButtonIndex: None,
  title: None,
  message: None,
  tintColor: None,
};
```
will give you the JS object

```js
{
  options: [ 'A', 'B' ],
  cancelButtonIndex: undefined,
  destructiveButtonIndex: undefined,
  title: undefined,
  message: undefined,
  tintColor: undefined,
};
```

So

1. This difference in behavior can lead to subtle bugs in some situations,
2. what you construct with the `bs.obj` creation function strictly speaking has *not* the same type as the record, as a record always needs to have all the fields, and
3. this may lead to code size issues.

I am therefore very hesitant to provide these record types for "input" params.

jfrolich: @cknitt: Yes there might be an issue with javascript code that will check for existence of a key instead of the null value. However I think that will be an edge case (need to test though!)

In the case when there is a check if keys exists on an object it's also possible to convert the record to Js.Json.t and only keep the non-null record fields (or just keep the `bs.obj` function). But when the interop works with records it provides quite an ergonomic interface.

idkjs: Hi folks. Sorry for the prolonged absence. I am back now. If I am reading this correctly, we were considering converting our `input` types back to `bs.obj`'s. Do we still want to do this?

jfrolich: Great you are back @idkjs! I think it could be records unless input API's rely on existence of keys.

idkjs: I can confirm that using the record syntax gets requires the following `option` code:

```reason
let options: ReactNative.ActionSheetIOS.options = {
  options: buttons,
  cancelButtonIndex: Some(4),
  destructiveButtonIndex: Some(3),
  tintColor: Some("green"),
  title:None,
  message:None
};
```


Which compiles to 
```js
var options = {
  options: buttons,
  cancelButtonIndex: options_cancelButtonIndex,
  destructiveButtonIndex: options_destructiveButtonIndex,
  title: undefined,
  message: undefined,
  tintColor: options_tintColor
};
```

and runs fine. Don't know if this answers the question. I think going with records is the way to go for what it is worth. At the very least, you will always be aware of what the `js` code is doing. Examples running against `bs.obj` [idkjs/ActionSheetIOSDemo](https://github.com/idkjs/ActionSheetIOSDemo). Against records in a minute.

idkjs:
Regarding these conflicts, Its just a question of deciding which way to go, `bs.obj` or `records`, correct? In the mean time, I will resolve them back to `bs.obj` until a decision is made.

jfrolich:
Why keep `bs.obj`?

idkjs:
@jfrolich I'm for records. The discussion is about which way to go, isn't it?

src/apis/CameraRoll.re
idkjs: Done.

src/apis/ActionSheetIOS.re
idkjs: What do you folks want to do with this? `bs.obj` or `records`?

sgny:
Consider the `Style.t` type. If we go for record, each style definition will have 101 keys, the great majority with value `undefined`.  Unless some key must exist, even if it's null or undefined, I think `bs.obj` is the cleaner solution.

jfrolich:
> Consider the `Style.t` type. If we go for record, each style definition will have 101 keys, the great majority with value `undefined`. Unless some key must exist, even if it's null or undefined, I think `bs.obj` is the cleaner solution.
> 
> Worse yet, we'd lose the `bs.string` decorator and have to resort to `bs.inline`. Having to use the latter for return values is painful enough, I don't think we want that for something like `Style.t` for example.

Agree. I think at some point Bucklescript can add more functionality to also have this use case covered for records. For style the type is an implementation detail and not user-facing anyway so it doesn't really matter.

idkjs:
> Agree. I think at some point Bucklescript can add more functionality to also have this use case covered for records. For style the type is an implementation detail and not user-facing anyway so it doesn't really matter.

@jfrolich what do you mean here by `implementation detail`? And the fact that it is not user facing means that we can use records because the end user wont ever see it?

jfrolich:
Sorry, I meant that even when Style.t is still a `Js.t` it's not user facing, because for the user it's an opaque type. So imo not really necessary to make it a record type if there are some downsides to it at the moment (much unnecessary generated code).

MoOx:
Too lazy to review all this thread. What is the status of this PR?

Arnarkari93:
I am willing to give a helping hand to push this forward. I think @cknitt makes a compelling case of keep using `[@bs.obj]` for objects passed to react-native.

src/apis/ActionSheetIOS.re
MoOx: I think we definitely don't want to type

```reason
let options: ReactNative.ActionSheetIOS.options = {
  options: buttons,
  cancelButtonIndex: None,
  destructiveButtonIndex: None,
  tintColor: None,
  title:None,
  message:None
};
```

When you can just ask people to write this

```reason
let options = ActionSheetIOS.options(~options=buttons, ());
```

So we must keep `bs.obj` for now.

MoOx:
@Arnarkari93 feel free to make another PR & bump to bs-platform 7.2 :)

sgny:
I resolved the conflicts to make sure we do not add `textInputEvent` back into the repo. I've also chosen to remove changes to `yarn.lock` that were due to an early v7 release of `bs-platform`. Any such dependencies should be updated with a newer release anyway.

MoOx:
Closing in favour of #666 that looks more promising. Thanks for everyone involved in this WIP.

PR #636, opened 2020-01-17
Add accessibilityRole and accessibilityLevel props

### LINKED ISSUE #635: Support for h1-6 tags on react-native-web
Allow `accessibilityRole` prop on Text, and add `aria-level` prop. 

More info:
https://github.com/necolas/react-native-web/commit/1b172319b9ea257830a82cb6800ad3e7a746190c

Naturalclar:
Closes #635 

Added `accessibilityRole` and `accessibilityLevel` props to Text component.

accessibilityLevel prop does not exist in `react-native` but will work in `react-native-web`

Ref: https://github.com/necolas/react-native-web/pull/402/files

### DIFF
--- a/src/components/Text.re
+++ b/src/components/Text.re
@@ -37,7 +37,7 @@ external make:
                           | `toolbar
                         ]
                           =?,
-    ~accessibilityLevel: int=?,
+    ~ariaLevel: int=?,
     ~allowFontScaling: bool=?,
     ~ellipsizeMode: [@bs.string] [ | `clip | `head | `middle | `tail]=?,
     ~numberOfLines: int=?,
--- a/src/components/Text.re
+++ b/src/components/Text.re
@@ -19,7 +19,6 @@ external make:
                           | `adjustable
                           | `imagebutton
                           | `header
-                          | `heading
                           | `summary
                           | `alert
                           | `checkbox
--- a/src/components/Text.re
+++ b/src/components/Text.re
@@ -8,44 +8,46 @@ external make:
     ~accessible: bool=?,
     ~accessibilityHint: string=?,
     ~accessibilityLabel: string=?,
+    // `accessibilityRole` communicates the purpose of a component to the user of an assistive technology.
+    // roles that are specific for react-native-web are also included:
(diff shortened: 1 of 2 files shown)


jfrolich:
Wow that was fast!

Naturalclar:
@jfrolich I noticed after I sent the PR that the actual props for `accessibilityLevel` is supposed to be `aria-level` 
I'm trying to find out whether I can set prop names with hyphen as a prop.

Naturalclar:
If I'm understanding [this](https://reasonml.github.io/reason-react/docs/en/invalid-prop-name) correctly,
ariaLevel will be translated into `aria-level` under `reason-react`
So maybe, setting an ariaLevel props will get the job done?

jfrolich:
Perhaps in the same PR we can also add the extra accessibility roles that are valid in `react-native-web`:

https://github.com/necolas/react-native-web/blob/36dacb2052efdab2a28655773dc76934157d9134/packages/react-native-web/src/modules/AccessibilityUtil/propsToAccessibilityComponent.js#L13-L23.

src/components/Text.re
```diff
@@ -8,6 +8,47 @@ external make:
     ~accessible: bool=?,
     ~accessibilityHint: string=?,
     ~accessibilityLabel: string=?,
+    ~accessibilityRole: [@bs.string] [
+                          | // below are additional roles added for react-native-web https://github.com/necolas/react-native-web/blob/36dacb2052efdab2a28655773dc76934157d9134/packages/react-native-web/src/modules/AccessibilityUtil/propsToAccessibilityComponent.js#L13-L23
```
MoOx: You can see that the compiler is moving the comment up there (which sucks but that's a fact...)
What about instead keeping everything Alpha ordered with just `// web` after web only values?

jfrolich: Or a general comment for the accessibility roles enumerating all the variants that apply to react-native-web. This is probably also better for tools that extract documentation from the source.

Naturalclar: @MoOx @jfrolich 
Thanks for the suggestion! :)
I've listed the strings in alphabetical order, and added comment about react-native-web specific roles.

Issue #643, opened 2020-02-01
Animated component refs do not point to the component


benadamstyles:
## The problem

Refs, when set from animated components (e.g. `Animated.FlatList`), are incorrectly typed. In React Native, the refs for these `Animated` components have a `_component` property which points to the actual component. This is not accounted for in the Reason React Native types, so if you call a method on a ref, your app crashes.

### How to reproduce

```reason
let make = () => {
  let flatListRef = React.useRef(Js.Nullable.null);

  <>
    <Animated.FlatList
      ref=flatListRef
      data={[||]}
      keyExtractor={(_item, index) => index->string_of_int}
      renderItem={props => <View />}
    />
    <TouchableOpacity
      onPress={_ =>
        flatListRef
        ->React.Ref.current
        ->Js.toOption
        ->Belt.Option.forEach(flatList =>
            Animated.FlatList.(
              flatList
              ->scrollToOffset(
                  scrollToOffsetParams(
                    ~offset=Device.screenWidth,
                    ~animated=true,
                    (),
                  ),
                )
            )
          );
      }
    />
  </>;
};
```

## Expected behavior

This code is type-correct, so tapping the button should not throw.

## Actual behavior

App crashes with `scrollToOffset is not defined` or similar. This is because on `Animated` components, `ref`s should point to a property on the ref called `_component`.

## Additional context

I have tried to make a PR to fix this but I need some advice – the way refs are set up for these `Animated` components, using [`VirtualizedListElement.re`](https://github.com/reason-react-native/reason-react-native/blob/master/src/elements/VirtualizedListElement.re), makes it very complicated for me to understand how we could customize the `ref` prop for just these components.

MoOx:
Maybe the way we currently define animated components isn't the best thing we can do as ref is by definition wrong (otherwise you would not have faced a runtime error).

Maybe it's time to stop avoiding some copy/paste & make "standard" bindings for `Animated.*` with adjustment for ref. 
We could offer an `Animated._component` accessor (or a better name) so people can do normal ref access like they do with normal RN.

What do you think? 

Poke @cknitt @sgny

sgny:
@benadamstyles @MoOx 

I think we can handle this rather easily if I have the right notion. I'll quickly put together a PR if it works.

@MoOx No need to start copy/pasting just yet. `createAnimatedComponent` wraps these components, so we need to access the actual component if we want to use any relevant methods.

sgny:
[...]

PR #644, opened 2020-02-03
Revise definition of ref for Animated components

sgny:
Here's one way to address #643. JS output is unchanged, but the ref is wrapped and has to be accessed with an appropriate function.

### DIFF
--- a/src/apis/Animated.bs.js
+++ b/src/apis/Animated.bs.js
@@ -95,29 +95,37 @@ function reset(prim) {
 
 var StyleProp = /* module */[];
 
+FlatList$ReactNative.Make(/* module */[]);
+
 var make = ReactNative.Animated.createAnimatedComponent(ReactNative.FlatList);
 
 var FlatList = /* module */[/* make */make];
 
+Image$ReactNative.Make(/* module */[]);
+
 var make$1 = ReactNative.Animated.createAnimatedComponent(ReactNative.Image);
 
-var $$Image = /* module */[
-  /* Source */Image$ReactNative.Source,
-  /* DefaultSource */Image$ReactNative.DefaultSource,
-  /* make */make$1
-];
+var $$Image = /* module */[/* make */make$1];
+
+ScrollView$ReactNative.Make(/* module */[]);
 
 var make$2 = ReactNative.Animated.createAnimatedComponent(ReactNative.ScrollView);
 
 var ScrollView = /* module */[/* make */make$2];
 
+SectionList$ReactNative.Make(/* module */[]);
+
 var make$3 = ReactNative.Animated.createAnimatedComponent(ReactNative.SectionList);
 
 var SectionList = /* module */[/* make */make$3];
 
+Text$ReactNative.Make(/* module */[]);
+
 var make$4 = ReactNative.Animated.createAnimatedComponent(ReactNative.Text);
 
 var $$Text = /* module */[/* make */make$4];
+
+View$ReactNative.Make(/* module */[]);
 
 var make$5 = ReactNative.Animated.createAnimatedComponent(ReactNative.View);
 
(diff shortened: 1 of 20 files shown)


sgny:
I removed `Obj.magic` where it wasn't needed, even if it would cause no additional cost.

sgny:
We could also define
```reason
type animatedRef('ref) = {
  .
  "_component": 'ref
};
```
to access the `_component` prop, but I understand that use has been deprecated in favor of the `getNode` method.

MoOx:
Nice changes! Not sure like you said that relying on `_component` is necessary, especially if we already offers an "official" way to do the same

cknitt:
Looks nice! Quite a lot of effort though for a problem that will be gone with 0.62 anyway. So we will then roll back these changes?

Why are there still some `Obj.magic` in some places? Are they required for some components but not for others?

sgny:
@MoOx I haven't checked if `_component` key still exists, but I don't think it's necessary as you said.

@cknitt 
I was hoping this could be handled while including module * in Animated.*, but I don't think that's possible yet with BuckleScript, so I took this approach.

We should roll these changes back because at some point `getNode` will be removed and code using it will cause errors. We could redefine `getNode` to wrap the external definition with a version check to avoid throwing deprecation warnings or add another function to simply cast `animatedRef('ref)` to `'ref`.

As `FlatList` and `SectionList` take parametrised types for some props, the compiler complains without `Obj.magic`. We could include `type item` alongside `type t` in the signature for modules passed to `FlatList.Make` or `SectionList.Make` but that would be more cumbersome for users.

sgny:
@MoOx @cknitt 
The last piece is a way to pass an `animatedRef`, so we should also add something like
```reason
external createAnimatedRef: React.Ref.t(Js.nullable('ref)) =>
   animatedRef(React.Ref.t(Js.nullable('ref))) = "%identity";
```
Perhaps it makes sense to be more precise with `getNode` as well, as below:
```reason
[@bs.send] external getNode: (animatedRef(React.Ref.t(Js.nullable('ref))), unit) =>
   React.Ref.t(Js.nullable('ref)) = "getNode";
```
Any ideas?

sgny:
@MoOx @cknitt 

I figured it would be easier to simply type cast some `ref` with such a function as above, but we could alternatively define
```reason
[@bs.module "react"] external createAnimatedRef: unit => animatedRef(Ref.t(Js.nullable('a))) = "createRef";
[@bs.module "react"] external useAnimatedRef: 'value => animatedRef(Ref.t('value)) = "useRef";
```
or even add the below to `Animated`

```reason
module Ref = {
  [@bs.module "react"] external createRef: unit => animatedRef(Ref.t(Js.nullable('a))) = "createRef";
  [@bs.module "react"] external useRef: 'value => animatedRef(Ref.t('value)) = "useRef";
};
```
to avoid namespace conflicts for those who prefer to `open` various modules.

MoOx:
@sgny I think we should use the thing that would be more intuitive for end user. I have the feeling that the solution that rely on `getNode` definition is what would make it transparent.

sgny:
> ```reason
> [@bs.send] external getNode: (animatedRef(React.Ref.t(Js.nullable('ref))), unit) =>
>    React.Ref.t(Js.nullable('ref)) = "getNode";
> ```

@MoOx I felt this would be easier to understand as well. So basically users have to use the additional methods `createAnimatedRef` and `getNode` to, respectively, pass and access `ref` to `Animated.*` components. I made the signature for `getNode` more explicit (as above) with this latest commit.

MoOx:
That's quite a big change in here. Do we really want this? 
We could also get it and revert it (to simplify codebase) as soon as we tackle RN 0.62?

sgny:
I held off a bit to get other opinions, but this can be undone with a simple `git revert`.

sgny:
@MoOx @cknitt 

It will be a while until 0.62 is out and stable, should we not have this interim fix until then?

MoOx:
0.62 is out, do we want this? :D

MoOx:
I guess not :D

PR #646, opened 2020-02-08
feat: add mouse event for react-native-web

### LINKED ISSUE #645: onMouseEnter/onMouseLeave prop for react-native-web
Earlier mentioned here: (https://github.com/reason-react-native/reason-react-native/issues/125).

I think it would be good to have this prop. With documentation that it only works for `react-native-web`.

Naturalclar:
Closes #645 

Adding onMouseEnter, onMouseLeave props for View, Text, TextInput

ref: https://github.com/necolas/react-native-web/blob/master/packages/react-native-web/src/exports/View/filterSupportedProps.js
https://github.com/necolas/react-native-web/blob/master/packages/react-native-web/src/exports/Text/index.js
https://github.com/necolas/react-native-web/blob/master/packages/react-native-web/src/exports/View/index.js
https://github.com/necolas/react-native-web/blob/master/packages/react-native-web/src/exports/TextInput/index.js

### DIFF
--- a/src/components/Text.md
+++ b/src/components/Text.md
@@ -74,7 +74,11 @@ external make:
     ~minimumFontScale: float=?,
     ~suppressHighlighting: bool=?,
     ~value: string=?,
-    ~children: React.element=?
+    ~children: React.element=?,
+    // React Native Web Props
+    ~onMouseEnter: ReactEvent.Mouse.t => unit=?,
+    ~onMouseLeave: ReactEvent.Mouse.t => unit=?
+
   ) =>
   React.element =
   "Text";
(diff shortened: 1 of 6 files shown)


MoOx:
While doing this, maybe we could handle all mouse event? See supported ones here https://github.com/necolas/react-native-web/blob/5b7f584f191d8fb6e5c92fa560c4e6babca7559f/packages/react-native-web/src/exports/View/filterSupportedProps.js#L58-L65

Naturalclar:
@MoOx 
Thanks! I've added the other mouse events as well :+1:

jfrolich:
Would be great if this can ship soon! 🚀

PR #650, opened 2020-03-03
Press-event, href and target

jfrolich:
This adds `href` and `target` for React-Native-Web (pretty important for accessibility and SEO). Additionally it fixes a bug where the press event was not passed in a `Text` component. This is related, because when you have client routing you'd want to provide a `href` and do a `preventDefault` on the press event.

### DIFF
--- a/src/components/ScrollView.re
+++ b/src/components/ScrollView.re
@@ -137,9 +137,7 @@ external make:
     ~shouldRasterizeIOS: bool=?,
     ~style: Style.t=?,
     ~testID: string=?,
-    ~children: React.element=?,
-    // temporary react-native-web-prop
-    ~useWindowScrolling: bool=?
+    ~children: React.element=?
   ) =>
   React.element =
   "ScrollView";
--- a/src/components/ScrollView.re
+++ b/src/components/ScrollView.re
@@ -137,7 +137,9 @@ external make:
     ~shouldRasterizeIOS: bool=?,
     ~style: Style.t=?,
     ~testID: string=?,
-    ~children: React.element=?
+    ~children: React.element=?,
+    // temporary react-native-web-prop
+    ~useWindowScrolling: bool=?
   ) =>
   React.element =
   "ScrollView";
--- a/src/components/Text.re
+++ b/src/components/Text.re
@@ -72,8 +72,8 @@ external make:
     ~rel: [@bs.string] [
             | `alternate
             | `author
-            | `dnsPrefetch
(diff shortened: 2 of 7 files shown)


src/components/ScrollView.re
```diff
@@ -137,7 +137,9 @@ external make:
     ~shouldRasterizeIOS: bool=?,
     ~style: Style.t=?,
     ~testID: string=?,
-    ~children: React.element=?
+    ~children: React.element=?,
+    // temporary react-native-web-prop
+    ~useWindowScrolling: bool=?
```
MoOx: This thing doesn't seems to be in RNW repo. Can you move this back & keep it in a patch if you use a specific hacked RNW version? :D

src/components/Text.re
```diff
@@ -55,8 +55,8 @@ external make:
     ~ellipsizeMode: [@bs.string] [ | `clip | `head | `middle | `tail]=?,
     ~numberOfLines: int=?,
     ~onLayout: Event.layoutEvent => unit=?,
-    ~onLongPress: unit => unit=?,
-    ~onPress: unit => unit=?,
+    ~onLongPress: Event.pressEvent => unit=?,
+    ~onPress: Event.pressEvent => unit=?,
```
MoOx: 😚

src/components/ScrollView.re
MoOx:
```suggestion
    ~children: React.element=?
```

MoOx: I have removed it, be sure to add it in your personal patch!

jfrolich: Sorry forget this PR was on my master branch! This was a personal patch for a PR that solves a current issue in react-native-web with window scrolling.

Normally I create a branch specifically for PR's will do again for the next PR :D

src/components/Text.re
```diff
@@ -69,6 +69,25 @@ external make:
...
+            | `next
+            | `pingback
+            | `preconnect
+            | `prefetch
+            | `preload
+            | `prerender
+            | `prev
+            | `search
+            | `stylesheet
+          ]
```
MoOx, line 87: @jfrolich curious to know where you find this list :) 
I saw some missings (eg: nofollow, noopenner & more...) and some incorrect (eg: preload, stylesheet and some others). I am referring to https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/rel (2nd column)

PR #657, opened 2020-03-26
[0.62] add accessibilityState prop and constructor for related object

sgny:
This PR adds the replacement prop for `accessibilityStates` removed in 0.62.

Before merging, I'd appreciate some feedback on the API. I have chosen to add a module called `Accessibility` and a type called `state` with a constructor of the same name. I plan to add another type called `value` and again a constructor of that name to handle the `accessibilityValue` prop which also seems to be missing from our bindings.

The JS API necessitates some rather unpleasant choices for the `checked` field, which can take either a `bool` or the string `"mixed"`. Since `unwrap` is still unavailable, this PR defines a type `checked` in `Accessibility`, a function to type cast `bool` to `checked` and uses `bs.inline` to handle `"mixed"`.

### DIFF
--- a/src/components/ActivityIndicator.md
+++ b/src/components/ActivityIndicator.md
@@ -45,6 +45,7 @@ external make:
                           | `imagebutton
                         ]
                           =?,
+    ~accessibilityState: Accessibility.state=?,
     ~accessibilityTraits: array(AccessibilityTrait.t)=?,
     ~accessibilityViewIsModal: bool=?,
     ~accessible: bool=?,
--- a/src/components/ActivityIndicator.re
+++ b/src/components/ActivityIndicator.re
@@ -38,6 +38,7 @@ external make:
                           | `imagebutton
                         ]
                           =?,
+    ~accessibilityState: Accessibility.state=?,
     ~accessibilityTraits: array(AccessibilityTrait.t)=?,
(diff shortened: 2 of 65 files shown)


sgny:
@MoOx 

I agree, zero cost does not seem possible with a significantly nicer API, at least for now while `@bs.unwrap` is unavailable for use with `@bs.obj`. `@bs.unboxed` does not seem to improve on our `"%identity"` tricks in this regard.

With that said, we could have also define

```reason
type checked = bool;

let checkedMixed: checked = "mixed"->Obj.magic;
```
as we have already use elsewhere in this repo. This is again quite low cost.

Then, instead of inlining as `string`, we could use inlining as `bool`, as below:

```reason
 [@bs.inline]
 let checkedTrue = true;

 [@bs.inline]
 let checkedFalse = false;
```
An appropriate interface would be needed, but perhaps this version would be nicer?

MoOx:
Would `let mixed: bool = "mixed"->Obj.magic;` be possible (so no checked/unchecked + interface) required?

sgny:
@MoOx It's certainly possible, but then `mixed` can be passed to all fields.

MoOx:
Would you want to pass mixed as a boolean everywhere? But yeah this could lead to a runtime error. I am ok for your last proposal then.

PR #661, opened 2020-03-27
[0.62] add module for DevSettings API

sgny:
Defined the `reason` type as `string` for the binding to be self-documenting.

### DIFF
--- a/src/apis/DevSettings.re
+++ b/src/apis/DevSettings.re
@@ -1,0 +1,10 @@
+type reason = string;
+
+[@bs.scope "DevSettings"] [@bs.module "react-native"]
+external addMenuItem: (string, unit => 'a) => unit = "addMenuItem";
+
+[@bs.scope "DevSettings"] [@bs.module "react-native"]
+external reload: unit => unit = "reload";
+
+[@bs.scope "DevSettings"] [@bs.module "react-native"]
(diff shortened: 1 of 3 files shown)


src/apis/DevSettings.re
```diff
@@ -0,0 +1,7 @@
+type reason = string;
+
+[@bs.scope "DevSettings"] [@bs.module "react-native"]
+external addMenuItem: (string, unit => 'a) => unit = "addMenuItem";
+
+[@bs.scope "DevSettings"] [@bs.module "react-native"]
+external reload: reason => unit = "reload";
```
MoOx: Per code, it seems to be optional https://github.com/facebook/react-native/blob/0b9ea60b4fee8cacc36e7160e31b91fc114dbc0d/Libraries/Utilities/DevSettings.js#L44 and it seems in some case it can be ignored.
The reason is also not mentioned in docs https://reactnative.dev/docs/devsettings#reload

Should we have reload & reloadWithReason?

sgny: I had both methods at first, but figured `reload("")` was no worse than `reload()` and allow avoiding a second `reloadWithReason` method.

I revised to bring the PR more in line with our usual practice, as you suggested.

PR #664, opened 2020-03-27
[0.62] DevSettings: add documentation, revise signatures

sgny:
I made some slight revisions to function signatures, adding type aliases (`title` as alias for `string` and `handler` as alias for `unit => unit`) to make it easier to discover how to use these methods without having to refer to documentation. While the handler can return `mixed` in JS, that does not really provide any benefit, so I have revised that so that `handler` does not need a type parameter, such as `handler('a)`.

I have prepared documentation for the module, as well.

### DIFF
--- a/src/apis/DevSettings.md
+++ b/src/apis/DevSettings.md
@@ -1,18 +1,50 @@
 ---
 id: apis/DevSettings
 title: DevSettings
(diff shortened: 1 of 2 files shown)


MoOx:
Awesome!

PR #668, opened 2020-04-02
[0.62] handle deprecated or removed components and APIs

sgny:
With this PR, components removed from React Native will be removed from these bindings. Components which are deprecated and set to be removed in the future remain, but links to replacement packages and relevant bindings have been added to their documentation.

`*` in tables below indicates availability of bindings for the replacement package.

**Removed Components**
| Module  |   |
|---|:-:|
| ToolbarAndroid | |
| ViewPagerAndroid | * |
| WebView | * |

**Removed APIs**
| Module  |   |
|---|:-:|
| AlertIOS | * |
| CameraRoll | * |
| GeoLocation | * |
| ImageEditor | * |
| NetInfo | * |
| TimePickerAndroid | * |

**Deprecated Components**
| Module  |   |
|---|:-:|
| CheckBox | * |
| MaskedViewIOS | |
| Picker | * |
| PickerIOS | * |
| ProgressBarAndroid | |
| ProgressViewIOS | |
| SegmentedControlIOS | |
| Slider | * |

**Deprecated APIs**
| Module  |   |
|---|:-:|
| AsyncStorage  | * |
| Clipboard  | * |
| DatePickerAndroid | * |
| DatePickerIOS | * |
| ImagePickerIOS | * |
| PushNotificationIOS | * |


src/apis/AsyncStorage.md
```diff
@@ -4,6 +4,15 @@ title: AsyncStorage
 wip: true
 ---
 
+`AsyncStorage` has been deprecated and will be removed from React Native in the
```
Arnarkari93, line 7: Do we want to take this even further and deprecate the `AsyncStorage` as well, by adding a `[@ocaml.deprecated]` in the `AsyncStorage.rei` file? Then users will get the message right in the compiler.

sgny: I proposed the same in #594 but @MoOx rightly pointed out that React Native has deprecation warnings, so users are presumably aware that they are using a deprecated component or API.

We could have deprecation warnings pointing to replacement bindings. I suppose they are a bit annoying when you cannot migrate to another package, but can be turned in `bsconfig.json`.

MoOx: You have RN deprecating warnings in simulator + in terminal, pretty sure that's more than enough.

MoOx: Already discussed in an other PR, I think RN does what it need to warn user already by its own.

PR #669, opened 2020-04-04
[security] update lint-staged

sgny:
`kind-of` version 6.03 is listed among `reason-react-native`'s resolved dependencies and accordingly the security advisory is not triggered. Nevertheless, it does depend on quite a few past versions of `kind-of` as well as other rather old packages, all of which we might as well do without. `yarn list` output becomes 704 lines long, instead of `949`.

Apparently, `lint-staged` since version 10 should not be run on empty repos, but that should not be an issue for any repo already present or created from our template.

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,7 +37,7 @@
   "devDependencies": {
     "bs-platform": "^5.0.4",
     "husky": "^1.3.1",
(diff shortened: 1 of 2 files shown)


MoOx:
Great, I will apply this to all repo!

PR #672, opened 2020-04-11
update Event documentation

sgny:
This PR is to revise `Event.md` as it has become out-of-sync with the new type definitions.


MoOx:
We will need to be more careful indeed!

PR #675, opened 2020-04-11
Fix Animated.ValueXY.getTranslateTransform(value) binding

benadamstyles:
As explained [here](https://animationbook.codedaily.io/get-translate-transform/), `getTranslateTransform` does not return an object but an array of transform objects.

### DIFF
--- a/src/apis/Animated.re
+++ b/src/apis/Animated.re
@@ -202,10 +202,6 @@ module ValueXY = {
 
   [@bs.obj] external jsValue: (~x: float, ~y: float) => jsValue;
 
-  type translateTransform = {
-    translateX: Value.t,
-    translateY: Value.t,
-  };
   type layout = {
     left: Value.t,
     top: Value.t,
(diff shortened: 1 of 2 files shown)


benadamstyles:
I just made this PR on the GitHub UI so there are probably scripts I need to run?

src/apis/Animated.re
```diff
@@ -214,7 +210,7 @@ module ValueXY = {
   external create: jsValue => t = "ValueXY";
   [@bs.send] external getLayout: t => layout = "getLayout";
   [@bs.send]
-  external getTranslateTransform: t => translateTransform =
+  external getTranslateTransform: t => array(ReactNative.Style.transform) =
```
sgny, line 213: Please revise as `external getTranslateTransform: t => array(Style.transform) =`

sgny:
Thanks for catching the issue and preparing the PR. I checked the RN codebase and your definition fits the implementation.

I'm guessing you've used this definition in your own project already and it works as expected, but please do confirm if you have.

Finally, there are no changes to the output, so no need to run any scripts, but do update the definition in the documentation as well.

benadamstyles:
Thanks, yep this works as expected in my own project. Will update the docs in this branch shortly

MoOx:
Thanks @benadamstyles

PR #628, opened 2019-12-08
resolves all bucklescript inferred name warning

idkjs:
This pull request resolves all the `inferred name` errors generated by `warning 105` from the compiler.

Example:
```sh
Warning 105: BuckleScript warning: fetch : the external name is inferred from val name is unsafe from refactoring when changing value name
```

PR #678, opened 2020-04-13
enable warning 105

sgny:
Thanks to @idkjs we no longer need to disable warning 105, so I believe it makes sense to revert my commit that introduced that change.

### DIFF
--- a/bsconfig.json
+++ b/bsconfig.json
@@ -16,6 +16,6 @@
       "subdirs": true
     }
   ],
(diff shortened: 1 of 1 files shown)


idkjs:
So I dont have to do anything here, then?

MoOx:
You already did :D

idkjs:
Damn, ok. I just got an email about it right before I wrote that. Still figuring out this git thing. Thanks.

MoOx:
You have indeed being mentioned in the description of this PR. It was because you recently contributed here => #628

PR #681, opened 2020-04-22
Fix layout-event, and some refactoring

jfrolich:
I refactored our app to the new `reason-react-native`. Great work!

I found a little bug where the layout event is incorrectly typed.

When using records (especially when field names are shared) it's a good practice to wrap the type in a module, otherwise inference doesn't work and annotating becomes very cumbersome.

So here the bugfix and also wrapped types. I also kept the original names so it's totally non-breaking.

### DIFF
--- a/src/apis/Event.bs.js
+++ b/src/apis/Event.bs.js
@@ -1,1 +1,40 @@
-/* This output is empty. Its source's type definitions, externals and/or unused code got optimized away. */
+'use strict';
+
+
+var Line = { };
+
+var TextLayout = {
+  Line: Line
+};
+
+var Layout = { };
+
+var LayoutEvent = {
+  Layout: Layout
+};
+
+var PressEvent = { };
+
+var Dimensions = { };
+
+var ContentOffset = { };
+
+var ContentInset = { };
+
+var ScrollEvent = {
+  ContentOffset: ContentOffset,
+  ContentInset: ContentInset
+};
+
+var SwitchChangePayload = { };
+
+var TargetEvent = { };
+
+exports.TextLayout = TextLayout;
+exports.LayoutEvent = LayoutEvent;
+exports.PressEvent = PressEvent;
+exports.Dimensions = Dimensions;
+exports.ScrollEvent = ScrollEvent;
+exports.SwitchChangePayload = SwitchChangePayload;
+exports.TargetEvent = TargetEvent;
+/* No side effect */
--- a/src/apis/Event.re
+++ b/src/apis/Event.re
@@ -63,38 +63,37 @@ type responderSyntheticEvent('a) = {
 };
 
 module TextLayout = {
-  module Line = {
-    type t = {
-      x: float,
-      y: float,
-      width: float,
-      height: float,
-      ascender: float, // verify
-      capHeight: float, // verify
-      descender: float, // verify
-      text: string,
-      xHeight: float // verify
-    };
+  type line = {
+    x: float,
+    y: float,
+    width: float,
+    height: float,
+    ascender: float, // verify
+    capHeight: float, // verify
+    descender: float, // verify
+    text: string,
+    xHeight: float // verify
   };
-  type t = {lines: array(Line.t)};
+  type t = {lines: array(line)};
+};
+
+module Layout = {
+  type t = {
+    x: float,
+    y: float,
+    width: float,
+    height: float,
+  };
 };
 
 module LayoutEvent = {
-  module Layout = {
-    type t = {
-      x: float,
-      y: float,
-      width: float,
-      height: float,
-    };
-  };
   type t = {layout: Layout.t};
 };
 
 type layout = LayoutEvent.t;
 type layoutEvent = syntheticEvent(layout);
 
-type textLayout = TextLayout.Line.t;
+type textLayout = TextLayout.line;
 type textLayouts = TextLayout.t;
 type textLayoutEvent = syntheticEvent(textLayouts);
 
@@ -125,32 +124,28 @@ module Dimensions = {
 type dimensions = Dimensions.t;
 
 module ScrollEvent = {
-  module ContentOffset = {
-    type t = {
-      x: float,
-      y: float,
-    };
+  type contentOffset = {
+    x: float,
+    y: float,
   };
 
-  module ContentInset = {
-    type t = {
-      bottom: float,
-      left: float,
-      right: float,
-      top: float,
-    };
+  type contentInset = {
+    bottom: float,
+    left: float,
+    right: float,
+    top: float,
   };
 
   type t = {
-    contentInset: ContentInset.t,
-    contentOffset: ContentOffset.t,
+    contentInset: contentOffset,
+    contentOffset: contentInset,
     contentSize: dimensions,
     layoutMeasurement: dimensions,
   };
 };
 
-type contentOffset = ScrollEvent.ContentOffset.t;
-type contentInset = ScrollEvent.ContentInset.t;
+type contentOffset = ScrollEvent.contentOffset;
+type contentInset = ScrollEvent.contentInset;
 type scrollEventPayload = ScrollEvent.t;
 type scrollEvent = syntheticEvent(scrollEventPayload);
 
--- a/src/apis/Event.re
+++ b/src/apis/Event.re
@@ -62,70 +62,108 @@ type responderSyntheticEvent('a) = {
   },
 };
 
-type textLayout = {
-  x: float,
-  y: float,
-  width: float,
-  height: float,
-  ascender: float, // verify
-  capHeight: float, // verify
-  descender: float, // verify
-  text: string,
-  xHeight: float // verify
+module TextLayout = {
+  module Line = {
+    type t = {
+      x: float,
+      y: float,
+      width: float,
+      height: float,
+      ascender: float, // verify
+      capHeight: float, // verify
+      descender: float, // verify
+      text: string,
+      xHeight: float // verify
+    };
+  };
+  type t = {lines: array(Line.t)};
 };
 
-type layoutEvent = syntheticEvent(layout)
-and layout = {
-  x: float,
-  y: float,
-  width: float,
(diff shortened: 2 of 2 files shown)


sgny:
I ran into this issue when updating the `CameraRoll` example. I do wonder if `Js.t` definitions weren't less pain in the end.

jfrolich:
> I ran into this issue when updating the `CameraRoll` example. I do wonder if `Js.t` definitions weren't less pain in the end.

I converted `graphql-ppx` from `Js.t` to objects, so working a lot with records now, and I can say that records are absolutely better in almost every way. You just have to help the compiler with inference in a small amount of cases by bringing the record in scope with a local open, or annotating the exact type. A big upside is that the errors are way better than with `Js.t`.

sgny:
I realise there was an oversight re: `layoutEvent` in the huge transition to records, thanks for catching that.

Perhaps inference will improve in the future, and local opens help, I hope I'll find them an overall improvement as well. I haven't had the chance to attempt a transition to records so I'm stuck with an older commit for the moment. 

I trust you got this right and in a cursory glance it looks fine. I'll take a proper look soon.

sgny:
@jfrolich @MoOx 

I regret I haven't tried records to any extent earlier. I thought having to use `##nativeEvent` before accessing records fields was a bit awkward, but 
 
```reason
let getWidth = e => LayoutEvent.(Layout.(e##nativeEvent.layout.width));
```
instead of
```reason
let getWidth = e => e##nativeEvent##layout##width;
```
is _really, really_ awkward.

Annotating helps, but for
```reason
let getWidth = (e: layoutEvent) => e##nativeEvent.layout.width;
```
the compiler returns
```
  Warning number 40 (configured as error) 
/git/reason-react-native/src/apis/Event.re 171:51-56
  
  169 │ type targetEvent = syntheticEvent(targetPayload);
  170 │ 
  171 │ let getWidth = (e: layoutEvent) => e##nativeEvent.layout.width;
  172 │ 
  173 │ 
  
  layout was selected from type LayoutEvent.t.
It is not visible in the current scope, and will not 
be selected if the type becomes unknown.
  
  Warning number 42 (configured as error) 
/git/reason-react-native/src/apis/Event.re 171:51-56
  
  169 │ type targetEvent = syntheticEvent(targetPayload);
  170 │ 
  171 │ let getWidth = (e: layoutEvent) => e##nativeEvent.layout.width;
  172 │ 
  173 │ 
  
  this use of layout relies on type-directed disambiguation,
it will not compile with OCaml 4.00 or earlier.
  
  Warning number 40 (configured as error) 
/git/reason-react-native/src/apis/Event.re 171:58-62
  
  169 │ type targetEvent = syntheticEvent(targetPayload);
  170 │ 
  171 │ let getWidth = (e: layoutEvent) => e##nativeEvent.layout.width;
  172 │ 
  173 │ 
  
  width was selected from type LayoutEvent.Layout.t.
It is not visible in the current scope, and will not 
be selected if the type becomes unknown.
  
  Warning number 42 (configured as error) 
/git/reason-react-native/src/apis/Event.re 171:58-62
  
  169 │ type targetEvent = syntheticEvent(targetPayload);
  170 │ 
  171 │ let getWidth = (e: layoutEvent) => e##nativeEvent.layout.width;
  172 │ 
  173 │ 
  
  this use of width relies on type-directed disambiguation,
it will not compile with OCaml 4.00 or earlier.
```

Obviously those warnings can be disabled and 42 is definitely irrelevant. However, should we advise that 40 should be disabled as well? I figure without the annotation, the relevant module could not be determined and the compiler should complain if the annotation were later removed. I can live with that. Can you think of any other issues due to disabling warning 40?

jfrolich:
Warning 42 can safely be ignored, and to only way to not have warning 40 is to explicitly bring the records in scope (which I agree has more awkward syntax).

jfrolich:
Btw 40 is by default ignored by Bucklescript

MoOx:
@sgny Sure that somehow awkward, but it's explicit & that's why this is my preferred version. That would need doc for sure to make this easier to find.

Can't we avoid the definition with module inside module & have a flat (whenever possible) declaration instead so we can just have `(e: Event.Layout.t)`maybe (that would make all warnings disappear?)

sgny:
@jfrolich 

It seems both are disabled by default, so I don't know why I was getting them. I had tried within `reason-react-native`, using the default `bsconfig`. I was back on 7.2.2, moving to 7.3.1 stopped that, but in between releases only warning 20 got disabled, so it cannot be that.

@MoOx 

If I'm reading you correctly, I've been thinking along the same lines. Some flattening would help. For example, instead of
```reason
module TextLayout = {
     type line = {
       x: float,
       y: float,
       width: float,
       height: float,
       ascender: float, // verify
       capHeight: float, // verify
       descender: float, // verify
       text: string,
       xHeight: float // verify
   };
   type t = {lines: array(line)};

```
```reason
module TextLayout = {
   module Line = {
     type t = {
       x: float,
       y: float,
       width: float,
       height: float,
       ascender: float, // verify
       capHeight: float, // verify
       descender: float, // verify
       text: string,
       xHeight: float // verify
     };
   };
   type t = {lines: array(Line.t)};
```
would require only one local open. 

@MoOx @jfrolich What do you think? I would prefer to type annotate (I almost always do), but improving the other use case would be nice.

sgny:
As @MoOx already commented, we could probably have some types in the top level. For example, `pressEvent` does not seem to share fields with others (as long as `syntheticEvent` is a `Js.t` type) and is a fairly common event to handle. Likewise for `contentInset`.

Another option might be to move some/all of these events to separate modules, which would save a local open of `Event`.

jfrolich:
> @sgny Sure that somehow awkward, but it's explicit & that's why this is my preferred version. That would need doc for sure to make this easier to find.
> 
> Can't we avoid the definition with module inside module & have a flat (whenever possible) declaration instead so we can just have `(e: Event.Layout.t)`maybe (that would make all warnings disappear?)

Yes I think we can flatten a bit more! Good point.

jfrolich:
How about this change?

cknitt:
Sorry, I am late to the "records" party, but if we are going all records now, shouldn't we make `syntheticEvent` a record, too? So that we can have

```e.nativeEvent.layout.width```

instead of

```e##nativeEvent.layout.width```

?

What is `[@bs.meth]` now could then be handled using `[@bs.send]`.

I know this is not 100% correct as an event record that you construct yourself won't have these methods. But it would improve overall consistency and ease of use.

What do you think?

**Comment by sgny** [comment]
`[@bs.meth]` annotation makes sure that the method has to be applied right away as the event is returned to the callback, with no currying. This makes sure that methods defined with `this` are properly handled. `[@bs.send]` does not really provide the same guarantee, as the result of the getter may be assigned to some function and called later.

Users may achieve the same result with `e.preventDefault()` and `e->preventDefault()` (where `preventDefault` is an external defined with `@bs.send`) as `e##preventDefault()`.

However, `e##preventDefault` is a dead end that the other two cannot replicate.

**Comment by cknitt** [comment]
@sgny `e.preventDefault()` and `e->preventDefault` surely do **not** yield the same result:

```reason
module Event = {
  type t = {
    someField: int,
    // This is just here for testing, it wouldn't be present as a record field
    // in the "real" bindings.
    preventDefault: unit => unit,
  };

  [@bs.send] external preventDefault: t => unit = "preventDefault";
};

let f = (evt: Event.t) => {
  evt.preventDefault();
  evt->Event.preventDefault;
};
```

compiles to

```js
function f(evt) {
  Curry._1(evt.preventDefault, undefined);
  evt.preventDefault(); 
}
```

Also, I don't know what you mean by "result of the getter". There is no getter, `bs.send` just invokes a method on the object. This even still works correctly if you define

```reason
let preventDefault2 = preventDefault;
```

in the Event module and do `evt->Event.preventDefault2`.

MoOx:
Speaking of Module & types, for this PR, I don't really understand the benefit of having some modules with types in them. Can't we just have flat types, even with longer names? For exemple for Dimensions, I don't see the value of having a module with a type t & later use "dimensions = Dimensions.t" (and same for similar places)

jfrolich:
> Speaking of Module & types, for this PR, I don't really understand the benefit of having some modules with types in them. Can't we just have flat types, even with longer names? For exemple for Dimensions, I don't see the value of having a module with a type t & later use "dimensions = Dimensions.t" (and same for similar places)

If you have one module with records that have overlapping fields, type inference doesn't work anymore so the compiler cannot distinguish between the two types due to shadowing. Wrapping it in a module fixes this. You can always constrain to an exact type, but in some cases that is not really ergonomic. (so for instance if you use `Layout.t` in a hook you can just `open Layout` and have the type inference work.

sgny:
The same end result meaning, as long as the function is applied in full at the same time, curried or not, it should be fine.

It appears I misspecified the method with another `unit` as below, but the compiler already takes care of that
```reason
[@bs.send] external preventDefault: (t, unit) => unit = "preventDefault";
```
then `let f = evt->Event.preventDefault;` becomes an issue. Without that unit, everything's fine.

jfrolich:
I also like the `@bs.send` a bit more API wise, as it's more idiomatic Reason.

MoOx:
I am ok with this change, we can handle preventDefault in another PR. 
@sgny @cknitt is this good for you in its current state?
(ps: I am trying new bindings by upgrading some code & it seems to be the missing piece for me ^^)

sgny:
`e->preventDefault`, etc. are pretty similar to what's available with `reason-react`, so we may as well implement that while we're breaking things. 

`reason-react`'s `MakeEventWithType` is more consistent with using `@bs.get/@bs.send` methods to access keys, but I think our approach in keeping the record for all non-method keys is more straightforward. I suppose the slight inconsistency is quite alright

How do we handle the `syntheticEvent` and `responderSyntheticEvent` when it comes to the methods however? Should we namespace one or the other or functorise all these event types instead?

MoOx:
@sgny Merging, lets discuss the rest in #688

PR #682, opened 2020-04-22
Add binding for Platform.Version

tienle:
Detects the version of the platform in which the app is running.

https://reactnative.dev/docs/platform-specific-code#detecting-the-android-version

### DIFF
--- a/src/apis/Event.re
+++ b/src/apis/Event.re
@@ -74,13 +74,16 @@ type textLayout = {
   xHeight: float // verify
 };
 
-type layoutEvent = syntheticEvent(layout)
+type layoutEvent = syntheticEvent(layoutPayload)
 and layout = {
   x: float,
   y: float,
   width: float,
   height: float,
-};
+}
+and layoutPayload = {
+  layout: layout
+} ;
 
 type textLayoutEvent = syntheticEvent(textLayouts)
 and textLayouts = {lines: array(textLayout)};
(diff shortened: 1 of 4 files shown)


cknitt:
According to the docs you linked, `Platform.Version` is a string for iOS, but an int for Android. This should be reflected correctly in the bindings IMHO.

One way to do this might be to have both

* `Platform.iosVersion` (string) and
* `Platform.androidVersionCode` (int)

?

cknitt:
LGTM - what do you think, @MoOx?

MoOx:
For better discoverability (autocomplete), I would call them `.versionIos` & `.versionAndroid`

cknitt:
@MoOx Another idea:

Looking at the TypeScript definitions, I noticed that there is some other OS-specific information (isPad, isTVOS).

So maybe we should have Platform.IOS.version (string), Platform.IOS.isPad, Platform.IOS.isTVOS and Platform.Android.version (int)?

MoOx:
@cknitt I would prefer to be close to original API without adding any complexity for new comers.
@sgny anything to add to help us with this decision ? 😁

sgny:
`Platform.constants` returns quite a bit of information, including some which may be useful, such as the RN version, fingerprint for Android, etc. Set of available keys seems rather different between iOS and Android. We could support access to that directly which should handle other functionality as well. For example, `isPad` and `isTV` (`isTVOS` is deprecated) are simply whether `interfaceIdiom` key is `"tv"` or `"pad"`.

MoOx:
Ok so you recommend that we bind Platform.constants instead of Platform.versionX ?

MoOx:
@sgny up for my question :)

MoOx:
@tuanmai up :)

PR #684, opened 2020-04-26
Add support for callback refs.

cknitt:
@MoOx @jfrolich This implements support of callback refs as discussed on Discord. It is a breaking change.

You would now use

```... ref={myRef->Ref.value} ...```

for "value refs" (like we have currently) and

```... ref={setRef->Ref.callback} ...```

for "callback refs".

### DIFF
--- a/src/elements/DrawerLayoutAndroidElement.re
+++ b/src/elements/DrawerLayoutAndroidElement.re
@@ -1,5 +1,5 @@
 type element;
-type ref = React.Ref.t(Js.nullable(element));
+type ref = Ref.t(element);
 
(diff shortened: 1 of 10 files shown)


MoOx:
What about `Ref.object` & `Ref.callback`? "object" is the term used on React doc https://reactjs.org/docs/glossary.html#refs

MoOx:
Oh I just saw the update with `Ref.value`. Maybe that make more sense this way.

sgny:
@MoOx

I think this is fine. So far, I have simply type cast callback refs to simple refs, but as that would not prevent calling `React.Ref.current` on callback refs, some nuisance is necessary.

PR #685, opened 2020-04-27
useCallback to memoize callback ref

sgny:
As long as we need a function to type cast refs, we may enforce the `useCallback` pattern while creating callback refs.

### DIFF
--- a/src/elements/Ref.re
+++ b/src/elements/Ref.re
@@ -4,4 +4,12 @@ type valueRef('element) = React.Ref.t(Js.nullable('element));
 type callbackRef('element) = Js.nullable('element) => unit;
 
 external value: valueRef('element) => t('element) = "%identity";
-external callback: callbackRef('element) => t('element) = "%identity";
+
+[@bs.module "react"]
+external callback:
+  (
+    [@bs.uncurry] (Js.nullable('element) => unit),
+    [@bs.as {json|[]|json}] _
+  ) =>
+  t('element) =
+  "useCallback";


cknitt:
I get the idea, but I am not really in favor of it. As a user of the bindings, I would expect them to behave exactly the same  way as their React Native counterparts and not to provide additional unexpected magic. Any additional convenience stuff should go into a higher layer or a separate utilities module or something like that IMHO.

For me, the purpose of the ReactNative.Ref.value and ReactNative.Ref.callback functions is just type conversion in order to be able to pass different types of refs and nothing more.

What do you think @MoOx?

MoOx:
Yeah sorry but not a fan too.

sgny:
Any interest in an additional `useCallback` with the same definition, then?

Not having this in the repo doesn't prevent its use anyway, so no problem.

MoOx:
@sgny this make more sense in reason-react codebase isn't it?

sgny:
I already have a PR there :grin:

Since we have our own `Ref`, so any method to incorporate its creation needs to be here, if at all. As I said, I don't particularly mind if this doesn't go in as I can still use this.

MoOx:
Ok I get it, maybe then we could have an explicit `useCallbackRef` or something like that?

jfrolich:
The current code is the same code as `ReactDOMRe`, so also comparable to how reason-react works with DOM.

sgny:
@MoOx I believe we can close this. There were comments arguing this may be suboptimal in certain cases in response to my similar PR in `reason-react`. Reading the implementation of `useCallback` I don't see the performance impact, but we can close this.

PR #687, opened 2020-05-01
[Style] use JS type conversion to simplify pct, deg and rad

sgny:
Sadly the below is not supported by the compiler, at least for the moment.
```reason
external pct: (float, [@bs.as "%"] _) => string = "#string_append";
```
We can, however, still simplify `pct`, `deg` and `rad` by removing the explicit conversion to string.

Admittedly, the performance difference is not really benchmarkable; this is simply because it's possible :grin:.

### DIFF
--- a/src/apis/Style.bs.js
+++ b/src/apis/Style.bs.js
@@ -2,15 +2,15 @@
 
 
 function pct(num) {
-  return num.toString() + "%";
+  return num + "%";
 }
 
 function deg(num) {
-  return num.toString() + "deg";
+  return num + "deg";
 }
 
 function rad(num) {
-  return num.toString() + "rad";
+  return num + "rad";
 }
 
 exports.pct = pct;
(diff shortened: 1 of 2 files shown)


MoOx:
What does give `100.->pct` output? If that's `100.%` that can be a problem on some platforms (eg: breaks on the web in most browser, same for `10.px`).

sgny:
I hadn't tried that :grin:
Now that I have, you'd get "100%". Float values such as `100.` get converted to `100`, so the function would be passed that.

MoOx:
Good. What about having a more explicit name (just for maintenance) & name the fonction "appendFloatWithString"?

MoOx:
@sgny up :)

sgny:
I'm hesitating as `#string_append` is not strictly legal for external use according to this [comment](https://github.com/BuckleScript/bucklescript/issues/4336#issuecomment-623806989). Perhaps it's better to hold off until there's clarification.

I've been running into this issue of decorators being unavailable for use with these primitives again and again, but it seems such use will be allowed in the future.

sgny:
@MoOx 
Sorry I was swamped under a work project for the past few weeks. Now that I submitted my bit yesterday,

I think we can close this for now. The benefit is not worth the risk of having the rug pulled from under us later.

PR #690, opened 2020-05-07
[Event] convert syntheticEvent to record

### LINKED ISSUE #688: Revisit some syntheticEvent/responderSyntheticEvent (eg: e->preventDefault)
Following https://github.com/reason-react-native/reason-react-native/pull/681

@sgny said

> `e->preventDefault`, etc. are what's available with `reason-react`, so we may as well implement that while we're breaking things. 
>
> `reason-react`'s `MakeEventWithType` is more consistent with using `@bs.get/@bs.send` methods to access keys, but I think our approach in keeping the record for all non-method keys is more straightforward. I suppose the slight inconsistency is quite alright.
>
> How do we handle the `syntheticEvent` and `responderSyntheticEvent` difference when it comes to the methods however? Should we namespace one or the other or functorise all these event types instead?

sgny:
closes #688

I added another type parameter to `syntheticEvent` to handle the additional `touchHistory` key in `responderSyntheticEvent`.

Discussion points to settle:

• `touchHistory` may be included in the record definition, then what previously used to be `syntheticEvent('a)` becomes `syntheticEvent('a, unit)` and `responderSyntheticEvent('a)` becomes `syntheticEvent('a, touchHistory)`. My testing suggests `unit` is fine as type parameter `'b`, but given such events will have `undefined` as the value for `e.touchHistory`, `option(unit)` may be even safer as `e.touchHistory == None` would hold. 

• If `touchHistory` is not included in the definition, `'b` becomes a phantom type to ensure that the `touchHistory` method only can be applied on `syntheticEvent('a, touchHistory)`. As expected, `e.touchHistory` is not available.

• Should `type responderSyntheticEvent('a) = syntheticEvent('a, touchHistory);` still be included and `pressEvent` remain defined to be of type `responderSyntheticEvent('a)`? Not needed, obviously, but would keep closer to the JS definitions.

• `[@bs.send]` to handle keys previously decorated with `[@bs.meth]` as discussed before.

• `[@bs.get]` to define getters for all non-method fields to achieve consistency with `reason-react`. While the record has `Js.Nullable.t` types for some keys, IMO, using `[@bs.return nullable]` to return `option` types is a better API.

An alternative approach may be to define parametrized modules for `SyntheticEvent` and `ResponderSyntheticEvent` and then define each event as an instance of either. I think this approach is straightforward, but based on feedback I can work on that instead.

### DIFF
--- a/src/apis/Event.re
+++ b/src/apis/Event.re
@@ -1,66 +1,79 @@
 // see https://github.com/facebook/react-native/blob/master/Libraries/Types/CoreEventTypes.js
 
-type syntheticEvent('a) = {
-  .
-  "bubbles": Js.Nullable.t(bool),
-  "cancelable": Js.Nullable.t(bool),
-  "currentTarget": float,
-  "defaultPrevented": Js.Nullable.t(bool),
-  "dispatchConfig": {. "registrationName": string},
-  "eventPhase": Js.Nullable.t(float),
-  [@bs.meth] "preventDefault": unit => unit,
-  [@bs.meth] "isDefaultPrevented": unit => bool,
-  [@bs.meth] "stopPropagation": unit => unit,
-  [@bs.meth] "isPropagationStopped": unit => bool,
-  "isTrusted": Js.Nullable.t(bool),
-  "nativeEvent": 'a,
-  [@bs.meth] "persist": unit => unit,
-  "target": Js.Nullable.t(float),
-  "timeStamp": float,
-  "_type": Js.Nullable.t(string),
+type syntheticEvent('a, 'b) = {
+  bubbles: Js.Nullable.t(bool),
+  cancelable: Js.Nullable.t(bool),
+  currentTarget: float,
+  defaultPrevented: Js.Nullable.t(bool),
+  dispatchConfig: registrationName,
+  eventPhase: Js.Nullable.t(float),
+  isTrusted: Js.Nullable.t(bool),
+  nativeEvent: 'a,
+  target: Js.Nullable.t(float),
+  timeStamp: float,
+  [@bs.as "type"]
+  _type: Js.Nullable.t(string),
+  touchHistory: 'b // this field can be omitted, making 'b a phantom type
+}
+and registrationName = {registrationName: string};
+
+// type responderSyntheticEvent('a) defined as type syntheticEvent('a, touchHistory)
+
+type touchHistory = {
+  indexOfSingleActiveTouch: float,
+  mostRecentTimeStamp: float,
+  numberActiveTouches: float,
+  touchBank: array(touchBank),
+}
+and touchBank = {
+  touchActive: bool,
+  startPageX: float,
+  startPageY: float,
+  startTimeStamp: float,
+  currentPageX: float,
+  currentPageY: float,
+  currentTimeStamp: float,
+  previousPageX: float,
+  previousPageY: float,
+  previousTimeStamp: float,
 };
 
-type responderSyntheticEvent('a) = {
-  .
-  // synthethicEvent keys
-  "bubbles": Js.Nullable.t(bool),
-  "cancelable": Js.Nullable.t(bool),
-  "currentTarget": float,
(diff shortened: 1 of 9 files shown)


src/apis/Event.re
```diff
@@ -1,66 +1,79 @@
...
-        "currentPageY": float,
-        "currentTimeStamp": float,
-        "previousPageX": float,
-        "previousPageY": float,
-        "previousTimeStamp": float,
-      }),
-  },
-};
+[@bs.get] [@bs.return nullable]
+external bubbles: syntheticEvent('a, 'b) => option(bool) = "bubbles";
```
MoOx, line 42: If I understand correctly, the benefit of this methods instead of directly hitting the record field is to have a option instead of a js nullable? I am assuming this is still zero cost (external), so how is the compiler making diff between Js.Nullable.t & option? Can't we directly type the record fields as option() & allow to hit record fields for getter & only have externals for send method? Maybe that's not optimal to have even.field & event->Event.method... ?

sgny: So far, we have not verified if the fields may have `null` values and we have taken the safer but more annoying-to-use path to wrap everything in `Js.Nullable.t`.

`[@bs.return nullable]` provides a safe way to handle the case when the return value may be a `null`. Assume `someExternal` returns `option('a)` with the decorator and `Js.Nullable.t('a)` without.  A `switch` statement such as

```
switch (someExternal(x)) {
| None => Js.log("a")
| Some(_) => Js.log("b")
}
```
will have the below JS output when `someExternal` is defined **with** the decorator:
```
if (match == null) {
  console.log("a");
} else {
  console.log("b");
}
```
compared to the output when defined **without** the decorator
```
if (match !== undefined) {
  console.log("b");
} else {
  console.log("a");
}
```

Basically, the compiler makes use of the deep inequality operator when it needs to differentiate between `null` and `undefined` and shallow equality when it doesn't.

We have already made use of this with some of these bindings. There is no overhead for pattern matching (as in the switch statement), but `someExternal(x) == Some(y);` adds
```
var Caml_obj = require("./stdlib/caml_obj.js");
var Caml_option = require("./stdlib/caml_option.js");
```
to the JS output.

sgny:
> If I get this right, the idea of `syntheticEvent('a, 'b)` is to have the same code reused for responder events too without having to duplicate it?

Since we cannot rely on record fields for `preventDefault`, etc., we'd need to define a method for each type. That would either lead to contrived naming to avoid name clashes, or submodules. To avoid having to reference `Event. SyntheticEvent ` or `Event.ResponderSyntheticEvent` when using those methods, we could define the modules `SyntheticEvent` and `ResponderSyntheticEvent` and include them in `LayoutEvent` or 'PressEvent` or define `SyntheticEvent` and `ResponderSyntheticEvent` as parametrized modules to define `LayoutEvent, `PressEvent` etc. based on those.

TBH, I don't have a preference between this and such approaches.
 
> From a user perspective I feel this can be ok (if I understand correctly, everything goes with event->Event.(getter|setter) instead of event##getter|setter?

For the most part; except for methods like `preventDefault`. We can define keys with `unit => unit` or `unit => bool` values, but it's probably better to make sure users have to apply the method immediately.

MoOx:
My only concern is the simplicity for end-user (as well as the safety). I am ok with whatever implementation as soon as the usage is not too verbose or weird.
I have the feeling that you understand way better than I do the consequences of the solutions you have in mind so I can just agree with you as soon the usage is simple enough :D 

Maybe @cknitt have more constructive feedback than I do as I feel lost and confused as soon as choices involved are too technical 😂

sgny:
@MoOx It's more likely I've done a poor job explaining. I'll prepare an example of the alternative, if we decide upon that path, I'll create the full PR.

src/apis/Event.re
cknitt: I also would prefer to just handle actual methods using bs.send and keep everything else as record fields, even if it means that these fields may have type `Js.Nullable.t('something)`.  As a user of the API, I would find it confusing if some fields were available directly in the record and others through a getter function.

PR #694, opened 2020-05-08
[Event] convert syntheticEvent and responderSyntheticEvent to record

### LINKED ISSUE #688: Revisit some syntheticEvent/responderSyntheticEvent (eg: e->preventDefault)

sgny:
closes #688

This is one alternate specification mentioned in #689. A module is defined for each event type, `include SyntheticEvent` allows the type and each method to exist in each module so that one local open is enough to access all fields.

For some reason [a similar approach](https://github.com/reason-react-native/reason-react-native/blob/event_functors/src/apis/Event.re) did not work out, as I kept getting errors while tying to access fields. This version is slightly simplified, but falls short with respect to DRY compared to that.


src/apis/Event.re
```diff
@@ -1,106 +1,143 @@
...
-        "currentPageX": float,
-        "currentPageY": float,
-        "currentTimeStamp": float,
-        "previousPageX": float,
-        "previousPageY": float,
-        "previousTimeStamp": float,
-      }),
-  },
+module SyntheticEvent = {
+  type t_('a) = {
```
jfrolich, line 38: Why `t_` and not `t`? Can the typename be more descriptive if it cannot be `t`?

src/apis/Event.re
```diff
@@ -1,106 +1,143 @@
...
+    dispatchConfig: registrationName,
+    eventPhase: Js.Nullable.t(float),
+    isTrusted: Js.Nullable.t(bool),
+    nativeEvent: 'a,
+    target: Js.Nullable.t(float),
+    timeStamp: float,
+    [@bs.as "type"]
+    _type: Js.Nullable.t(string),
+  }
+  and registrationName = {registrationName: string};
```
jfrolich, line 52: Doesn't need dependent definition (`and`)

src/apis/Event.re
```diff
@@ -1,106 +1,143 @@
...
+    touchHistory,
+  }
+  and registrationName = {registrationName: string}
+  and touchHistory = {
+    indexOfSingleActiveTouch: float,
+    mostRecentTimeStamp: float,
+    numberActiveTouches: float,
+    touchBank: array(touchBank),
+  }
+  and touchBank = {
```
jfrolich, line 83: No dependent definition necessary.

src/apis/Event.re
sgny: When `SyntheticEvent` is included in some event module, `type t` would also be included and the compiler won't allow types to be redefined. I couldn't come up with a better name, so suggestions are welcome.

Since `SyntheticEvent` and `ResponderSyntheticEvent` need to be used to define new events, we can add two modules (such as `_SyntheticEvent` and `_ResponderSyntheticEvent`)  for all modules to inherit from. Then in `SyntheticEvent` and `ResponderSyntheticEvent` we could have `type t ('a) = t_('a)`.

src/apis/Event.re
sgny: I made the choice to have `t` first, but as you said it's not needed; they are not recursively defined.

src/apis/Event.re
sgny: Same here, organisational choice. I don't suppose there's a significant performance impact. I expect there can't be further impact once `reason-react-native` is compiled.

src/apis/Event.re
sgny: That would allow users to define new events either by inhering from `_SyntheticEvent` or using `SyntheticEvent.t('payload)` without a new module.

src/apis/Event.re
jfrolich: Actually gentype cannot handle this. Not sure if that’s relevant but I also believe it’s better to keep types as simple as possible so it’s easier for the compiler.

cknitt: I agree with @jfrolich, and I think it would also make it easier for the reader, not just the compiler. When I read "let...and", at first glance I automatically assume that this is a recursive definition. Then I start scanning for where the recursion actually is and find none.

PR #695, opened 2020-05-10
Events as records, approach 3.

cknitt:
I took @sgny's #694 and rewrote it to get rid of type parameters and to be more similar to the pattern I used for native elements and their methods.

For the moment, I still kept the accessor functions for nullable fields, but I am thinking of removing them.

This compiles, but is otherwise untested.

### DIFF
--- a/src/apis/Animated.bs.js
+++ b/src/apis/Animated.bs.js
@@ -108,6 +108,9 @@ var make$1 = ReactNative.Animated.createAnimatedComponent(ReactNative.Image);
 var $$Image = {
   Source: Image$ReactNative.Source,
   DefaultSource: Image$ReactNative.DefaultSource,
+  ImageLoadEvent: Image$ReactNative.ImageLoadEvent,
+  $$ErrorEvent: Image$ReactNative.$$ErrorEvent,
+  $$ProgressEvent: Image$ReactNative.$$ProgressEvent,
   make: make$1
 };
 
--- a/src/apis/Event.bs.js
+++ b/src/apis/Event.bs.js
@@ -1,28 +1,37 @@
 'use strict';
 
 
-var TextLayout = { };
+function EventMethods(T) {
+  return { };
+}
 
-var Layout = { };
+function SyntheticEvent(T) {
+  return { };
+}
+
+function ResponderSyntheticEvent(T) {
+  return { };
+}
 
 var LayoutEvent = { };
 
-var PressEvent = { };
+var TextLayoutEvent = { };
 
-var Dimensions = { };
+var PressEvent = { };
 
 var ScrollEvent = { };
 
-var SwitchChangePayload = { };
+var SwitchChangeEvent = { };
 
 var TargetEvent = { };
 
-exports.TextLayout = TextLayout;
-exports.Layout = Layout;
+exports.EventMethods = EventMethods;
+exports.SyntheticEvent = SyntheticEvent;
+exports.ResponderSyntheticEvent = ResponderSyntheticEvent;
 exports.LayoutEvent = LayoutEvent;
+exports.TextLayoutEvent = TextLayoutEvent;
 exports.PressEvent = PressEvent;
-exports.Dimensions = Dimensions;
 exports.ScrollEvent = ScrollEvent;
-exports.SwitchChangePayload = SwitchChangePayload;
+exports.SwitchChangeEvent = SwitchChangeEvent;
 exports.TargetEvent = TargetEvent;
 /* No side effect */
--- a/src/apis/Event.re
+++ b/src/apis/Event.re
@@ -1,68 +1,117 @@
 // see https://github.com/facebook/react-native/blob/master/Libraries/Types/CoreEventTypes.js
 
-type syntheticEvent('a) = {
-  .
-  "bubbles": Js.Nullable.t(bool),
-  "cancelable": Js.Nullable.t(bool),
-  "currentTarget": float,
-  "defaultPrevented": Js.Nullable.t(bool),
-  "dispatchConfig": {. "registrationName": string},
-  "eventPhase": Js.Nullable.t(float),
-  [@bs.meth] "preventDefault": unit => unit,
-  [@bs.meth] "isDefaultPrevented": unit => bool,
-  [@bs.meth] "stopPropagation": unit => unit,
-  [@bs.meth] "isPropagationStopped": unit => bool,
-  "isTrusted": Js.Nullable.t(bool),
-  "nativeEvent": 'a,
-  [@bs.meth] "persist": unit => unit,
-  "target": Js.Nullable.t(float),
-  "timeStamp": float,
-  "_type": Js.Nullable.t(string),
+type registrationName = {registrationName: string};
+
+module EventMethods = (T: {type event;}) => {
+  [@bs.get] [@bs.return nullable]
+  external bubbles: T.event => option(bool) = "bubbles";
+  [@bs.get] [@bs.return nullable]
+  external cancelable: T.event => option(bool) = "cancelable";
+  [@bs.get] external currentTarget: T.event => float = "currentTarget";
+  [@bs.get] [@bs.return nullable]
+  external defaultPrevented: T.event => option(bool) = "defaultPrevented";
+  [@bs.get]
+  external dispatchConfig: T.event => registrationName = "dispatchConfig";
+  [@bs.get] [@bs.return nullable]
+  external eventPhase: T.event => option(float) = "eventPhase";
+  [@bs.send] external preventDefault: T.event => unit = "preventDefault";
+  [@bs.send]
+  external isDefaultPrevented: T.event => bool = "isDefaultPrevented";
+  [@bs.send] external stopPropagation: T.event => unit = "stopPropagation";
+  [@bs.send]
+  external isPropagationStopped: T.event => bool = "isPropagationStopped";
+  [@bs.get] [@bs.return nullable]
+  external isTrusted: T.event => option(bool) = "isTrusted";
+  [@bs.get] external nativeEvent: T.event => 'a = "nativeEvent";
+  [@bs.send] external persist: T.event => unit = "persist";
+  [@bs.get] [@bs.return nullable]
+  external target: T.event => option(float) = "target";
+  [@bs.get] external timeStamp: T.event => float = "timeStamp";
+  [@bs.get] [@bs.return nullable]
+  external _type: T.event => option(string) = "type";
 };
 
-type responderSyntheticEvent('a) = {
-  .
-  // synthethicEvent keys
-  "bubbles": Js.Nullable.t(bool),
-  "cancelable": Js.Nullable.t(bool),
-  "currentTarget": float,
-  "defaultPrevented": Js.Nullable.t(bool),
-  "dispatchConfig": {. "registrationName": string},
-  "eventPhase": Js.Nullable.t(float),
-  [@bs.meth] "preventDefault": unit => unit,
-  [@bs.meth] "isDefaultPrevented": unit => bool,
-  [@bs.meth] "stopPropagation": unit => unit,
-  [@bs.meth] "isPropagationStopped": unit => bool,
-  "isTrusted": Js.Nullable.t(bool),
-  "nativeEvent": 'a,
-  [@bs.meth] "persist": unit => unit,
-  "target": Js.Nullable.t(float),
-  "timeStamp": float,
-  "_type": Js.Nullable.t(string),
-  // responderSyntheticEvent additional key
-  "touchHistory": {
-    .
-    "indexOfSingleActiveTouch": float,
-    "mostRecentTimeStamp": float,
-    "numberActiveTouches": float,
-    "touchBank":
-      array({
-        .
-        "touchActive": bool,
-        "startPageX": float,
-        "startPageY": float,
-        "startTimeStamp": float,
-        "currentPageX": float,
-        "currentPageY": float,
-        "currentTimeStamp": float,
-        "previousPageX": float,
-        "previousPageY": float,
-        "previousTimeStamp": float,
-      }),
-  },
+module SyntheticEvent = (T: {type _payload;}) => {
+  type t = {
+    bubbles: Js.Nullable.t(bool),
+    cancelable: Js.Nullable.t(bool),
+    currentTarget: float,
+    defaultPrevented: Js.Nullable.t(bool),
+    dispatchConfig: registrationName,
+    eventPhase: Js.Nullable.t(float),
+    isTrusted: Js.Nullable.t(bool),
+    nativeEvent: T._payload,
+    target: Js.Nullable.t(float),
+    timeStamp: float,
+    [@bs.as "type"]
+    _type: Js.Nullable.t(string),
+  };
+
+  include EventMethods({
+    type event = t;
+  });
 };
 
-module TextLayout = {
+module ResponderSyntheticEvent = (T: {type _payload;}) => {
+  type touchBank = {
+    touchActive: bool,
+    startPageX: float,
+    startPageY: float,
+    startTimeStamp: float,
+    currentPageX: float,
+    currentPageY: float,
+    currentTimeStamp: float,
+    previousPageX: float,
+    previousPageY: float,
+    previousTimeStamp: float,
+  };
+
+  type touchHistory = {
+    indexOfSingleActiveTouch: float,
+    mostRecentTimeStamp: float,
+    numberActiveTouches: float,
+    touchBank: array(touchBank),
+  };
+
+  type t = {
+    bubbles: Js.Nullable.t(bool),
+    cancelable: Js.Nullable.t(bool),
+    currentTarget: float,
+    defaultPrevented: Js.Nullable.t(bool),
+    dispatchConfig: registrationName,
+    eventPhase: Js.Nullable.t(float),
+    isTrusted: Js.Nullable.t(bool),
+    nativeEvent: T._payload,
+    target: Js.Nullable.t(float),
+    timeStamp: float,
+    [@bs.as "type"]
+    _type: Js.Nullable.t(string),
+    touchHistory,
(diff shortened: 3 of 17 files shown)


sgny:
I had already [tried](https://github.com/reason-react-native/reason-react-native/blob/event_functors/src/apis/Event.re) a similar approach except for defining

```reason
module ErrorEvent =
  Event.SyntheticEvent({
    type payload = {error: string};
  });
```
instead of
```reason
module ErrorEvent = {
  type payload = {error: string};

  include Event.SyntheticEvent({
    type _payload = payload;
  });
};
```
and kept getting errors trying to access fields, both with local opens and type annotations. However, as I kept getting errors even with fields such as `nativeEvent`, I didn't suspect inline definition of `payload`.

Anyway, I've tried your version and it works. Further, it can be broken with inline definition.  I switched to the simplified version and as that seems to work even with inline definition, perhaps my definition is an edge case that is slightly buggy.

sgny:
Checked just now,
```reason
module ErrorPayload = {
  type payload = {error: string};
};
module ErrorEvent =
  Event.SyntheticEvent(ErrorPayload);
```
works, so the issue I had really is that the module was defined inline.

sgny:
This doesn't seem to be a BuckleScript issue. I tested on OCaml and when `module Layout` is not defined, the signature of `LayoutEvent` has `t` as an opaque type.

However, that's not the case when the module passed as argument to `SyntheticEvent` is `{type t = string;}`.

jfrolich:
> I had already [tried](https://github.com/reason-react-native/reason-react-native/blob/event_functors/src/apis/Event.re) a similar approach except for defining
> 
> ```reason
> module ErrorEvent =
>   Event.SyntheticEvent({
>     type payload = {error: string};
>   });
> ```
> 
> instead of
> 
> ```reason
> module ErrorEvent = {
>   type payload = {error: string};
> 
>   include Event.SyntheticEvent({
>     type _payload = payload;
>   });
> };
> ```
> 
> and kept getting errors trying to access fields, both with local opens and type annotations. However, as I kept getting errors even with fields such as `nativeEvent`, I didn't suspect inline definition of `payload`.
> 
> Anyway, I've tried your version and it works. Further, it can be broken with inline definition so there might be a bug with the compiler for this edge case, either it should work or it should lead to an error referring to the inline definition.

I just read it, did you try the following:

> ```reason
> module ErrorEvent = {
>   type payload = {error: string};
> 
>   include Event.SyntheticEvent({
>     type payload;
>   } with type payload := payload;;
> };
> ```

This is the syntax to redefine types in an include.

sgny:
> > I had already [tried](https://github.com/reason-react-native/reason-react-native/blob/event_functors/src/apis/Event.re) a similar approach except for defining
> > ```reason
> > module ErrorEvent =
> >   Event.SyntheticEvent({
> >     type payload = {error: string};
> >   });
> > ```
> > 
> > 
> > instead of
> > ```reason
> > module ErrorEvent = {
> >   type payload = {error: string};
> > 
> >   include Event.SyntheticEvent({
> >     type _payload = payload;
> >   });
> > };
> > ```
> > 
> > 
> > and kept getting errors trying to access fields, both with local opens and type annotations. However, as I kept getting errors even with fields such as `nativeEvent`, I didn't suspect inline definition of `payload`.
> > Anyway, I've tried your version and it works. Further, it can be broken with inline definition so there might be a bug with the compiler for this edge case, either it should work or it should lead to an error referring to the inline definition.
> 
> I just read it, did you try the following:
> 
> > ```reason
> > module ErrorEvent = {
> >   type payload = {error: string};
> > 
> >   include Event.SyntheticEvent({
> >     type payload;
> >   }) with type payload := payload;
> > };
> > ```
> 
> This is the syntax to redefine types in an include.
> 
> Possibly we can even remove the need for it to be a functor (why is it a functor?)
> 
> This would possibly enable the following:
> 
> > ```reason
> > module ErrorEvent = {
> >   type payload = {error: string};
> > 
> >   include Event.SyntheticEvent with type payload := payload;
> > };
> > ```
> 
> Much simpler 😄

That syntax doesn't work for me on 7.3.2 or 8.0.0. Something like 
```reason
module SyntheticEvent = {
	type payload;

  type t = {
    bubbles: Js.Nullable.t(bool),
    cancelable: Js.Nullable.t(bool),
    currentTarget: float,
    defaultPrevented: Js.Nullable.t(bool),
    dispatchConfig: registrationName,
    eventPhase: Js.Nullable.t(float),
    isTrusted: Js.Nullable.t(bool),
    target: Js.Nullable.t(float),
    timeStamp: float,
    [@bs.as "type"]
    _type: Js.Nullable.t(string),
  };

  include EventMethods({
    type event = t;
  });
};

module ErrorEvent = {
	type payload = {error: string};

	include (SyntheticEvent:  (module type of SyntheticEvent) with type payload := payload);
};
```
works only when `t` does not depend on `payload` as above. Once the `nativeEvent` key is added, it appears we have to override `t` as well.

Sorry for dropping off the conversation, I've been swamped with a work project in the past weeks, but now I'm free(r) now that I've submitted my part yesterday.

jfrolich:
Sorry you probably need to do this in the module type definition instead of the module implementation.

jfrolich:
```reason
module type ErrorEvent {
  type payload = {error: string};

  include Event.SyntheticEvent with type payload := payload;
};
```

sgny:
> ```reason
> module type ErrorEvent {
>   type payload = {error: string};
> 
>   include Event.SyntheticEvent with type payload := payload;
> };
> ```

I've tried similar ideas the other day, but I couldn't exactly figure out a solution. I believe we would need to either define a module of that type or a functor that outputs that type to actually make use of it. Further, it appears any of the additional definitions can't reference that signature so we would have to repeat definitions. Have you been able to make this work?

jfrolich:
How about this:

```reason

module type SyntheticEvent_intf = {
  type payload;
  type otherType = int;
};

module SyntheticEvent: SyntheticEvent_intf = {
  type payload;
  type otherType = int;
};

module ErrorEvent = {
  type payload = {error: string};

  include (SyntheticEvent: { 
    include SyntheticEvent_intf with type payload := payload 
  });
  let payload = {error: "error"}; 
};

Js.log(ErrorEvent.payload.error);
```


See the following link for an example: https://reasonml.github.io/en/try?rrjsx=true&reason=LYewJgrgNgpgBAFwJ4Ad4GUkDsEAsYICWAxgKIBuMOA+oTgGZwC8cA3gFByKrwoCGSKCD5gA3J25o4IPDABOAFR7M4dBOIC+49qEiw4mHPiJlKOAFwHsskxSoJaDFRy7Ip-QcLES38GfkVlFjVNbV1oeFI5ORA5OxxnH2UPIRFneRi5SwBnBDk6AHMtdgk6YigIMHgACkMbEniES1Y4CS4yiqqrIwIGswc1RgB3QjxJXgFUsDhzFhSvVq4NAEpxLlgEOHm0llYM2MsAIn25Q61W4vYAKWyAOiEC6qjMxtvtsFuT1aA

sgny:
I don't think that works when type t depends on payload. See this [version](https://reasonml.github.io/en/try?rrjsx=true&reason=LYewJgrgNgpgBAFwJ4Ad4GUkDsEAsYICWAxgKIBuMOA+oTgGZwC8cA3gFByKrwoCGSKCD5gA3J25o4IPDABOAFR7M4dBOIC+49qEiw4mHPiJlKOAFwHsskxSoJaDFRy7Ip-QcLES38GfkVlFjVNbV1oeFI5ORA5OxxnH2UPIRFqZ3kYuUsAZwQ5OgBzLXYJOmIoCDB4AApDGxJ4hEtWOAkucsrqqyMCRrMHNUYAd0I8SV4BVLA4cxYUr3SJDQBKbRL2WAQ4BZFLKKymgDpdmZZWTNjLACJLuWutNvYAKRyjoUKa06O7taA) where

```reason
  include (SyntheticEvent: { 
    include SyntheticEvent_intf with type payload := payload_ 
  });
```

should bring the definition of `payload` into the module, but that doesn't seem to be the case. It's not an export issue either, even if you move the instance into the module, it still doesn't work as the constructor cannot be found.

jfrolich:
There was a typo in your example, here a fixed [version](https://reasonml.github.io/en/try?rrjsx=true&reason=LYewJgrgNgpgBAFwJ4Ad4GUkDsEAsYICWAxgKIBuMOA+oTgGZwC8cA3gFByKrwoCGSKCD5gA3J25o4IPDABOAFR7M4dBOIC+49qEiw4mHPiJlKOAFwHsskxSoJaDFRy7Ip-QcLES38GfkVlFjVNbV1oeFI5ORA5OxxnH2UPIRFqZ3kYuUsAZwQ5OgBzLXYJOmIoCDB4AApDGxJ4hEtWOAkucsrqqyMCRrMHNUYAd0I8SV4BVLA4cxYUr3SJDQBKbRL2WAQ4BZFLKKymgDpdsHSWVkzYywAiK7kbrTb2ACkco6FCmtOj+7WgA)

jfrolich:
I don't think that works when type t depends on payload, the interface has to specify t as well, as far as I have found.

Why don't you think it doesn't work? Yes type t needs to be in the interface (if you don't want to specify the module type you can do `module type of Xxx`, but I think it's good practice to specify module types.

sgny:
> There was a typo in your example, here a fixed [version](https://reasonml.github.io/en/try?rrjsx=true&reason=LYewJgrgNgpgBAFwJ4Ad4GUkDsEAsYICWAxgKIBuMOA+oTgGZwC8cA3gFByKrwoCGSKCD5gA3J25o4IPDABOAFR7M4dBOIC+49qEiw4mHPiJlKOAFwHsskxSoJaDFRy7Ip-QcLES38GfkVlFjVNbV1oeFI5ORA5OxxnH2UPIRFqZ3kYuUsAZwQ5OgBzLXYJOmIoCDB4AApDGxJ4hEtWOAkucsrqqyMCRrMHNUYAd0I8SV4BVLA4cxYUr3SJDQBKbRL2WAQ4BZFLKKymgDpdsHSWVkzYywAiK7kbrTb2ACkco6FCmtOj+7WgA)

That was my point actually. You get the type `payload_` but not `payload` that should be brought into scope with the `include` statement.

jfrolich:
> That was my point actually. You get the type `payload_` but not `payload` that should be brought into scope with the `include` statement.

Because the `payload` has been replaced with `payload_`.

sgny:
> > That was my point actually. You get the type `payload_` but not `payload` that should be brought into scope with the `include` statement.
> 
> Because the `payload` has been replaced with `payload_`.

I had understood that syntax as replacing the abstract type `payload` with the type `payload`  as defined in `ErrorEvent`, but that was incorrect I see that `othertype` is included in `ErrorEvent`.

Assuming `otherType` is a record where the key `payload` is of type `payload`, we can handle that as in this [version](https://reasonml.github.io/en/try?rrjsx=true&reason=LYewJgrgNgpgBAFwJ4Ad4GUkDsEAsYICWAxgKIBuMOA+oTgGZwC8cA3gFByKrwoDcnbmjgg8MAE4AVHszYoAhkigh5YAFxwUAXwE727UJFhxMOfETKUcG02IsUqCWg1kcuyYf0Ef4o-FJkWVgUlFXVNPT0DcGh4UnFxEHEHHFdvGRDlVVcJRPENAGcEcToAcz13GT8JaWEgzLCorjpiKAgweAAKW3MSFIQNVjhBZqxW9oxsOz6rJzoERgB3QjwhXjg1Fgbs+SwwNZExAOFNw-9amEEtAEpdAXZYBDh5DXi8-oA6auP4esUs8KsXJJDQAIiKJSwpVBWii7CAA), but that seems to require defining `t` for each type of event. As in the example, `otherType` can't be defined inline and now that I understand what the syntax actually does, that's hardly suprising - you cannot have an unnamed type.

However if we already have the definitions of `payload` and `t` in the module, what benefit do we get from the include statement? If the modules had further definitions, that did not depend on `payload` or `t`, sure we can simply replace those and proceed. However, all methods in `EventMethods` also depend on `t`, so we'd have to define them as well.

jfrolich:
> > > That was my point actually. You get the type `payload_` but not `payload` that should be brought into scope with the `include` statement.
> > 
> > 
> > Because the `payload` has been replaced with `payload_`.
> 
> I had understood that syntax as replacing the abstract type `payload` with the type `payload` as defined in `ErrorEvent`, but that was incorrect I see that `othertype` is included in `ErrorEvent`.
> 
> Assuming `otherType` is a record where the key `payload` is of type `payload`, we can handle that as in this [version](https://reasonml.github.io/en/try?rrjsx=true&reason=LYewJgrgNgpgBAFwJ4Ad4GUkDsEAsYICWAxgKIBuMOA+oTgGZwC8cA3gFByKrwoDcnbmjgg8MAE4AVHszYoAhkigh5YAFxwUAXwE727UJFhxMOfETKUcG02IsUqCWg1kcuyYf0Ef4o-FJkWVgUlFXVNPT0DcGh4UnFxEHEHHFdvGRDlVVcJRPENAGcEcToAcz13GT8JaWEgzLCorjpiKAgweAAKW3MSFIQNVjhBZqxW9oxsOz6rJzoERgB3QjwhXjg1Fgbs+SwwNZExAOFNw-9amEEtAEpdAXZYBDh5DXi8-oA6auP4esUs8KsXJJDQAIiKJSwpVBWii7CAA), but that seems to require defining `t` for each type of event. As in the example, `otherType` can't be defined inline and now that I understand what the syntax actually does, that's hardly suprising - you cannot have an unnamed type.
> 
> However if we already have the definitions of `payload` and `t` in the module, what benefit do we get from the include statement? If the modules had further definitions, that did not depend on `payload` or `t`, sure we can simply replace those and proceed. However, all methods in `EventMethods` also depend on `t`, so we'd have to define them as well.

If you don't want to redefine `t` you can do that with a functor I think.

jfrolich:
Ok how about this:

```reason
module type SyntheticEvent_intf = {
  type payload;
};

module MakeSyntheticEvent = (M: SyntheticEvent_intf) => {
  type t = {embeddedPayload: M.payload};
};

module ErrorPayload = {
  type payload = {error: string};
};

module ErrorEvent = {
  include MakeSyntheticEvent(ErrorPayload);
  let payload = {
    embeddedPayload: {
      error: "error",
    },
  };
};

Js.log(ErrorEvent.payload.embeddedPayload.error);
```

Working version [here](https://reasonml.github.io/en/try?rrjsx=true&reason=LYewJgrgNgpgBAFwJ4Ad4GUkDsEAsYICWAxgKIBuMOA+oTgGZwC8cA3gFByKrwoCGSKCD5gA3OwC+49qEiw4AWT4BrGJhz4iZSjmZwAFAoBccdXgIkKVBLQYBKZgD42nbmkR7WMYACMYYMH8ABQEhERMFADp+QWEwKUlpWWh4UgAnNJA0kNiRT1dkdxiwsE8YDKyTAGcENLoAcwSEmXAUuHTMtKtdFg4uOmIoCEDFFTVscy1uhH0OrJySu3EuWAQ4Yrj8ri5vPwDg0LiTPu2dirSTACJyzsuAGlcuCQen8WaAKSrIoXrZ8+noocRJFdv5AmAFnEQeclkA)

jfrolich:
I probably miss context here by why not just parameterize type `t`?

jfrolich:
> Checked just now,
> 
> ```reason
> module ErrorPayload = {
>   type payload = {error: string};
> };
> module ErrorEvent =
>   Event.SyntheticEvent(ErrorPayload);
> ```
> 
> works, so the issue I had really is that the module was defined inline.

My suggestion is very similar to this. What is the problem with this?

sgny:
> > Checked just now,
> > ```reason
> > module ErrorPayload = {
> >   type payload = {error: string};
> > };
> > module ErrorEvent =
> >   Event.SyntheticEvent(ErrorPayload);
> > ```
> > 
> > 
> > works, so the issue I had really is that the module was defined inline.
> 
> My suggestion is very similar to this. What is the problem with this?

What I tried at first was
```reason
module ErrorEvent =
  Event.SyntheticEvent( type payload = {error: string};);
```

which seems reasonable enough. Doesn't throw any compilation errors anyway. However, the resulting module signature has an abstract `t`, so none of the fields are accessible. In contrast, if you try something like 
```reason
module ErrorEvent =
  Event.SyntheticEvent( type payload = string;);
```
you'd have the type defined as expected in the module signature. Since I hadn't identifed the difference, I ended up with the other PR.

MoOx:
Maybe we should focus on UX instead of implementation to help us decide what is the best?
Is there any difference between the 3 pending implantation in term of developer experience?

sgny:
> Maybe we should focus on UX instead of implementation to help us decide what is the best?
> Is there any difference between the 3 pending implantation in term of developer experience?

I've been sitting out on the UX question, but #2 and #3 are identical in use. Scope to access fields and use methods will be the actual event submodules, such as `PressEvent`, etc. This could allow a single local open, such as `PressEvent.(....)`

With #1, the methods are defined in `Event` and not included in each submodule (though it could be revised), so the scope for methods is `Event` but to access fields, one would need `PressEvent`.

I also think we have stalled a new release for too long, so we should quickly decide on this.

jfrolich:
AFAIK 3 is just an improved 2 (better implementation). And 2-3 are better in UX than 1 because it allows a single local open. Or did I overlook something?

MoOx:
Ok thanks for the explanation.
We need to move on & work on 0.63 :)

Are you ok to merge this PR & close the 2 first implementations? @sgny @jfrolich 
@sgny I know you worked hard on this & don't want you to feel that you don't get credit for this as this PR is by @cknitt

sgny:
> Ok thanks for the explanation.
> We need to move on & work on 0.63 :)
> 
> Are you ok to merge this PR & close the 2 first implementations? @sgny @jfrolich
> @sgny I know you worked hard on this & don't want you to feel that you don't get credit for this as this PR is by @cknitt

No worries, I'm fine with it.

MoOx:
Ok, merging & preparing release!!!

PR #698, opened 2020-05-23
Reason react 8.0.0

idkjs:
Greeting sir.

Not closing issues. I was updating `reason-expo` to the ExpoSDK37/`reason-react@8.0.0` and `bs-platform@7.3.2` and hit a renaming of `React.Ref.t` to `React.ref` issue. 

I dont know if this is a thing you want at this point it being so new but here it is if you want it.

Also, should we remove the [bs-platform-7.0.1](https://github.com/reason-react-native/reason-react-native/tree/bs-platform-7.0.1) to maybe keep the branches clean?

Let me know if I should pull this and or remove the other. Thank you sir and the team for all the great stuff going on here.

The two affected files are ``src/elements/NativeElement.re` and `src/elements/Ref.re`. There are no additional `*.bs.js` files produced other the ones that were already there so this doesn't negatively affect the `zero-cost` mission as far as I can tell.


idkjs:
I add a test app generated from the `npx @react-native-community/cli init --template @reason-react-native/template` script then added `react-native` from the pull request.

sgny:
@idkjs It's difficult to see what changes would be implemented with this PR because there are so many changed files and most of those changes have already been merged into master. I see you have 6 commits, but they are against an old base, could you rebase?

Also, there is already #689 to transition to reason-react 0.8.0. If you are proposing further changes please prepare a separate PR (probably once we have merged that one).

idkjs:
Extra files were probably from the example to show that it worked. The change is two lines. Either way, it’s just 8.0.0 so if it’s handled. Great! Thank you, sir.

idkjs:
Thanks @sgny. This is a dupe then of #689

PR #700, opened 2020-06-17
Deprecate Style.list & Style.listOption

### LINKED ISSUE #699: bs-platform 8.0.0 breaks Style.list
BuckleScript 8 changed the encoding of `list`, so `Style.list` does not work anymore.

`Style.list([style1, style2])` now gets compiled to

```js
{
  hd: style1,
  tl: {
    hd: style2,
    tl: /* [] */0
  }
}
```

My suggestion is to deprecate `Style.list`, adding a warning saying that it's incompatible with BS8 and `Style.array` should be used instead, and remove it in one of the next versions.

/cc @MoOx

MoOx:
@cknitt @jfrolich Does this looks enough to you?

Closes #699

### DIFF
--- a/src/apis/Style.rei
+++ b/src/apis/Style.rei
@@ -2,7 +2,13 @@ type t;
 
 external array: array(t) => t = "%identity";
 external arrayOption: array(option(t)) => t = "%identity";
+[@deprecated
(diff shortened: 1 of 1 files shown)


jfrolich:
Looking good!

PR #701, opened 2020-06-25
Update reason-react in peer dependencies

janmeier:
Continuation of https://github.com/reason-react-native/reason-react-native/pull/689


MoOx:
Nice catch!

Issue #707, opened 2020-08-04
RRN doesn't compile on BS8.2


jfrolich:
Due to warnings about redundant use of `bs.string`. Obviously we cannot refactor those yet because of backwards compatibility. But the setting of warnings as errors is not correct. We shouldn't enforce that as a dependency. This means that projects will stop compiling once a future version of Bucklescript has warnings of certain things.

MoOx:
You mean that our setting to block compilation on warnings affect users of this lib?! I was thinking that this settings is internal 😱
If that the case, we should either: just change this, or change this before publication of the repo (should be easy to do)

jfrolich:
Or perhaps run bsb with a flag on the CI? Anyway I posted a PR to fix this in the short term. Would be good to expedite this before everyone starts upgrading.

MoOx:
I guess we could add this flag by hand in a script yeah.

cknitt:
I think this can be closed now?

MoOx:
I guess yeah for now...

MoOx:
FYI I am going to push this change in all repo https://github.com/reason-react-native/safe-area-context/pull/4

PR #708, opened 2020-08-04
No warnings as errors

jfrolich:
Addresses #707


MoOx:
0.62.3 is out and should help :)

MoOx:
The proper way to do this is apparently like this! https://github.com/reason-react-native/safe-area-context/pull/4

PR #714, opened 2020-08-16
Removing tintColor from Switch Component

gedeagas:
This PR Fixes https://github.com/reason-react-native/reason-react-native/issues/705  Remove Switch tintColor


MoOx:
Thanks!

PR #716, opened 2020-08-26
Add binding for PlatformColor

celsobonutti:
This PR adds PlatformColor

### DIFF
--- a/src/apis/PlatformColor.bs.js
+++ b/src/apis/PlatformColor.bs.js
@@ -1,0 +1,234 @@
+'use strict';
+
+var Js_mapperRt = require("bs-platform/lib/js/js_mapperRt.js");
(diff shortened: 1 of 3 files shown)


MoOx (changes requested):
Hi. Thanks for the contribution!
We are trying to have only zero cost bindings to avoid additional cost & layer prone to error.
Ideally, except some very minor thing that we should be able to remove thanks to BS 8.2, the JavaScript code produced should directly hit React & React Native JavaScript implementation, without any extra runtime & bundle cost.
We prefer low level, more verbose bindings instead of "an interpretation" layer. Do you think you can adjust your bindings to match this philosophy?

MoOx:
Closed by #730 

Thanks anyway, your work will be credited!

PR #717, opened 2020-08-26
Add binding for DynamicColorIOS

celsobonutti:
This PR adds [DynamicColorIOS](https://reactnative.dev/docs/dynamiccolorios).

### DIFF
--- a/src/apis/DynamicColorIOS.bs.js
+++ b/src/apis/DynamicColorIOS.bs.js
@@ -1,0 +1,1 @@
+/* This output is empty. Its source's type definitions, externals and/or unused code got optimized away. */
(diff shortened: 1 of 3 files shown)


MoOx (commented):
Hi, thanks for this contribution!

We can read in the official doc "In the future, more keys might become available for different user preferences, like high contrast."
I interpret this as probably optional values but I am not sure... Btw, can this api work with a single color (eg: only light) even if that's a weird choice?
Maybe it's a better idea to have a bs.obj to generate the object instead of a record that would make mandatory all (futur) keys?

MoOx (approved):
Let's merge this for now. We could still add another external binding with bs.obj if more keys are added with a different name (eg: color)

Issue #705, opened 2020-07-20
0.63


MoOx:
Memo for anyone asking: we don't support 0.63 yet. It's to be done.

See https://reactnative.dev/blog/2020/07/06/version-0.63 & https://github.com/react-native-community/releases/blob/master/CHANGELOG.md#0630 for what is to be done

- [x] LogBox #711 
- [x] remove Global.unstable_enableLogBox #718 
- [x] Pressable #712 
- [x] PlatformColor ~~#716~~ #730
- [x] DynamicColorIOS #717 
- [x] Remove Switch tintColor #714

Also things I see, not in the changelog but in the code itself...

- [x] Remove accessibilityComponentType https://github.com/facebook/react-native/commit/f70e58f3559b2f22c41ed867940dcfe170955548
- [x] Remove accessibilityTraits
- [x] Update AccessibilityRole https://github.com/facebook/react-native/blob/3c9e5f1470c91ff8a161d8e248cf0a73318b1f40/Libraries/Components/View/ViewAccessibility.js#L16-L43
- [x] Add AccessibilityActionInfo & AccessibilityActionEvent, onAccessibilityAction https://github.com/facebook/react-native/blob/3c9e5f1470c91ff8a161d8e248cf0a73318b1f40/Libraries/Components/View/ViewAccessibility.js#L46-L55 & https://github.com/facebook/react-native/search?q=AccessibilityActionEvent&unscoped_q=AccessibilityActionEvent

Also, should be checked/done for some repo in the org

- [ ] Dropping iOS 9 and Node.js 8 support (we already dropped node 8 in the repo, we would only have to adjust examples & template for iOS9)

MoOx:
Only DynamicColor is missing with the update of the example, hopefully this will be done soon!

PR #718, opened 2020-08-27
Remove enableLogBox from Global

gedeagas:
Removing enableLogBox from Global because RN63 will be using the new log box as the default logger. 

Related issues.
#705


MoOx (approved):
Thanks!

Issue #720, opened 2020-09-08
BS 8.2 polyvars as strings


cknitt:
With BuckleScript 8.2, polymorphic variants are finally compiled to strings.

This would allow us to get rid of several workarounds where we are using abstract types instead of (polymorphic) variants, e.g. `Platform.os`.

Currently:

```reason
if (Platform.os === Platform.ios) {
  doSomethingForIOS();
} else if (Platform.os === Platform.android) {
  doSomethingForAndroid();
} else if (Platform.os === Platform.web) {
  doSomethingForWeb();
};
```

Better:

```reason
switch (Platform.os) {
| `ios => doSomethingForIOS()
| `android => doSomethingForAndroid()
| `web => doSomethingForWeb()
};
```

Other examples are `AppState.t` and several view prop types like `AccessibilityTrait.t`.

This would be a breaking change though.

/cc @MoOx

MoOx:
Yeah we discussed about that somewhere, but I can't find where.

Indeed that seems obvious to me that we should do such a breaking change as this clearly offer a better DX!
We can probably do this change for the next release.

cknitt:
As a first step, I will prepare a PR that upgrades the BS peer and dev dependency to 8.2 and removes the superfluous `[@bs.string]` annotations.

cknitt:
Actually, as just the upgrade to BS 8.2 causes slight changes to the generated JS code, I think I will do it in 3 steps/PRs:

1. Upgrade BS to 8.2
2. Remove superfluous `[@bs.string]` annotations
3. Breaking changes (replace abstract types by polymorphic variants where possible like in the example above)

cknitt:
Just realized that my original example was probably stupid.

`Platform.OS` is a bit of a special case because of the inlining that Metro bundler does (see #239) and should therefore likely remain the way it is so that we don't tempt people into using it in a way that prevents inlining.

A better example would be `AppState.t` where we would then have:

```reason
  switch (appState) {
  | `active => doSomethingIfActive()
  | `background => doSomethingIfBackground()
  | `inactive => doSomethingIfInactive()
  };
```

PR #724, opened 2020-09-15
Use polymorphic variants instead of abstract types where possible

cknitt:
Part 3 of #720.

I left the `Platform` bindings unchanged because of #239. I do not want to tempt people into `switch`ing on `Platform.os` which will generate code that Metro cannot optimize correctly.


MoOx (approved):
Nice! Thanks for handling this

sgny (approved):
Great simplification, never liked using `@bs.inline` to handle return values.

PR #728, opened 2020-09-27
Accessibility changes for 0.63

cknitt:
This is the accessibility stuff from #705.

* Removed accessibilityComponentType and accessibilityTraits.
* Accessibility.role was actually already up to date (except for the value "switch" which we can't add because it's a keyword, unless we move this back to `@bs.string` which I would like to avoid).
* Accessibility.actionInfo was already there, but broken (missing `unit` to terminate argument list) => fixed.
* Added `AccessibilityActionEvent` and `onAccessibilityAction`.


MoOx (approved):

MoOx:
Thanks for handling this @cknitt !

PR #730, opened 2020-10-23
Add binding for PlatformColor (verbose approach)

rifaldhiaw:
Since it's been quite a while no further responses on the #716 PR, I would like to make this PR as an update.  
Using the verbose approach for better Js Implementation

Related Issue #705


MoOx (approved):
Thanks for this! Sorry for the delay!

PR #733, opened 2020-11-22
Improve View.edgeInsets (add record for easier usage)

MoOx:
I found usually specify all edgeInsets args like this

```reason
<TouchableOpacity
  onPress={(_) => ()}
  hitSlop={View.edgeInsets(
    ~top=20.,
    ~bottom=20.,
    ~left=20.,
    ~right=20.,
    (),
  )}
>
```

With this change, it should allow us to simplify specify a record directly

```reason
<TouchableOpacity
  onPress={(_) => ()}
  hitSlop={{
    top: 20.,
    bottom: 20.,
    left: 20.,
    right: 20.,
  }}
>
```

The function still works, but if you want to specify all args it's far easier imo (less characters).

Thoughts?

### DIFF
--- a/src/components/View.re
+++ b/src/components/View.re
@@ -1,6 +1,11 @@
 include NativeElement;
 
-type edgeInsets;
+type edgeInsets = {
+  left: float,
+  right: float,
+  bottom: float,
+  top: float,
+};
 [@bs.obj]
 external edgeInsets:
   (~left: float=?, ~right: float=?, ~top: float=?, ~bottom: float=?, unit) =>


sgny (commented):
While I agree that this would be a nice improvement, would it not break the convention of using `@bs.obj` for creation and records for consumption? We have made use of this style of double definition before but when the object could be consumed as well as created.

MoOx:
I see this as an alternative. In fact the record is "wrong" because all fields are optional. bs.obj make sense when most things are optional because you can easily add only one field.
Anyway, I guess instead of a record I should do a small function that accept a single argument that I use everywhere...

cknitt:
Certainly convenient, but only for the case when you want to set all four values, like in your example, and as @sgny says it would break our existing convention, so I don't think it is worth it.

MoOx:
I agree :D

PR #735, opened 2021-02-13
Pressable with child function and interaction states

Freddy03h:
Update `<Pressable>` binding with use of function with interactionState param for children and style props.

It's a breaking change with the current binding but I think it's better than having an `<PressableWithChildFunction>` or something like that.
Because the point of using Pressable component is to use those fine interaction states (`pressed` on react-native, with the addition of `hovered` and `focused` on react-native-web) for building a component.

If the developper don't want them he still can use Touchable components or simply
`<Pressable> {_ => <Text/> }</Pressable>`

`hovered` and `focused` are Option because they don't exist (yet?) in react-native, so if you want to create a component that work on both react-native and react-native-web envs, it's better to check values like this
`hoverd->Option.getWithDefault(false)`
But if you have a better idea to handle this, I would like to know it!

### DIFF
--- a/src/components/Pressable.md
+++ b/src/components/Pressable.md
@@ -22,8 +22,15 @@ open ReactNative;
 
 [@react.component]
 let make = () => {
-    <Pressable onPress={_ => Js.log("Pressed")}>
-        <Text> "Press Me"->React.string </Text>
-    </Pressable>;
+  <Pressable
+    onPress={_ => Js.log("Pressed")}
+    style={({pressed}) =>
+      Style.(
+        style(~backgroundColor=pressed ? "rgb(210, 230, 255)" : "white", ())
+      )
+    }>
+    {({pressed}) =>
+        <Text> {pressed ? "Pressed!" : "Press Me"}->React.string </Text>}
+  </Pressable>;
 };
 ```
--- a/src/components/Pressable.md
+++ b/src/components/Pressable.md
@@ -19,7 +19,7 @@ You can read more about this [on the official Pressable documentation](https://r
 
 ## Example
 
-Current Pressable (deprecated)
+Current Pressable (deprecated because doesn't provide pressed state).
 
 ```reason
 open ReactNative;
--- a/src/components/Pressable.md
+++ b/src/components/Pressable.md
@@ -6,7 +6,7 @@ officialDoc: https://reactnative.dev/docs/pressable
 
 Pressable is a Core Component wrapper that can detect various stages of press interactions on any of it's defined children.
 
-⚠️ Current `Pressable` is deprecated, prefer `Pressable_` with the use of interactionState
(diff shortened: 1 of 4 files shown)


MoOx:
I would prefer no breaking changes for now.
Can we got with an addition as `Pressable_` (and keep current Pressable as it is?)
And for 0.64, introduce the breaking change.

Also for RNW, I guess you copied props from other (existing) components?

MoOx:
One thing we could do is to add a deprecation notice to encourage people to use new Pressable_, the in 0.64 breaks things: we remove current Pressable without state props & replace with with Pressable AND we keep Pressable_ with a deprecation notice to go back to Pressable (will be very easy for people to just rename).

Freddy03h:
I'm ok for the `Pressable_` temporary solution to delay the breaking change to the next bigger release: I'll do the change.

For the RNW props, I copied them from touchable components. All thoses props will need to be changed with the new `hrefAttrs` in RNW 0.15, but it's not urgent. we still need to wait ReasonReact to be compatible with React 17 first.

src/components/Pressable.md
```diff
@@ -17,12 +19,27 @@ You can read more about this [on the official Pressable documentation](https://r
 
 ## Example
 
+Current Pressable (deprecated)
```
MoOx, line 22:
```suggestion
Current Pressable (deprecated because doesn't provide pressed state).
```

src/components/Pressable.re
```diff
@@ -7,14 +7,9 @@ external rippleConfig:
...
 
-type interactionState = {
-  pressed: bool,
-  // React Native Web
-  hovered: option(bool),
-  focused: option(bool),
-};
-
-[@react.component] [@bs.module "react-native"]
+[@deprecated "Please use Pressable_ instead"]
```
MoOx, line 10:
```suggestion
[@deprecated "Please use `Pressable_` instead of `Pressable` to benefit of the full benefit of Pressable component (eg: pressed state in style props and children function). In next major release (0.64), `Pressable_` will become `Pressable`."]
```

src/components/Pressable.md
```diff
@@ -6,6 +6,8 @@ officialDoc: https://reactnative.dev/docs/pressable
 
 Pressable is a Core Component wrapper that can detect various stages of press interactions on any of it's defined children.
 
+⚠️ Current `Pressable` is deprecated, prefer `Pressable_` with the use of interactionState
```
MoOx, line 9:
```suggestion
⚠️ Current `Pressable` is deprecated, prefer `Pressable_` that offers interaction states (eg: `pressed` state).
```

MoOx (commented):
I tried to make this more explicit and verbose, are you ok with that?

Freddy03h:
Yes of course I'm ok, thank's for the review :D

MoOx (approved):

MoOx:
Published as 0.63.3

PR #736, opened 2021-02-23
Fix type measure functions

Freddy03h:
Measure functions don't return object like it's now in the binding type.

react-native types : https://github.com/facebook/react-native/blob/master/Libraries/Renderer/shims/ReactNativeTypes.js

I typed callback function directly in the `NativeMethods` file because it's not a big file, so I think it's easier to find the informations we want about the typing of measure functions for someone browsing source files.

But if you prefer `type mesureCallback` & co in the `NativeTypes` file I can update the PR.

Thank's.


MoOx (approved):
LGTM !

MoOx:
I guess you tested this. Merging & releasing :)

PR #738, opened 2021-03-10
ReasonML -> ReScript

### LINKED ISSUE #731: ReScript...
Hi everyone,

Here is a few question for you, do not hesitate to comment or use reactions.

- 👍 / 👎  Now that we require ~~BuckleScript~~ ReScript 8.2 (bs-platform, soon to be renamed to match rescript rebranding), should we convert the codebase to new ReScript syntax? (Is that possible for all bindings btw ? I didn't checked what *isn't* possible with the new syntax) - I think we can even script that part !
- 😄 / 😕  Do we want to rename this organisation and all repo here and on npm to `rescript-*` ?
- 🚀 / 👀 Should we do this now 🚀 (like NOW (republish latest version of current packages with new names) or for 0.64 release (what about other packages?) or should we wait for bs-platform being renamed 👀 ? 
- ❤️ / 🎉 Also for the website I want to do some dogfooding (for fun) but maybe we should go for a docusaurus... ❤️(keep homemade)  / 🎉 (go docusaurus)?

smorimoto:
Closes #731 

The `*.bs.js` files don't really need to be managed by the version control system, so they are removed at the same time.


MoOx:
Hi @smorimoto & thanks for helping. But like I explained this morning in #731 with my last comment, I would like to wait for 0.64 to be published (anytime soon as RN 0.64 is RC4) to publish a last version in ReasonML & then handle the conversion.
Also, .bs.js files are to visualise the generated js code. One goal of this bindings are to ensure zero cost, and visualising generated code help to ensure that we don't produce unwanted & unecessary javascript indirection.
Removing them would make us forget about that goal. Also we cannot just ship this in this package without any minimal checks and the generated JS is a (tiny) part of this.

Last part, keeping ReScript formatting in the git hook is essential to ensure a correctly formatted codebase. So we will have to revisit this PR after changes for 0.64.

We can't just push this as a breaking change just like that. We will do that with a proper transition when we handle the change of name for the package.
For big changes like this, please next time you want to help ask us in issue before handling that. I hope it didn't take you too much time, but I guess thanks to [the command I added to rescript doc](https://rescript-lang.org/docs/manual/latest/migrate-from-bucklescript-reason#upgrade-an-entire-folder) it was fast to do.

Closing this for now. I hope you understand.

MoOx:
> Do you intend to make them available on GitHub? If not, we have no reason to track them.
We are keeping them on GitHub to visualise during PR the generated JS code involved, to ensure there is 0 cost. FYI, previous bindings of this lib where "not zero cost at all" (eg: https://github.com/reasonml-community/bs-react-native/blob/master/bs-react-native/src/style.bs.js). Keeping the generated JS is for us a way to be aware of the code that we are going to add to the end user. For most bindings it doesn't make sense indeed, but since rrn is pretty large, I think it's still better to keep them.

[They are not published in npm module](https://github.com/reason-react-native/reason-react-native/blob/decdb4cb6025d48ab45152ce118e6a14cf94a787/package.json#L24)

> which will only cause problems in the future.

What kind of problems?

smorimoto:
If you commit unnecessary files and make git objects grow unnecessarily, the clones will generally be very slow. That's why I don't commit any files that are not needed.

smorimoto:
Well, I needed a formatting stuff pretty urgently, so I wrote terrible scripts. Feel free to use it if anyone needs them.

`Makefile`:

```makefile
SRCFILES = $(shell git ls-files "*.res" "*.resi")

fmt:
	$(foreach fname,$(SRCFILES),sh scripts/fmt.sh $(fname);)
.PHONY: fmt

fmt-check:
	$(foreach fname,$(SRCFILES),sh scripts/fmt-check.sh $(fname);)
.PHONY: fmt-check
```

`scripts/fmt.sh`:

```sh
#!/bin/sh

fmt() {
  src="$1"
  backup="${src}.bak"

  node_modules/.bin/bsc -format "$src" >"$backup" && mv "$backup" "$src"
}

fmt "$1"
```

`scripts/fmt-check.sh`:
```sh
#!/bin/sh

fmt_check() {
  src="$1"
  expected="${src}.expected"

  node_modules/.bin/bsc -format "$src" >"$expected"

  diff=$(diff "$src" "$expected")

  rm "$expected"

  if [ "$diff" != "" ]; then
    echo "Error: $src"
  fi
}

fmt_check "$1"
```

PR #739, opened 2021-04-16
StyleSheet PPX extension

### LINKED ISSUE #667: Easy syntax for stylesheet without Js.t
## The problem

For new comers, having to use `{"title": ...}` & `styles##title` isn't straightforward.

## Considered solution

We cannot use record in `StyleSheet.create` so we could write a simple ppx to help

```reason
[@react.stylesheet]
let styles = StyleSheet.create({title: ...})

// which will allow
styles.title;
```

I discussed with @bloodyowl about this & he told me a ppx solution should be easy to do.

Do you think that's a good idea to investigate?

(note: this will be optional & shouldn't be a breaking change in any sort)

## Alternatives solutions

I don't see any.

diaozheng999:
Closes #667 

## Introduction

This PR attempts to address the StyleSheet issue with the `%%stylesheet` extension, and also introduce typed styles.

## Usage

Add the following to `bsconfig.json`:
```json
{
  "ppx-flags": [
    ["rescript-react-native/ppx.exe", "-as-ppx"]
  ]
}
```

## Stylesheet

This PR adds the `%%stylesheet` extension, which removes the need for `Js.t` types for stylesheet calls. It allows the use of:

```rescript
%%stylesheet(let style = {
  main: Style.create(~flex=1., ()),
  text: Style.create(~fontSize=14., ()),
})
```

This will allow you to use `style.main` and `style.text` further down in the code (or in another module) if required.

### Interaction with `%style`

This PR also allows for `%style` extension to be used within a `%%stylesheet` extension as such:
```rescript
%%stylesheet(let style = {
  main: %style({ flex: 1. }),
  text: %style({ fontSize: 14. }),
})
```
Here, you'll get `style.main` with type `Style.typed_t<[> #flex ]>>` and `style.text` with type `Style.typed_t<[> #fontSize ]>`.

### Translation

Consider the following:

```rescript
%%stylesheet(let style = {
  style1: Style.create(~flex=1., ()),
  style2: Style.create(~fontSize=14., ()),
})
```

This roughly translates to:
```rescript
type stylesheet = {
  style1: Style.t,
  style2: Style.t,
}

let style: stylesheet = StyleSheet.unsafeCreate({
  style1: Style.create(~flex=1., ()),
  style2: Style.create(~fontSize=14., ()),
})
```

In case a `%style` extension is present, then the tags are also added to the type definition. Consider the following:
```rescript
%%stylesheet(let style = {
  style1: Style.create(~flex=1., ()),
  style2: Style.create(~fontSize=14., ()),
  style3: %style({ flex: 1. }),
  style4: %style({ fontSize: 14. }),
})
```

This would translate to:
```rescript
type stylesheet<+'a, +'b> = {
  style1: Style.t,
  style2: Style.t,
  style3: Style.typed_t<'a>,
  style4: Style.typed_t<'b>,
}

let style: stylesheet<[> #flex ], [> #fontSize ]> = StyleSheet.unsafeCreate({
  style1: Style.create(~flex=1., ()),
  style2: Style.create(~fontSize=14., ()),
  style3: Obj.magic(Style.create(~flex=1, ())),
  style4: Obj.magic(Style.create(~fontSize=14., ())),
})
```

In both these cases, `StyleSheet.unsafeCreate` has the following definition:

```rescript
@module("react-native") @scope("StyleSheet")
external unsafeCreate: 'a => 'a = "create"
```

## Style Typing

Similar to `TyXML` and `bs-css`, we can use tagged types and variance to include/exclude specific css elements. I've defined
```rescript
// Style.resi
type typed_t<+'a>
```

Which allows labels, such as `#flex`, `#alignContent` etc. to be attached. So, a JavaScript style
```javascript
const style = {
  flex: 1,
  marginTop: 20,
  justifyContent: "center",
}
```
would be typed as:
```rescript
let style: Style.typed_t<[> #flex | #marginTop | #justifyContent ]>
```

At consumption, we can write:
```rescript
module Element = {
  @react.component
  let make: (~style:Style.typed_t<[< #flex ]>=?) => React.element
}
```
where we put all the supported style attributes in the list. In the example above, this element only accept the `flex` attribute, so if we were to write:
```rescript
<Element style /> // Compile error
```
This would raise a compile error, since `style` contains at least `#flex`, `#marginTop` and `#justifyContent`, and `Element` wants at most `#flex`.

However, if we had a style:
```rescript
let style1: Style.typed_t<[> #flex ]>
```

Then

```rescript
<Element style=style1 />
```

would be allowed.

### The `%style` extension

Normally, it would be pretty difficult to achieve this kind of covariant tagging, since the only thing that would make sense for it would be lists, and you'll end up with lots of code that breaks the zero-cost principle. So, an extension is introduced to specifically perform this translation. It also has the benefit of making the code look closer to how it's written in JavaScript. In the example above, we fine the JavaScript as
```javascript
const style = {
  flex: 1,
  marginTop: 20,
  justifyContent: "center",
}
```

Currently, we would have to write:
```rescript
let style = Style.style(~flex=1., ~marginTop=dp(20.), ~justifyContent=#center, ())
```

With the `%style` extension, we'll write:
```rescript
let style = %style({
  flex: 1.,
  marginTop: dp(20.),
  justifyContent: #center,
})
```

#### Translation
The extension roughly translates the above call to:
```rescript
let style: Style.typed_t<[> #flex | #marginTop | #justifyContent ]> = Obj.magic(
  Style.style(
    ~flex=1.,
    ~marginTop=dp(20.),
    ~justifyContent=#center,
    ()
  )
)
```

The tags are read from record fields themselves, and the field/value pair gets translated into a labelled argument to the function `style`. To give a very arbitrary example:
```rescript
let arbitrary = %style({
  field1: value1,
  field2: value2,
  field3: value3,
})
```
translates to
```rescript
let arbitrary: Style.typed_t<[> #field1 | #field2 | #field3 ]> = Obj.magic(
  Style.style(
    ~field1=value1,
    ~field2=value2,
    ~field3=value3,
    ()
  )
)
```

#### Practical Usage (tags)

I've included `TypedStyle.res` with definitions scraped from `@types/react-native` definitions, and are clustered similarly.

They include:
1. `flexStyle`
2. `transformsStyle`
3. `shadowStyleIOS`
4. `viewStyle`
5. `textStyleIOS`
6. `textStyleAndroid`
7. `textStyle`
8. `imageStyle`

I've also included utility functions to convert subsets of the above typed styles to `Style.t`. To give an example, we have:
```rescript
let flexStyle: Style.typed_t<[< flexStyle ]> => Style.t
```

We can begin to migrate component definitions to utilise the tagged styles and use these functions for conversion in the meantime.

#### Composition

I've also included 2 ways of composing typed styles. These are:

1. The `compose` function defined in `TypedStyle.res`
2. Overloading the `++` operator defined in module `Infix`.

Both methods are defined using arrays, that is:
```rescript
open TypedStyle.Infix // shadows the ++ operator (the ^ operator in OCaml)

let composed = style1 ++ style2 ++ style3
```
would produce something like:
```javascript
const composed = $caret(style1, $caret(style2, style3))
```
which is equivalent to
```javascript
const composed = [style1, [style2, style3]]
```

## Implementation

The entire project is built on `esy` and `dune`, and exists in the `ppx` directory. In order to allow development work on Windows, OCaml `4.11.x` was used with `ocaml-migrate-parsetree` to allow the code to understand `4.06.1` parsetree generated by `bs-platform`.

A postinstall script is created to copy the `ppx.exe` to the top-level of the package.

## Limitations

The `%%stylesheet` extension only works on "structure items", so:
```rescript
%%stylesheet(let style1 = { ... }) // works

module A = {
  %%stylesheet(let style2 = { ... }) // should work, not tested
}

let make = () => {
  %%stylesheet(let style3 = { ... }) // not supported
}
```

## Next Steps

This is still more or less a work-in-progress, there's some things I'll need to do

1. [ ] Validate build on Windows
2. [ ] Validate multiple stylesheets with same fields
3. [ ] Validate stylesheets within modules

Additionally, I'd like the community's help to validate if the extension can be built and provide your thoughts on using the extension.

You can try this version by changing `package.json`:
```json
{
  "dependencies": {
    "rescript-react-native": "diaozheng999/reason-react-native#stylesheet-ppx-rebased",
  }
}
```


MoOx:
A few things I see that bother me:

- syntax doesn't really match ReScript syntax that use `@annotation`  (eg: `@react.component let make `...)
- postinstall hook is clearly a bummer the way it is (not even sure npm & yarn will support them on the long run, as they cause various issues that we won't discuss here)
- Ppxlib is an issue too, it has been clearly discouraged by ReScript team
- I saw Obj.magic a few times, is this mandatory?
- Infix operator is not a good idea too, and I think ReScript team would be against it usage
- The "Style Typing" is confusing me a little. We already have `style` (not recommended), `viewStyle`, `textStyle` and `imageStyle` that are the only official accepted different styles if I am correct (see https://github.com/facebook/react-native/blob/master/Libraries/StyleSheet/StyleSheetTypes.js).

I am mitigated with this as some parts are cool, but after all this seems to be more complicated. My main idea was to make it easier to write & read style for new comers, not to have something weird and more complex to understand. Having too many way to author styles doesn't seems a good idea. Just having 2 annotations seems way too much.

I would have found something cool if we could just have something like

```rescript
@react.native.styles
let styles = {
  container: viewStyle(/*...*/)
  bodyText: texStyle(/*...*/) 
}

<View style=styles.container>
```

that would translate to

```rescript
let styles = {
  open Style
  StyleSheet.create({
    "container": viewStyle(/*...*/)
    "bodyText": texStyle(/*...*/) 
  }
}

<View style=styles["container"]>
```

This would be a minimal changes (and maintenance). Not more (too complex), not less (otherwise useless).

Curious to know what ReScript think about this (poke @chenglou @bobzhang @ryyppy) as well as other contributors (poke @cknitt @sgny @jfrolich @Freddy03h etc) and users

Note that in all case, if we don't accept this

jfrolich:
It's an interesting idea. Not 100% sure about the benefits. Sometimes it's better to be explicit. Some notes:

- I think extension points are better here, because it fails when the ppx is not run. Annotations are (like the name says) annotations.
- ppxlib is not encouraged by the ReScript team, but neither are ppx's, so I think they won't be in favor of this in any form. In my opinion ppxlib is the most sane way to write a ppx as a maintainer while Hongbo is in favor of a more low level approach of directly working with the parsetree of the current ReScript version, so opinions differ.
- I agree mostly with the comments of @MoOx, but I don't think what he proposes is possible because a ppx works on the parse tree and not the typed tree. The transform of `styles.foo` => `styles["foo"]` is not possible because we won't know the type. Unless the ppx generates a record type definition and uses that for style sheets.
- This is a nice opportunity to infer `dp` and `pct` from the number in the stylesheet; transform `1.0` to `dp` and `100%` to `pct`
- Perhaps just have a `%style` extension for the styles itself, and not have the extension point at the style sheet level, so it can also be used for inline styles
- A thing that I always wanted to build (that would benefit our production performance) is building a ppx that converts all static inline styles to static `Stylesheet`'s.

Freddy03h:
Hi! I'm not a rescript/reason expert, I have 2 projects on production (a react-native app and a react-native-web electron app) but I never wrote ppx myself.

I find ppx intimidating and as a user of lib using it, not really clear of the code it will generate (never found a way to see the "generate reason/rescript" the same way we can see the "generated JS").

For example, I understood only recently how `@react.component` ppx really work and who I can do without it to customize something I wanted. It something I don't yet master with `lenses-ppx` when I use `reform`, still the impression to do dark magic.

Freddy03h:
So I'm more with @MoOx way of seing it as a thin layer.

Maybe is there a way by doing multiple `StyleSheet.create` ?
In react-native, this an identity function and in react-native-web, it will do a forEach on the object to register the style. So I don't se a big difference between calling  StyleSheet.create 1 time with a 4 attributs object or calling StyleSheet.create 4 times with 1 attribut object.

Maybe an outpout like this is acceptable (but I don't know if it can be helpful):
```reason
module Styles = {
  open Style;

  let container =
    StyleSheet.create({
      "idontcare": style(~flexDirection=`row, ~alignItems=`center, ()),
    })##idontcare;
  let inner =
    StyleSheet.create({
      "idontcare": style(~flexDirection=`column, ~alignItems=`flexStart, ()),
    })##idontcare;
};

<View style=Styles.container>
```

chenglou:
I'm no longer supervising anything RRN-related but imo this really isn't worthwhile.

diaozheng999:
Hi guys, Let me address some of the concerns here:

1. ReScript uses a combination of both extensions (e.g. `%json`) and annotations (e.g. `@react.component`). Using extensions here make the PPX extension itself more manageable, since Ppxlib takes away a lot of hard work with parsing the AST.
2. Regarding postinstall script, I do agree. However, the PPX extensions are platform-native binaries. Ideally, when we're distributing, we'll a version for each platform and architecture, and copy the appropriate one to the root directory. If none matches, we'll still need to fallback to using postinstall script to manually do the compilation. This behaviour is similar to other highly native-dependent packages such as `fs-watcher`, and even `bs-platform` itself.
3. The `Obj.magic` is there to allow the expanded ReScript to spit out zero-overhead JS.
4. Having PPX introduces a whole can of worms no matter how you slice it, which may also be why it's highly discouraged by the ReScript team. I guess we have to weigh the benefits of the PPX vs the cost of maintaining it. It won't ever be a thin layer. That was as maintainable as I can make it...
5. I am able to remove the `%style` extension from within the `%%stylesheet` extension, so we can clean up the syntax to look like:
```rescript
%%stylesheet(
  let styles = {
    container: {
      height: Style.dp(0.),
      overflow: #visible,
      flexDirection: #column,
      justifyContent: #flexEnd,
    },
    wrapper: {
      backgroundColor: Color.purple,
      flexDirection: #columnReverse,
      margin: Style.dp(8.),
      alignItems: #flexEnd,
    },
  }
)
```


To give examples of what this PPX is doing currently, these are from the project I'm working on right now:

```rescript
// this was part of an component
%%stylesheet(
  let styles = {
    container: %style({
      height: Style.dp(0.),
      overflow: #visible,
      flexDirection: #column,
      justifyContent: #flexEnd,
    }),
    wrapper: %style({
      backgroundColor: Color.purple,
      flexDirection: #columnReverse,
      margin: Style.dp(8.),
      alignItems: #flexEnd,
    }),
  }
)
```
compiles to
```javascript
import * as ReactNative from "react-native";

var styles = ReactNative.StyleSheet.create({
      wrapper: {
        backgroundColor: "purple",
        alignItems: "flex-end",
        flexDirection: "column-reverse",
        margin: 8
      },
      container: {
        flexDirection: "column",
        height: 0,
        justifyContent: "flex-end",
        overflow: "visible"
      }
    });
```

I've also had style files that contains reusable styles:

```rescript

%%stylesheet(
  let title = {
    style: %style({
      marginTop: dp(49.),
      marginHorizontal: dp(26.),
      height: dp(109.),
      justifyContent: #center,
    }),
  }
)

%%stylesheet(
  let emphasis = {
    outerContainer: %style({marginHorizontal: dp(16.), marginTop: dp(10.)}),
    innerContainer: %style({flexDirection: #row, alignItems: #center}),
    iconContainer: %style({borderRadius: 32.}),
    textContainer: %style({flex: 1., paddingLeft: dp(16.), justifyContent: #center}),
    contentTitle: %style({fontWeight: #bold, fontSize: 16.}),
    contentText: %style({fontSize: 14.}),
    xpContainer: %style({paddingRight: dp(8.), justifyContent: #center}),
    xpAlignmentWrapper: %style({
      flexDirection: #row,
      alignItems: switch ReactNative.Platform.os {
      | p if p == ReactNative.Platform.android => #center
      | _ => #baseline
      },
      justifyContent: #flexEnd,
    }),
    xpValue: %style({fontWeight: #bold, fontSize: 20., paddingRight: dp(2.)}),
    xpLabel: %style({fontWeight: #_400, fontSize: 14.}),
    checkbox: %style({
      width: dp(30.),
      height: dp(30.),
      alignItems: #center,
      justifyContent: #center,
    }),
    checkboxAnimation: %style({width: dp(200.), height: dp(200.)}),
  }
)
```

This compiles to

```javascript
// Generated by ReScript, PLEASE EDIT WITH CARE

import * as ReactNative from "react-native";

var title = ReactNative.StyleSheet.create({
      style: {
        height: 109,
        justifyContent: "center",
        marginHorizontal: 26,
        marginTop: 49
      }
    });

var p = ReactNative.Platform.OS;

var emphasis = ReactNative.StyleSheet.create({
      checkboxAnimation: {
        height: 200,
        width: 200
      },
      checkbox: {
        alignItems: "center",
        height: 30,
        justifyContent: "center",
        width: 30
      },
      xpLabel: {
        fontSize: 14,
        fontWeight: "400"
      },
      xpValue: {
        fontSize: 20,
        fontWeight: "bold",
        paddingRight: 2
      },
      xpAlignmentWrapper: {
        alignItems: (function () {
              switch (p === "android" ? "center" : "baseline") {
                case "flexStart" :
                    return "flex-start";
                case "flexEnd" :
                    return "flex-end";
                case "center" :
                    return "center";
                case "stretch" :
                    return "stretch";
                case "baseline" :
                    return "baseline";
                
              }
            })(),
        flexDirection: "row",
        justifyContent: "flex-end"
      },
      xpContainer: {
        justifyContent: "center",
        paddingRight: 8
      },
      contentText: {
        fontSize: 14
      },
      contentTitle: {
        fontSize: 16,
        fontWeight: "bold"
      },
      textContainer: {
        flex: 1,
        justifyContent: "center",
        paddingLeft: 16
      },
      iconContainer: {
        borderRadius: 32
      },
      innerContainer: {
        alignItems: "center",
        flexDirection: "row"
      },
      outerContainer: {
        marginHorizontal: 16,
        marginTop: 10
      }
    });

export {
  title ,
  emphasis ,
  
}
/* title Not a pure module */

```

bloodyowl:
I think an approach like:

```res
@rn.stylesheet
module Styles = {
  let x = {"foo": "bar"}
  let y = {"foo": "bar"}
}
```

that'd compile to

```res
module Styles_ = {
  let x = {"foo": "bar"}
  let y = {"foo": "bar"}
}

module type Styles_Type = module type of Styles_

module Styles: Styles_Type = unpack(StyleSheet.create(module(Styles_: Styles_Type)))
```

would be both simpler and more useful: dead-code elimination + nice dot access syntax.

diaozheng999:
Regarding typed styles:

The following is allowed in the current implementation
```rescript
let style = Styles.imageStyle(~resizeMode=#cover, ())

@react.component
let make = () => <Text style />
```

However, the current TypeScript implementation doesn't allow this, and React Native is also not happy with this at runtime:
```
Warning: Failed prop type: Invalid props.style key `resizeMode` supplied to `Text`.
Bad object: {
  "resizeMode": "cover"
}
```

When we tag the styles, we can have:
```rescript
let style : Style.t<[> resizeMode ]> = // however we create this

@react.component
let make = () => <Text style />
```

This will produce the following error at compile time:
```
  This has type:
    ReactNative.Style.typed_t<[> #resizeMode] as 'a> (defined as
      \"Style-ReactNative".typed_t<'a>)
  Somewhere wanted:
    ReactNative.TypedStyle.t<[< ReactNative.TypedStyle.textStyle] as 'b>
      (defined as \"Style-ReactNative".typed_t<'b>)

  The incompatible parts:
    'a vs
    'b (defined as
      [<
  | #_end
  | #alignContent
  | #alignItems
  | #alignSelf
  | #aspectRatio
 ...
  | #translateY
  | #width
  | #writingDirection
  | #zIndex
])
  The second variant type does not allow tag(s) `resizeMode
```

This mirrors the error message that React Native gives at runtime:

```
 ERROR    Warning: Failed prop type: Invalid props.style key `resizeMode` supplied to `Text`.
Bad object: {
  "resizeMode": "cover"
}
Valid keys: [
  "display",
  "width",
  "height",
  ...
  "textDecorationLine",
  "textDecorationStyle",
  "textDecorationColor",
  "textTransform",
  "writingDirection"
]
```

Additionally, some libraries read only a subset of the styles. Examples include:
1. `lottie-react-native`: does not support borders
2. `react-native-vector-icons`: does not support margins

We will be able to define what values are acceptable in these cases.

MoOx:
For this last comment: that was a design choice. We didn't wanted to make things more difficult that what they are for new comers. 
We could have used different type for style/viewStyle/textStyle/image styles but we didn't so people can easily start. Anyway, RN will provide a red screen so during development you will be immediately blocked. Not sure if we want to complexity this a lot. This currently allow to reuse some basic styles across some various components.
We also already do things more agressive than ReScript React regarding unit (ReactDOM.Style accept string in most props for example). We didn't wanted to create a huge gap between both layers.

diaozheng999:
I actually prefer @bloodyowl 's suggestion if we can implement that without PPX at all.

bobzhang:
Hi, I am not an expert in styled css. Here are some updates for next release that structural typing is simplified a lot in case you missed it:
- Js.t is not needed any more
- `##` is not needed, you can just write `a#b` (reason syntax) or `a["b"]` (rescript)
we may consider add pattern match over structural objects

bloodyowl:
The main issues with the structural objects APIs is that they don't allow to simply report unused styles, and are pretty syntax heavy. 

I'd say that the ideal PPX-less solution would be the ability to create a external "functor":

```res
@module("react-native") @scope("StyleSheet") external Create: 'a => 'a = "create"
```

So that we'd be able to do:

```res
module Styles = StyleSheet.Create({
  let x = style(~a, ~b, ~c)
  let y = style(~a, ~b, ~c)
})
```

Is it something possible @bobzhang?

cknitt:
👍 for a PPX-less and zero-cost solution.

For the styles themselves, this could be very helpful: https://forum.rescript-lang.org/t/rfc-more-general-type-checking-for-structural-typings/1485

MoOx:
I didn't wanted to close this topic. I renamed master to main branch. Feel free to reopen something on main branch with a PPX-less solution.

PR #741, opened 2021-05-19
add autoGrow to TextInput

jfrolich:


MoOx:
Could you give more info about this prop? It doesn't exist in RN or RNW AFAIK:
- remove in RN https://github.com/facebook/react-native/commit/dabb78b1278d922e18b2a84059460689da12578b
- can't find it in RNW on latest
  - https://github.com/necolas/react-native-web/issues/287#issuecomment-307458609
  - https://github.com/necolas/react-native-web/search?q=autoGrow&type=commits

Where does this comes from?

jfrolich:
Ah ok, I didn't know about this! I implemented a new feature with an autoGrow'able TextInput, but didn't realize that this is already default behavior

Issue #766, opened 2022-08-04
ReScript 10 / records with optional fields


cknitt:
ReScript 10 will support records with optional fields. This would allow us to simplify a lot of things/move them closer to how they are done in JS.

E.g., styles: Instead of

```rescript
let styles = StyleSheet.create({
  open Style
  {
    "statusBar": style(
      ~flexDirection=#row,
      ~height=dp(22.),
      ~alignItems=#center,
      ~paddingLeft=dp(12.),
      (),
    ),
    "text": style(~color=NativeColors.white, ~marginLeft=dp(4.), ()),
  }
})
```

we could have

```rescript
let styles = StyleSheet.create({
  open Style
  {
    "statusBar": style({
      flexDirection: #row,
      height: dp(22.),
      alignItems: #center,
      paddingLeft: dp(12.),
    }),
    "text": style({
      color: "white",
      marginLeft: dp(4.),
    }),
  }
})
```

Or for APIs taking some config object, like

```rescript
Share.share(Share.content(~title="Title", ~url="https://example.com", ()))
```

we could have

```rescript
Share.share({title: "Title", url: "https://example.com"})
```

@MoOx For me the question is, can these be breaking changes or should we try to remain compatible with the existing way where possible?

If we go for breaking changes, then we could also do #748 at the same time.

MoOx:
I guess we can do breaking changes. This will be painful to migrate but it will be more easy for new comers.
Alternatively, since we have bindings, we could offers 2 version, deprecate one, and later kill old ones ?
Maybe a `open ReactNative_Alt` or something that could help people to slowly migrate ?

cknitt:
@MoOx We can do it like this:

```rescript
type content = {
  title?: string,
  message?: string,
  url?: string,
}

@obj @deprecated("Directly create record instead")
external content: (~title: string=?, ~message: string=?, ~url: string=?, unit) => content = ""
```

The record with optional fields will not even compile with pre-10.x compiler versions though.

MoOx:
That's why I didn't merged #767. I wanted to do a release before a v10 only version.

cknitt:
Oh, sorry, I thought #767 was ready to merge as you had approved it.

If you want all breaking changes in the v10 release, then we can either revert #767 or continue with the v10 stuff on master and create a 9.x "maintenance branch".  What would you prefer?

MoOx:
[...]

PR #767, opened 2022-08-06
Get rid of @string/@int for polyvars

### LINKED ISSUE #748: Replace poly variants that use @as with strings
## The problem

When using poly variants for string enums, we transformed a lot of "string-with-dash" to camel case stringWithoutDash, relying on ReScript `@as` annotation.
Now that ReScript allows poly variants to have string syntax, we could  change this choice to make it more straight forward to new ReScript user that are familiar with React Native.

## Considered solution

We could make this a breaking change. Maybe a codemod will help.

## Alternatives solutions

We can also:

* add strings values (so people can use both syntax (with-dash or withoutDash) and keep both (maintenance++)
* integrate both & deprecate later withoutDash if that possible (syntax seems to be accepted but it doesn't seems to show the warning - I tested  `| @deprecated("use #\"plain-text\" instead of #plainText") @as("plain-text") #plainText` in `Alert.res`)

cknitt:
Closes #748.

These are breaking changes, but I think that's ok - more breaking changes likely to come with #766.


MoOx (approved):
Awesome ! Thanks for spending time on this!

cknitt:
Just noticed I had forgotten the `=?` in
```rescript
  ~pointerEvents: pointerEvents=?,
```
in a few places. Updated the PR.

MoOx:
@cknitt  Could you reopen this PR (create a new one yourself so you get proper credit) ? https://github.com/rescript-react-native/rescript-react-native/compare/main...simplify-polyvars?expand=1

cknitt:
Done, thanks!

PR #770, opened 2022-08-18
Get rid of @string/@int for polyvars

cknitt:
Resurrected #767.
Merge for ReScript 10 release.


MoOx (approved):

MoOx (approved):
Awesome work !

Do you think we could do something in PlatformColor for Android ? (I can handle it, I just want your opinion on this inglorious code xD

cknitt:
I have never used the `PlatformColor` API myself. The current implementation looks a bit like "overkill". 🙂

Compare to TypeScript which just has this: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/16db88356a79c458dbe5d88d9706aadec1d20db5/types/react-native/index.d.ts#L6041

I don't quite understand why one can pass an array of color names, what is that used for?

MoOx:
Yeah that's clearly some copy paste here...
The idea is not to have string, but "known" strings that correspond to "registered" color names ( by the OS) or "unsafe" (created by the developer).

And why an array ? Because if you pass more than one string value to the PlatformColor function, it will treat the first value as the default and the rest as fallback depending on their availability (make more sense for "user" values, with system fallbacks).

I will update PlatformColor later then.

cknitt:
@MoOx Ah, but if you can define a type now with `#"@android:color/background_dark"` etc. and no `@as(...)` anymore, then you can use `@variadic`, right? That should drastically simplify things.

MoOx:
Exactly.

PR #771, opened 2022-09-15
Use polymorphic variant for Platform.os

cknitt:
While we are at it, here is another breaking change that we could release with the ReScript 10 version.

Since polymorphic variants are compiled to strings, we can make `Platform.os` a polymorphic variant, enabling pattern matching on it.


MoOx (approved):
Great ! Thanks again !

PR #776, opened 2022-10-18
Fix fontWeight binding (make it an abstract type)

### LINKED ISSUE #775: fontWeight broken in master
In master, we have

```rescript
type fontWeight = [
  | #normal
  | #bold
  | #100
  ...
  | #900
]
```

Unfortunately, this is not correct. We need a string `"100"`, not an int `100`. But this can not be represented as an enum / polymorphic variant (`#"100"` just gets reformatted to `#100`, and as far as I can see it is not currently possible to represent the difference in the AST).

We could use an abstract type + an `@obj external make` with the `@string` annotation instead, but this usage of `@string` is likely to get deprecated in ReScript 11, too.

Any other ideas?

cknitt:
Closes #775

### DIFF
--- a/src/apis/Style.bs.js
+++ b/src/apis/Style.bs.js
@@ -13,7 +13,10 @@ function rad(num) {
   return num.toString() + "rad";
 }
 
+var FontWeight = {};
+
 exports.pct = pct;
 exports.deg = deg;
 exports.rad = rad;
+exports.FontWeight = FontWeight;
 /* No side effect */
--- a/src/apis/Style.res
+++ b/src/apis/Style.res
@@ -51,19 +51,36 @@ type resizeMode = [#cover | #contain | #stretch | #repeat | #center]
 
 type fontStyle = [#normal | #italic]
 
-type fontWeight = [
-  | #normal
-  | #bold
-  | #100
-  | #200
-  | #300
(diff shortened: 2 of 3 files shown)


MoOx (commented):
Nice. Sorry, I didn't had time to look for ideas as my life is a mess ATM :) 
I guess we don't have much choice until syntax allow this.
Just thinking what user will have to do...
Previously `style(~fontWeight=#_100, ())` & now `style(~fontWeight=FontWeight._100, ())`, that's correct ?

Since there is not a lot of value, why not just putting those directly at the root of Style module ?
If opened, this means people can do `style(~fontWeight=_100, ())`.

cknitt:
Yes, that's correct.

I am not in favor of putting things directly into `Style` because:

* Conceptually, these constants clearly belong to `FontWeight`; `Style._100` seems weird compared to `Style.FontWeight._100`.
* If we get another property later that also needs a constant `normal`, we'll have a name clash.
* We are already doing it the same way with `ActivityIndicator.Size.small` (not `ActivityIndicator.small`).

MoOx (approved):

MoOx:
Not a fan (too verbose for me), but I guess you are right.

PR #782, opened 2022-10-22
Set version to 0.70-0.rc.1

cknitt:
I think we should do a release candidate before the final 0.70 release.


cknitt:
If we do an rc, we can e.g. put out a call for testing on the forum.

If you think that's not needed, that's fine with me, too. I will use the git url to test again on our current main project now.

cknitt:
Tested with both JSX 3 and JSX 4, looks good to me.

PR #783, opened 2022-10-24
Escape prop names with \"" instead of _

cknitt:
Needed for JSX 4 compatibility because of https://github.com/rescript-lang/syntax/issues/696.


MoOx (approved):
Oh weird, pretty sure I did those fixes somewhere.
Anyway, thanks !

PR #785, opened 2022-11-01
Add Style.empty

cknitt:
This is nice to avoid options / `Style.arrayOption`, as in

```rescript
Style.array([style1, condition ? style2 : Style.empty])
```


MoOx (approved):
Great idea !

PR #786, opened 2022-11-02
Platform: Add doc comment to prevent user from using switch

fhammerschmidt:
While we cannot prevent the user to use a switch on `Platform.os` which prevents Metro bundler from inlining, we totally can at least mention it in the editor tooltips. 

image


cknitt (approved):

MoOx (approved):
Great !

PR #792, opened 2023-02-15
add lineBreakStrategyIOS Props for Text component

jeong-sik:
Thanks for the good bindings here.
We are developing in a region where Korean text is used. We need this prop for line breaking. 

Closes #<number-of-the-issue>


MoOx:
Included in #793.
Thanks for the PR anyway !

PR #797, opened 2023-03-07
Add gap, columnGap, rowGap to style

cknitt:
See https://reactnative.dev/blog/2023/01/12/version-071#simplifying-layouts-with-flexbox-gap


cknitt:
Yes, indeed, sorry for the duplicate.

PR #800, opened 2023-03-22
Add FlatList method - scrollToIndex

Yisuk:
Hi! Thanks for the great React Native binding.

I added a scrollToIndex method in FlatList.

Thanks!


MoOx:
It's already in here https://github.com/rescript-react-native/rescript-react-native/blob/b1a0741b26ebd21ad1c36fdbf17b1d9cdd69d3a3/src/elements/VirtualizedListMethods.res#L23

PR #803, opened 2023-07-03
change onPress signature

jeong-sik:
Closes #<number-of-the-issue>

The exact signature for onPress looks like this.
```res
onPress?: option<string> => unit
```
https://github.com/facebook/react-native/blob/main/packages/react-native/Libraries/Alert/Alert.d.ts#L15

### DIFF
--- a/src/apis/Alert.res
+++ b/src/apis/Alert.res
@@ -11,7 +11,7 @@ type style = [#default | #cancel | #destructive]
 
 type button = {
   text?: string,
-  onPress?: unit => unit,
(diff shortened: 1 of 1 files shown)


Freddy03h:
Is it really used ? There is no sepecific mention on this in the doc https://reactnative.dev/docs/alert and I don't see an implementation of this in the code : https://github.com/facebook/react-native/blob/7bb7a6037bd78bbfa6d9e8499973ea921e9c70e1/packages/react-native/Libraries/Alert/Alert.js#L100-L109

PR #806, opened 2024-03-04
Fix fontWeight again

### LINKED ISSUE #775: fontWeight broken in master

cknitt:
Fixes #775 again by reverting the `fontWeight` type definition to its pre-#793 state.

Unfortunately this is a breaking change yet again. But better than crashes on Android...

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog of `rescript-react-native`
 
+## Next
+
+- BREAKING: Fix `fontWeight` type again. https://github.com/rescript-react-native/rescript-react-native/pull/806
+
 ## 0.72.0 - 2023-08-04
 
 - Remove deprecated Slider, DatePickerIOS & ProgressViewIOS [82dc6e1](https://github.com/rescript-react-native/rescript-react-native/commit/82dc6e1) by [@MoOx](https://github.com/MoOx)
--- a/src/apis/Style.bs.js
+++ b/src/apis/Style.bs.js
@@ -13,10 +13,13 @@ function rad(num) {
   return num.toString() + "rad";
 }
 
+var FontWeight = {};
+
 var empty = {};
 
 exports.pct = pct;
 exports.deg = deg;
 exports.rad = rad;
+exports.FontWeight = FontWeight;
 exports.empty = empty;
 /* No side effect */
--- a/src/apis/Style.res
+++ b/src/apis/Style.res
@@ -50,7 +50,36 @@ type resizeMode = [#cover | #contain | #stretch | #repeat | #center]
 
 type fontStyle = [#normal | #italic]
 
-type fontWeight = [#normal | #bold | #100 | #200 | #300 | #400 | #500 | #600 | #700 | #800 | #900]
+module FontWeight = {
+  // Note: we cannot model this as a polymorphic variant
+  // because #"100" = #100 = the number 100 in JS, but we need the string "100" here.
+  type t = string
+
+  @inline
+  let normal = "normal"
+  @inline
+  let bold = "bold"
+  @inline
+  let _100 = "100"
+  @inline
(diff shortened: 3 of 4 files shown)


fhammerschmidt (approved):

fhammerschmidt (approved):

Freddy03h:
@cknitt which version of `react-native` are you using?

Number values are accepted since `0.71` : https://github.com/facebook/react-native/pull/34598

I don't experienced any crash on Android or iOS.

cknitt:
That's weird as I had the crash on Android with 0.73.5.

Freddy03h:
I'm still on 0.71.14 so maybe there is a regression on this in `react-native@0.73` but it's seems that those modifications and the tests still exists on `main` branch… so I don't know ^^'

cknitt:
Sorry, patched this in the project I was working on at the time and then lost track of it.

Now a colleague has encountered the same problem in another project with RN 0.74.2. So this is definitely still an issue. Will merge now.

Freddy03h:
I still not experienced the issue since 0.71

Are you using a custom font?

Are you using the new architecture?

Freddy03h:
@cknitt Hi!

- I tried using number on a 0.74 project and it doesn't crash on android.
- I also never experienced a crash linked to this issue on my project with 25k users/day and a variety of android versions and configuration.
- Tests with numbers existing on the react-native project.
- I don't find any related issues on the RN project.

It's seems to be definitively an issue… but only on your project? Can you repro the issue outside?

This type was change 1 year ago, no one else seems to experienced the issue since, and I'm worried the revert to the old type will annoy people…

cknitt:
This is not specific to a particular project of ours. My colleague encountered the issue in a new project, not in the one where I originally experienced it.

I will try to create a repo with a simple reproduction.

cknitt:
@Freddy03h Ok, here is a simple repro:

1. Create a new RN project with `npx react-native init`.
2. Run the Android project.
3. At the bottom of App.tsx, in the `highlight` style, change fontWeight from `'700'` to `700`.

The problem does not occur for the other styles in that file.
Maybe it is because the highlight style is applied on a nested `<Text>`.

Freddy03h:
@cknitt Hi! Thank you for all the informations! I opened an issue an the `react-native` repo https://github.com/facebook/react-native/issues/45285 and the error only happen in Dev mode with the use of `Stylesheet.create` and without the use of an array in the `style` prop.

I never experienced the issue on my project because now with dynamic style (dark/light mode) there's always a variable color to dynamically pass to the style prop in an array.

And because the error is not effective on Release mode, I never saw it on crashlytics (even with 2 text in my entire app where the issue should happen).

It's possible to create a patch-package with this tiny fix that will be fixed soon I hope : https://github.com/facebook/react-native/pull/45299/files

So, what's your opinion about reverting the type to the polymorphic variant? I think it would be nice not to have to rewrite the type for everyone for an issue that happen only in Dev mode in the old architecture.

Thank you.

cknitt:
Hi @Freddy03h! Thank you very much for opening the issue on the React Native repo to get to the bottom of this.

> It's possible to create a patch-package with this tiny fix that will be fixed soon I hope : https://github.com/facebook/react-native/pull/45299/files

I can confirm that this fixes the issue. This solution is fine with me, I'll make a PR to vert the type to the polymorphic variant.

PR #809, opened 2024-06-08
fix/#808/fix for keyboard remove listener and adding missing binding to check keyboard visibility

ashishsingh101:
Closes #<number-of-the-issue>

1. fix for removing keyboard listener
2. adding missing binding for keyboard


ashishsingh101:
@MoOx

Freddy03h (approved):

PR #811, opened 2024-08-08
Upgrade rescript 11

Freddy03h:
- upgrade to rescript 11
- update ActivityIndicator Size Type
- records props for components
- ~~remove deprecated `@obj`~~ > officially mark them as deprecated
- split records styles
- remove HitSlop type
- jsx4 and remove `@react.component` for component with props record


cknitt:
@Freddy03h This is awesome! Thanks a lot for your efforts!

I am currently on vacation and not in front of the computer too much, but I'll try to find some time to test this against our projects in the next days and give some more feedback.

One thing I noticed while skimming through the changes: Instead of `View.viewProps`, `View.viewCoreProps`, `TextInput.textInputProps`, `VirtualizedList.virtualizedListProps` etc. can't we just have `View.props`, `View.coreProps`, `TextInput.props`, `VirtualizedList.props` etc.?

Freddy03h:
You can't name a record `props` directly because `@react.component` will create one. On the SVG binding I named them `_props` but since then I found a convention on the `@rescript/react` doc : https://rescript-lang.org/docs/react/latest/import-export-reactjs#defining-props-types

Also, I'm currently reverting the commit about deleting all the `@obj` and mark them as officially deprecated

cknitt:
Ah, right. But if we go JSX4 only, we don't need the `@react.component` annotation anymore AFAIK.

cknitt:
Like we did in rescript-mui, e.g. https://github.com/cca-io/rescript-mui/blob/867d06d3396563d3a29793d59ccaf9dd7c0ef524/packages/rescript-mui-material/src/components/Button.res#L182

Freddy03h:
I don't remember well but I experienced issue on the svg project but not using `@react.component`… I'll try again.

src/apis/Animated.bs.js
```diff
@@ -106,7 +106,9 @@ function reset(prim) {
 
 var StyleProp = {};
 
-var make = ReactNative.Animated.createAnimatedComponent(ReactNative.FlatList);
+var make = ReactNative.Animated.createAnimatedComponent(function (prim) {
+      return ReactNative.FlatList(prim);
```
cknitt, line 110: This change in output will most likely cause problems. I wonder why it occurs only for Flatlist?

cknitt:
Tried to build one of our projects with the current state of this PR and got lots of errors for `FlatList` about unknown props fields, e.g., "The record field style can't be found", "  The record field contentContainerStyle can't be found", "  The record field ref can't be found".

cknitt:
Also getting this for `SectionList`:

```
  Some required record fields are missing: getItem
  getItemCount. If this is a component, add the missing props.
```

(These props were optional before, and `SectionList` can work without them.)

src/apis/Animated.bs.js
Freddy03h: imho : we should import Animated.* instead of recreating them

cknitt:
Tried again, now I just have

```
  The field ItemSeparatorComponent does not belong to type ReactNative.FlatList.props
```

cknitt:
I have one more issue regarding `FlatList`: The order of props has changed, and with it the inference of the item type.

Before, the item type was inferred from the `data` you passed, now it seems to be inferred from the `keyExtractor` instead.

In some cases I have

```rescript
  ...
  keyExtractor={({id}, _index) => id}
  ...
```

where the item type is now incorrectly inferred to be `ReactIntl.message` (which also has an `id` field) as I have opened `ReactIntl` at the top of the file.

Freddy03h:
Thank's for your feedback on Lists components, I'm not using those on my project so can't test them.

cknitt (approved):
Works fine for me now! Thanks!

PR #815, opened 2024-09-16
0.74 changes

Freddy03h:
- add `space-evenly` to `align-content`
- add `static` to` position`
- `PushNotificationIOS` is deprecated

### DIFF
--- a/src/apis/PushNotificationIOS.res
+++ b/src/apis/PushNotificationIOS.res
@@ -56,16 +56,32 @@ external localNotification: (
   unit,
 ) => localNotification = ""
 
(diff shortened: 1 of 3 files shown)


cknitt (approved):

Freddy03h:
@cknitt thanks for the approval

Do you think we should remove all deprecated `@obj` for 0.74 release?
Also, should we stick to 11.0 or upgrade to 11.1?

cknitt:
Good question. I think maybe we can keep them around for another release or two.
And as long as everything still compiles with 11.0, why not stay with that.

PR #818, opened 2024-10-31
0.75 changes

Freddy03h:
- change `gap`, `columnGap` and `rowGap` type to `size`
- change `translateX` and `translateY` `transform` type to `size`
- remove `UIManager` `showPopupMenu` and `dismissPopupMenu`
- remove `PushNotificationIOS`

https://reactnative.dev/blog/2024/08/12/release-0.75

PR #819, opened 2024-11-07
Js.Promise.t -> promise

cknitt:


Freddy03h:
No need to update `PushNotificationIOS` because it will be deleted in #818

Freddy03h (approved):

PR #821, opened 2024-12-09
Fix transform origin

Freddy03h:
The typescript type on `transformOrigin` is very weak

```
  transformOrigin?: Array<string | number> | string | undefined;
```
https://github.com/facebook/react-native/blob/4165884b700341ba549ed8385480fd98128e7bb1/packages/react-native/Libraries/StyleSheet/StyleSheetTypes.d.ts#L296

Compare to the `processTransformOrigin` function that analyse the transformOrigin value on DEV mode
https://github.com/facebook/react-native/blob/main/packages/react-native/Libraries/StyleSheet/processTransformOrigin.js

### DIFF
--- a/src/apis/Style.res
+++ b/src/apis/Style.res
@@ -47,13 +47,20 @@ type transform
 external unsafeTransform: Js.t<'a> => transform = "%identity"
 
 @unboxed
-type transformOrigin =
-  | @as("top") Top
-  | @as("bottom") Bottom
+type transformOriginX =
   | @as("left") Left
   | @as("right") Right
   | @as("center") Center
   | Size(size)
+
+@unboxed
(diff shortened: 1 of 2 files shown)


cknitt:
Ok, I get the split in x and y 👍, but can't we keep `| @as("left") Left` instead of `| @as("0%") Left`?

Freddy03h:
No because I found that those keywords aren't handled in the array syntax :(
(that's why I split the change in 2  different commits)

I'll open an issue on the `react-native` repo, but, since the `processTransformOrigin` function transform those keywords to `0%`, `50%` and `100%` for the string syntax. I think it's ok to be handled these way, instead of removing the keywords.

cknitt (approved):

PR #822, opened 2024-12-10
Js.Dict.t -> dict

cknitt:


Freddy03h:
You want to remove all ref to `Js` module before the `0.75` release?

cknitt:
Where possible, yes.
There may be some follow-up PRs. 🙂

Freddy03h:
Ok, and you also want to add `Core` ?

Freddy03h (approved):

cknitt:
No, this would be hard, because right now it is an extra dependency, but in v12 it will be part of the `rescript` npm package.

PR #825, opened 2024-12-13
deprecate other @obj

Freddy03h:
I deprecate some missing `@obj` on the project but I can't decide what to do with some of them.

Example with `transform`

```rescript
type transform
@obj external perspective: (~perspective: float) => transform = ""
@obj external rotate: (~rotate: angle) => transform = ""
@obj external rotateX: (~rotateX: angle) => transform = ""
@obj external rotateY: (~rotateY: angle) => transform = ""
@obj external rotateZ: (~rotateZ: angle) => transform = ""
@obj external scale: (~scale: float) => transform = ""
@obj external scaleX: (~scaleX: float) => transform = ""
@obj external scaleY: (~scaleY: float) => transform = ""
@obj external translateX: (~translateX: size) => transform = ""
@obj external translateY: (~translateY: size) => transform = ""
@obj external skewX: (~skewX: angle) => transform = ""
@obj external skewY: (~skewY: angle) => transform = ""
```

In the react-native doc it's:

> array of objects: `{matrix: number[]}`, `{perspective: number}`,`{rotate: string}`, `{rotateX: string}`, `{rotateY: string}`, `{rotateZ: string}`, `{scale: number}`, `{scaleX: number}`, `{scaleY: number}`, `{translateX: number}`, `{translateY: number}`, `{skewX: string}`, `{skewY: string}` or string

We cannot type it with `@unboxed` so I only see two strategies:

1. Record with optional fields
```rescript
type transform = {
  rotate?: angle,
  scale?: float,
  translateX?: float,
  …
}
```
Pro: JS output is clean and  easy to write
Con: Can be error prone

2. Variant
```rescript
type transform2 =
  | Rotate({rotate: angle})
  | Scale({scale: float})
  | TranslateX({translateX: float})
```
Pro: Enforce type
Con: JS output an object with a `TAG` field (even if we can cutomize it with `@tag`)

What's your opinion on it @cknitt ?

### DIFF
--- a/src/apis/Animated.res
+++ b/src/apis/Animated.res
@@ -207,9 +207,9 @@ external parallel: (array<Animation.t>, parallelPayload) => Animation.t = "paral
 @module("react-native") @scope("Animated")
 external stagger: (float, array<Animation.t>) => Animation.t = "stagger"
 
-type loopConfig
-
-@obj external loopConfig: (~iterations: int) => loopConfig = ""
+type loopConfig = {iterations?: int, resetBeforeIteration?: bool}
+@deprecated("Directly create record instead") @obj
+external loopConfig: (~iterations: int=?, ~resetBeforeIteration: bool=?) => loopConfig = ""
 
 // multiple externals
 @module("react-native") @scope("Animated")
--- a/src/apis/AppRegistry.res
+++ b/src/apis/AppRegistry.res
@@ -15,16 +15,20 @@ type appParameters
 
 external asAppParameters: 'a => appParameters = "%identity"
 
-type appConfig
-
-@obj
+type appConfig<'a> = {
+  appKey: string,
+  component?: componentProvider<'a>,
+  run?: appParameters => unit,
+  section?: bool,
+}
+@deprecated("Directly create record instead") @obj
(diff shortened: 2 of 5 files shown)


Freddy03h:
@cknitt ?

cknitt:
Sorry for the late reply. I think we should stay with an abstract type for correctness.
Either keep things the way they are or maybe something like

```rescript
type transform

type perspectiveTransform = {perspective: float}
external perspective: perspectiveTransform => transform = "%identity"

type rotateTransform = {rotate: angle}
external rotate: rotateTransform => transform = "%identity"
...
```
?

Freddy03h:
Thank's for your reply!

I don't think there's a big difference between those two syntaxes. It's not enough to justify a rewriting imho.


```rescript
let transformStyle = [perspective(~perspective=2.0), scale(~scale=1.5)]
```

```rescript
let transformStyle = [perspective({perspective: 2.0}), scale({scale: 1.5})]
```

Freddy03h:
Hi @cknitt, can I have a review? thanks :)

cknitt (approved):

PR #826, opened 2025-01-19
0.76 changes

Freddy03h:
- add `boxShadow` style _(only work with new architecture)_
- add `filter` style _(only work with new architecture)_

https://reactnative.dev/blog/2024/10/23/release-0.76-new-architecture

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -15,7 +15,7 @@
         "npm-run-all": "^4.0.0",
         "npmpub": "^5.0.0",
         "prettier": "^2.0.0",
-        "rescript": "^11.0.1"
+        "rescript": "^11.1.0"
       },
       "funding": {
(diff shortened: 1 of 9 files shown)


Freddy03h:
I don't see other documented changes for react-native 0.76

Do you see some rewrite than can be done if we upgrade to rescript 11.1 @cknitt ?

cknitt:
I still see some occurrences of trailing unit that we could get rid of in `Animated`, `Vibration`, and `DrawerLayoutAndroidMethods`.

cknitt:
Actually also `Alert.alert`, `Alert.prompt`.

cknitt (approved):

Freddy03h:
thanks for the review 😊

I'll wait a week or two before merging it, just in case if we need to patch the 0.75

PR #847, opened 2026-04-15
ImageBackground: add missing children prop

cknitt:
Somehow, ImageBackground's children prop seems to have got lost recently.


Freddy03h (approved):

Freddy03h:
was lost when I refactored all props using records, sorry 😅

